PR #55, opened 2022-04-11
Validate in backend that an user is not added twice 

### LINKED ISSUE #19: Backend validation if user is already member of a repository
Right now, we get an exception with "unknown error" returned to the frontend. We should have a nicer error message here.

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/sqlalchemy/engine/base.py", line 1802, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.9/dist-packages/sqlalchemy/engine/default.py", line 732, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "repository_user_association_pkey"
DETAIL:  Key (username, repository_name)=(user@example.com, repoName) already exists.
```

henrik997:
This PR fixes #19. In the `backend` function `add_user_to_repository` it is checked that an user is not added a second time. Otherwise a `HTTPException` is raised.

### DIFF
--- a/backend/t4cclient/routes/repositories/users.py
+++ b/backend/t4cclient/routes/repositories/users.py
@@ -1,5 +1,6 @@
 import typing as t
 
+# from backend.t4cclient.core.authentication.database import check_username_exists
 import t4cclient.extensions.modelsources.t4c.connection as t4c_manager
 import t4cclient.schemas.repositories as schema_repositories
 from fastapi import APIRouter, Depends, HTTPException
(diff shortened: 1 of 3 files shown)


backend/t4cclient/routes/repositories/users.py
```diff
@@ -1,5 +1,6 @@
 import typing as t
 
+# from backend.t4cclient.core.authentication.database import check_username_exists
```
MoritzWeber0, line 3: Please remove unused comments.

backend/t4cclient/routes/repositories/users.py
```diff
@@ -9,6 +10,7 @@
     is_admin,
     verify_repository_role,
     verify_write_permission,
+    check_username_not_in_repository,
```
MoritzWeber0, line 13: The import is at the wrong location. Could you please apply `isort` for the file?

Issue #70, opened 2022-05-03
Protect user from connection attempts to unprepared sessions


vik378:
At the moment the Session card does not consider session readiness and "Reconnect" button is always enabled. 
This enables user to make connection attempt to a not-yet-existing session container. We should only enable the "(Re)Connect" button when the session status is "started".

![image](https://user-images.githubusercontent.com/8280844/166417427-d2c4e7b6-f74d-4b75-aa4c-d728b3d39822.png)
--> 
![image](https://user-images.githubusercontent.com/8280844/166417460-43f5575b-5e85-46dc-b2c2-58de3444dc83.png)

And also to simplify things a bit we should re-name "Reconnect" to "Connect".

PR #87, opened 2022-05-13
Disable reconnect button as long as session has not started

henrik997:
This PR fixes issue #70.

### DIFF
--- a/frontend/src/app/active-sessions/active-sessions.component.ts
+++ b/frontend/src/app/active-sessions/active-sessions.component.ts
@@ -1,7 +1,7 @@
 // Copyright DB Netz AG and the capella-collab-manager contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, HostListener } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { DeleteSessionDialogComponent } from '../delete-session-dialog/delete-session-dialog.component';
 import { Session } from '../schemes';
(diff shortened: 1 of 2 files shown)


frontend/src/app/active-sessions/active-sessions.component.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright DB Netz AG and the capella-collab-manager contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import { Component, OnInit } from '@angular/core';
+import { Component, OnInit, HostListener } from '@angular/core';
```
MoritzWeber0, line 4: What is the import for?

frontend/src/app/active-sessions/active-sessions.component.ts
```diff
@@ -47,8 +47,10 @@ export class ActiveSessionsComponent implements OnInit {
...
 
   openReconnectDialog(session: Session): void {
-    this.dialog.open(ReconnectDialogComponent, {
-      data: session,
-    });
+    if (session.state == "Started" || session.state == "START_SESSION"){
+      this.dialog.open(ReconnectDialogComponent, {
+        data: session,
+      });
+    }
```
MoritzWeber0, line 54: The check on the button is sufficient. We don't need to verify the state here again.

frontend/src/app/active-sessions/active-sessions.component.ts
henrik997: It's a relict :)

Issue #36, opened 2022-03-14
CDO repo admin account sync


vik378:
We should update repository create / update process such that application admins are also set by default as repository admins

MoritzWeber0:
Done in the referenced commits.

PR #91, opened 2022-05-16
Set admins by default as t4c-repository-admins during project creation

henrik997:
This PR relates to issue #36. Here, application admins are set by default as t4c repository admins during project creation.

### DIFF
--- a/backend/t4cclient/extensions/modelsources/t4c/connection.py
+++ b/backend/t4cclient/extensions/modelsources/t4c/connection.py
@@ -1,13 +1,17 @@
 # Copyright DB Netz AG and the capella-collab-manager contributors
 # SPDX-License-Identifier: Apache-2.0
 
+# Standard library:
 import json
 import logging
 import typing as t
 from socket import timeout
 
+# 3rd party:
 import requests
 from requests.auth import HTTPBasicAuth
+
+# local:
 from t4cclient.config import config
 from t4cclient.core.credentials import generate_password
 
@@ -84,14 +88,14 @@ def add_user_to_repository(
     repository: str,
     username: str,
     password: str = generate_password(),
-    isAdmin: bool = False,
+    is_admin: bool = False,
 ):
     r = requests.post(
         config["modelsources"]["t4c"]["restAPI"] + "/users",
         params={"repositoryName": repository},
         json={
             "id": username,
-            "isAdmin": isAdmin,
+            "isAdmin": is_admin,
             "password": password,
         },
         auth=T4C_BACKEND_AUTHENTICATION,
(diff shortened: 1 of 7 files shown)


mocks/t4c-server/mock/users.py
```diff
@@ -2,13 +2,22 @@
...
+    id: str
+    isAdmin: bool
+    password: str
+
+
 @router.post("/")
-def add_user():
-    return {}
+def add_user(body: PostServerRequest, repositoryName: str):
+    print(body)
```
MoritzWeber0, line 19: Please remove the `print(body)` statement.

backend/t4cclient/extensions/modelsources/t4c/connection.py
```diff
@@ -81,14 +81,17 @@ def fetch_last_seen(mac_addr: str):
...
+    password: str = generate_password(),
+    isAdmin: bool = False,
 ):
     r = requests.post(
         config["modelsources"]["t4c"]["restAPI"] + "/users",
         params={"repositoryName": repository},
         json={
             "id": username,
-            "isAdmin": False,
+            "isAdmin": isAdmin,
```
MoritzWeber0, line 94:
```suggestion
            "isAdmin": is_admin,
```

Please use snake_case for the python variables.

backend/t4cclient/extensions/modelsources/t4c/connection.py
```diff
@@ -81,14 +81,17 @@ def fetch_last_seen(mac_addr: str):
 
 
 def add_user_to_repository(
-    repository: str, username: str, password: str = generate_password()
+    repository: str,
+    username: str,
+    password: str = generate_password(),
+    isAdmin: bool = False,
```
MoritzWeber0, line 87:
```suggestion
    is_admin: bool = False,
```

backend/t4cclient/routes/repositories/__init__.py
```diff
@@ -82,6 +82,7 @@ def create_repository(
     token=Depends(JWTBearer()),
 ):
     verify_admin(token, db)
+    connection.add_user_to_repository(body.name, get_username(token), isAdmin=True)
```
MoritzWeber0, line 85: What is your intention here? Beside the fact, that the repository doesn't even exist at this point (because it is only created in the next step), we also don't need to create the user here as it would use a random password.

MoritzWeber0 (changes requested):
Please implement the behavior for the function `patch_repository_user` too (otherwise the admin status would be overwritten if a user changes their password). And make that you pass `is_admin` to all calls of `add_user_to_repository`.

Issue #85, opened 2022-05-10
Introduce session monitoring with Prometheus


ewuerger:
From the [kubernetes configuration](https://github.com/DSD-DBS/capella-collab-manager/blob/staging/helm/values.yaml) it clear that there is a "sessions" namespace.

Steps to introduce monitoring for individual sessions:

1. Make a new kubernetes namespace "monitoring"
2. Make a kubernetes ConfigMap for Prometheus
3. Make a Prometheus Pod (Is this required or already there?)
    1. Create a kubernetes service for prometheus (prometheus-service.yaml)
    2. Use label selector for pods labeled with name: prometheus
    3. Kubernetes opens NodePort in each node in cluster (9090)

PR #99, opened 2022-05-23
Monitor inactivity in project sessions

henrik997:
This PR adds the functionality to measure the `last_seen` status of sessions by integrating [Prometheus](https://prometheus.io/).

This fixes issue #85.


MoritzWeber0:
We have a severe privacy risk with this PR: There is no authentication in front of Prometheus - which means that everybody can access the idletime for each container.

PR #105, opened 2022-06-07
Update documentation

TimoBbz:
# Description

This PR make small adjustements to the documentation, that were needed for a first installation. It also fixes some undocumented behaviors.

# Testing

No test needed for a CONTRIBUTING.md

# Checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,6 +106,7 @@ Run the following steps:
 1. The backend uses various configuration settings. You can find them in the `config`
    directory.
    Please copy the file `config_template.yaml` to `config.yaml` and adjust the values.
+   Specify the admin username replacing the value at the key `initial:admin`.
 
    *Hint*: If you already have the k8d cluster running and the if you have the
    application deployed, then no configuration values need to be adjusted.
@@ -221,14 +222,14 @@ Run the following steps:
 1. Optional: If you like to use your custom favicon, please copy it to `src/favicon.ico`
 1. Optional: If you like to use your custom theme, replace the file `src/custom-theme.scss`.
    You can generate custom themes [here](http://mcg.mbitson.com/)
-1. Copy the file `src/environment.ts` to `src/environment.dev.ts` and adjust the values.
+1. Copy the file `src/environment.ts` to `src/environment.dev.ts` and adjust the values. In order to access the backend, set `backend_url` to your backend (defaults to "http://localhost:8000/api/v1").
 1. Run the frontend with:
 
    ```sh
    make dev
    ```
 
-1. You should see the frontend running von port 4200.
+1. You should see the frontend running from port 4200.
 
 We additionally recommend that you set up your editor / IDE as follows.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -107,6 +107,7 @@ Run the following steps:
    directory.
    Please copy the file `config_template.yaml` to `config.yaml` and adjust the values.
    Specify the admin username replacing the value at the key `initial:admin`.
+   Specify the url to the Prometheus server at the key `prometheus:url`.
 
    *Hint*: If you already have the k8d cluster running and the if you have the
    application deployed, then no configuration values need to be adjusted.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -221,14 +223,14 @@ Run the following steps:
...
-1. Copy the file `src/environment.ts` to `src/environment.dev.ts` and adjust the values.
+1. Copy the file `src/environments/environment.ts` to `src/environments/environment.dev.ts` and adjust the values. In order to access the backend, set `backend_url` to your backend (defaults to "http://localhost:8000/api/v1").
 1. Run the frontend with:
 
    ```sh
    make dev
    ```
 
-1. You should see the frontend running von port 4200.
+1. You should see the frontend running from port 4200.
```
MoritzWeber0, line 233:
```suggestion
1. You should see the frontend running on port 4200.
```

Thanks for fixing the typo. `on` is a bit more beautiful in my eyes, what do you say?

backend/t4cclient/core/authentication/provider/oauth/__main__.py
```diff
@@ -1,11 +1,14 @@
...
 
+# local:
 from .keystore import KeyStore
+from t4cclient.config import config
 
 cfg = config["authentication"]["oauth"]
 
 
-def get_jwk_cfg(token: str) -> dict[str, any]:
+def get_jwk_cfg(token: str) -> t.Dict[str, any]:
```
MoritzWeber0, line 11: Type hinting in standard collection was introduced with PEP 585 in Python 3.9. 
We should rather set the minimum Python version to 3.9.

CONTRIBUTING.md
```diff
@@ -106,6 +106,8 @@ Run the following steps:
 1. The backend uses various configuration settings. You can find them in the `config`
    directory.
    Please copy the file `config_template.yaml` to `config.yaml` and adjust the values.
+   Specify the admin username replacing the value at the key `initial:admin`.
+   Specify the url to the Prometheus server at the key `prometheus:url`.
```
MoritzWeber0, line 110: These two values are already set in the `config_template.yaml` (https://github.com/DSD-DBS/capella-collab-manager/blob/update-documentation/backend/config/config_template.yaml#L124-L125). Could you please check you local configuration?

CONTRIBUTING.md
TimoBbz: I though it was the german word “von” and not a typo on “on”, so no problem, I put back “on”.

backend/t4cclient/core/authentication/provider/oauth/__main__.py
TimoBbz: But when I check in Python 3.8, the typing module contains a Dict generic, and also since dict is a type, it's not (and has never been) subscriptable, that means that if we want to use dict instead of Dict, we cannot specify the type of the keys.

TimoBbz: We can either use
```python
def get_jwk_cfg(token: str) -> dict:
```
or 
```python
def get_jwk_cfg(token: str) -> t.Dict[str, any]:
```
but no
```python
def get_jwk_cfg(token:str) -> dict[str, any]:
```

CONTRIBUTING.md
MoritzWeber0: You can just leave it admin, then you can login with username "admin" :) 
It is somehow documented here: https://github.com/DSD-DBS/capella-collab-manager/blob/staging/doc/getting_started.md

Wuestengecko (changes requested):

backend/t4cclient/core/authentication/provider/azure/__main__.py
```diff
@@ -1,5 +1,6 @@
 from t4cclient.config import config
-
+from __future__ import annotations
+import typing as t
```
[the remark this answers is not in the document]
TimoBbz: It does, but actually change many files, is that a problem for this PR ?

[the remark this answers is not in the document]
MoritzWeber0: I will rewrite the precommit configuration today and also include `isort`.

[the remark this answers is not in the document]
TimoBbz: Ok, do you prefer to do it on this branch or on another?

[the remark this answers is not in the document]
MoritzWeber0: On another branch, but we can rebase it afterwards.

PR #111, opened 2022-06-09
Replace ngx simplebar

### LINKED ISSUE #83: Remove dependency "simplebar-angular"
There are some dependency conflicts with the current Angular version. Additionaly, the last version was published 8 months ago. We should consider removing "simplebar-angular".

TimoBbz:
# Description

The ngx-simplebar wrapper was removed from the project and its two occurrences were replaced. Now it uses .simple-scroll class and relies on ::webkit-scrollbar for the style. The new style is accessible with the .simple-scroll class for all scrollable blocks.

image

There is no transition between the normal and hovered effect, as it's not possible without ~~a sorcery~~ a significant complexity overhead that for the sake of reusability I prefer not using.

Resolves #83 

# Testing

# Checklist

### DIFF
--- a/frontend/src/custom-theme.scss
+++ b/frontend/src/custom-theme.scss
@@ -46,3 +46,19 @@ $theme: mat.define-light-theme(
 );
 
 @include mat.all-component-themes($theme);
+
+/* width */
+.simple-scroll::-webkit-scrollbar {
+  width: 10px;
+}
+
+/* Handle */
+.simple-scroll::-webkit-scrollbar-thumb {
+  background: grey;
+  border-radius: 10px;
+}
+
+.simple-scroll::-webkit-scrollbar-thumb:hover {
+  background: lightgrey;
+  border-radius: 10px;
+}
--- a/backend/.dockerignore
+++ b/backend/.dockerignore
@@ -1,1 +1,4 @@
 config
+.venv
+.vscode
+build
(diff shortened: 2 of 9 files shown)


frontend/src/custom-theme.scss
```diff
@@ -46,3 +46,19 @@ $theme: mat.define-light-theme(
...
+/* Handle */
+.simple-scroll::-webkit-scrollbar-thumb {
+  background: grey;
+  border-radius: 10px;
+}
+
+.simple-scroll::-webkit-scrollbar-thumb:hover {
+  background: lightgrey;
+  border-radius: 10px;
+}
```
MoritzWeber0, line 64: Can we find an alternative to this approach? 

The docs (https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar) state:
"Non-standard: This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future."

TimoBbz: As far as I know, it's the only way to style a scrollbar. Simplebar also used this solution. I would advise to keep the default scrollbar for now

MoritzWeber0: I checked firefox, looks good too. The default style is fine.

Issue #124, opened 2022-07-11
Add option for whole history and only latest commit pull for read-only sessions


ewuerger:
The second stage for spawning read-only sessions takes very long. That may be caused by fetching the whole .git history. When the user doesn't need to check out older versions this unnecessarily takes time for information that is never used.

### Solution 
Add another option to specify a concrete version. When this option is filled a partial/shallow clone is done. This is the main target.

### Optional/Quality of Life
Add an option to fetch commits by a specified tag. So the user can choose from a drop-down menu of available tags. 

### Error handling
Raise InvalidCommit/Tag errors.

linked issue #103

Issue #103, opened 2022-06-03
Readonly containers: Allow selection of branches and tags


MoritzWeber0:
Allow the selection of branches and tags of the linked Git repository.

PR #138, opened 2022-07-26
Accelerate loading of readonly sessions

henrik997:
After selection of a project repository for a readonly session, two selection fields appear (based on the existence of the attribute `chosenRepository`) where a specific branch/tag (or all) and the commit history depth (latest commit or complete commit history) can be chosen.
As a project repository is chosen, the primary git model of that project repository is taken and theirs branches and tags are fetched through suprocess. These are communicated to the frontend by the HTTP GET Request, `get_revisions`.
Additionally, a checkbox is created for marking if all branches should be downloaded. If it was selected, the selection field of revisions is grayed. **Note:** Choosing all revisions and only the latest commit clones only the latest commits of all branches near the tips.

Default values are set such that the cloning of a git model is as fast as possible, i.e. `branch` is set to latest used branch (or main or master branch) and `depth` to latest commit.

Solves issues #124, #103 and #4.


backend/t4cclient/sessions/routes.py
```diff
@@ -135,13 +135,42 @@ def request_session(
...
+        elif body.depth == DepthType.CompleteHistory:
+            depth = 0
+        else:
+            raise HTTPException(
+                status_code=400,
+                detail={
+                    "err_code": "wrong_depth_format",
+                    "reason": f"Depth type {depth} is not allowed.",
+                },
+            )
```
MoritzWeber0, line 164: The type check is already done with the pydantic models - the code will never be executed. If the body does not match the model, fastAPI will return a bad request error.

backend/t4cclient/routes/repositories/__init__.py
```diff
@@ -67,6 +71,76 @@ def get_repositories(db: Session = Depends(get_db), token=Depends(JWTBearer())):
...
+):
+    remote_refs: dict[str, list[str]] = {}
+    remote_refs["branches"] = []
+    remote_refs["tags"] = []
+
+    git_model = get_primary_model_of_repository(db, project)
+    try:
+        url = git_model.path
+        log.info("Fetch revisions of git-model, %s,  with url: %s", git_model.name, url)
+    except AttributeError:
```
MoritzWeber0, line 88: When did this occur?

MoritzWeber0: I assume it does just occurs if there is no git model assigned. I changed it, please check my changes.

backend/t4cclient/extensions/modelsources/git/routes.py
```diff
@@ -148,12 +148,8 @@ def get_revisions(
         elif ref.startswith("refs/tags/"):
             remote_refs["tags"].append(ref[len("refs/tags/") :])
 
-    if git_model.revision != "HEAD":
-        remote_refs["default"] = git_model.revision
-    elif "master" in remote_refs["branches"]:
-        remote_refs["default"] = "master"
-    else:
-        remote_refs["default"] = "main"
+    remote_refs["default"] = git_model.revision
```
henrik997, line 147: I think it would be better to keep the if else statements since we then set in either case a default branch in the frontend. 
Think about when `git_model.revision` is HEAD which would be neither in `remote_refs["tags"]` nor `remote_refs["branches"]`. Then, we would have no branch by default in the frontend. Alternatively, we could think about to resolve the HEAD, but the HEAD could potentially e.g. refs/pull/81/head which would be again not part of `remote_refs["tags"]` or `remote_refs["branches"]` and therefore, no branch would be set by default in the frontend.
Same applies when `git_model.revision` has a typo error which would be then again not included in `remote_refs["tags"]` nor `remote_refs["branches"]`.

MoritzWeber0: To take your example: Let's imagine repository where the default branch is `staging`, which is used as `HEAD`. 
If one specifies `HEAD` in the backend as revision, the implementation would default to main, if such an branch exists - therefore, it would suggest a wrong branch. 

Another example is one large model inside DB, where the branch "nightly" is used as main branch. 

In general, a specific and existing branch or tag should be used when assigning a repository. We should rather validate the branch or tag selection with a validator in the frontend.

PR #150, opened 2022-08-04
Add model handling

TimoBbz:
# Description

- Adding the following userstory in order to add a model to a project. It starts when a user clicks on the Add model button in a project. The user enters basic informations about the model (name, description and tool) and submit, then choose a source from a fixed list (this should depend on the tool in the future). The only available source for now is "existing repository". The user then enters the informations about the repository and submits. Then he choose the initialization of the model in a fixed list. The only available initialization is the empty model initialization. The user then enters the empty initialization informations (version and type of tool) and submits. The model is updated in database.

- The project service was reworked to work as a state machine to reduce the number of backend fetchs.
- The model service was created.
- Now use the newly created project slug to identify projects in urls instead of name (name are case sensitives and can use different alphabets, so cannot be used as path. path are supposed to be ascii strings: https://url.spec.whatwg.org/#url-representation)
- Various changes in CSS, mainly for accessibility purposes (links should be links even when they look like buttons, texts should be in inline containers, not in blocks, etc.)
- Updated the database to:
  - add the Tools, Versions, and Types tables
  - alter the Model table
  - add the slug to the Project table
- Added a route to fetch revisions from a git repository

Resolves #issue

# Testing

How were the changes tested?

# Checklist


backend/capellacollab/core/database/__main__.py
```diff
@@ -40,6 +42,7 @@ def migrate_db():
 
         context = MigrationContext.configure(conn)
         current_rev = context.get_current_revision()
+        tools_exist = inspect(engine).has_table('tools')
```
henrik997, line 45: I think it would be more intuitive this way: 
`if not inspect(engine).has_table('tools'):`

backend/capellacollab/core/database/__main__.py
```diff
@@ -51,6 +54,9 @@ def migrate_db():
             command.stamp(alembic_cfg, "head")
             initialize_admin_user()
             initialize_default_repository()
+        
+        if not tools_exist:
+            create_tools()
```
MoritzWeber0, line 59: This should be done by the migration scripts and the alembic scripts should also be the only location where that should happen.

backend/capellacollab/models/crud.py
```diff
@@ -0,0 +1,75 @@
...
+
+def get_all(db: Session, project_slug: str) -> t.List[Model]:
+    project = (
+        db.query(DatabaseProject).filter(DatabaseProject.slug == project_slug).first()
+    )
+    assert project is not None
+    return db.query(Model).filter(Model.project_id == project.id).all()
+
+
+def get_id(db: Session, id_: int) -> Model:
```
henrik997, line 22: Same here, because you return no id, but a model. So, I would rename it to `get_model_by_id`

backend/capellacollab/core/database/__main__.py
TimoBbz: The check must be performed before the tables are created, but the insertion must be performed after, hence the separation.

MoritzWeber0: Why do we need the check at all? When the migration scripts were successful, they should always be there, right?

backend/capellacollab/models/crud.py
```diff
@@ -0,0 +1,75 @@
...
+from sqlalchemy.orm import Session
+
+# 1st party:
+import capellacollab.projects.crud as projects_crud
+from capellacollab.models.models import EmptyModel, Model, NewModel, ToolDetails
+from capellacollab.projects.models import DatabaseProject
+from capellacollab.tools.models import Tool, Type, Version
+
+
+def get_all(db: Session, project_slug: str) -> t.List[Model]:
```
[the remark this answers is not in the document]
TimoBbz: Is it really worth it? This would create some redundancy with the name of the crud module. For exemple if `models.crud` is imported as `models_crud`, The function will be called with `models_crud.get_all_models`.

backend/capellacollab/core/database/__main__.py
henrik997, line 45: Because `tools_exist` is used only once, I would suggest not to create it, i.e.: <br>
`if not inspect(engine).has_table('tools'):` <br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; `create_tools()`

TimoBbz: It’s not possible. The condition must be evaluated before a possible creation of the tools table, but the insertion of the tools must be executed after such creation.

backend/capellacollab/models/crud.py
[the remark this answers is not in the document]
henrik997: It's not super important, but I think it's better to make the repository a bit more consistent and maybe `get_all_git_models` is even better.

backend/capellacollab/extensions/modelsources/git/crud.py
```diff
@@ -68,3 +77,13 @@ def delete_model_from_repository(db: Session, capellamodel_id: int, model_id: in
...
     db.commit()
+
+
+def create(db: Session, project_slug: str, model_slug: str, source: NewGitSource):
+
+    model = models_crud.get_slug(db, project_slug, model_slug)
+    new_source = DB_GitModel.from_new_git_source(model.id, source)
+    db.add(new_source)
+    db.commit()
+    print(new_source)
```
henrik997, line 88: Please delete the print statement

backend/capellacollab/models/crud.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def create_empty(db: Session, project_slug: str, new_model: EmptyModel) -> Model:
+    project = (
+        db.query(DatabaseProject).filter(DatabaseProject.slug == project_slug).first()
+    )
+    tool = db.query(Tool).filter(Tool.id == new_model.tool_id).first()
+    version = db.query(Version).filter(Version.id == new_model.version_id).first()
+    model_type = db.query(Type).filter(Type.id == new_model.type_id).first()
+    assert tool is not None and version is not None and model_type is not None
```
henrik997, line 62: Are the assert statements necessary? If so, then a error message would be nice

backend/capellacollab/models/crud.py
```diff
@@ -0,0 +1,78 @@
+# Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
+
```
henrik997, line 3: Please apply isort on all files

backend/capellacollab/models/routes.py
```diff
@@ -0,0 +1,116 @@
...
+@router.get("/{project_slug}/details/", response_model=ResponseModel)
+def get_slug(
+    project_slug: str,
+    slug: str,
+    db: Session = Depends(get_db),
+    token=Depends(JWTBearer()),
+) -> ResponseModel:
+
+    project = projects_crud.get_slug(db, project_slug)
+    verify_project_role(project, token, db)
```
henrik997, line 43: By default `allowed_roles` contains all roles: "user", "manager", "administrator".  In this case, there is no reason for verification because it would be always true. We should define e.g. `allowed_roles=["admin", "manager"]`

backend/capellacollab/projects/crud.py
```diff
@@ -37,3 +41,13 @@ def create_project(db: Session, name: str) -> DatabaseProject:
...
     db.commit()
+
+
+def get_slug(db: Session, slug: str) -> DatabaseProject:
+    project = db.query(DatabaseProject).filter(DatabaseProject.slug == slug).first()
+    assert project is not None
+    return project
+
+
+def get_id(db: Session, id: int):
```
henrik997, line 52: The return type is missing: `DatabaseProject`

backend/capellacollab/tools/crud.py
```diff
@@ -0,0 +1,76 @@
...
+from ensurepip import version
+
+from sqlalchemy.orm import Session
+
+from capellacollab.tools.models import Tool, Type, Version
+
+############
+### Tool ###
+############
+
```
henrik997, line 14: Please delete these comments to make it more consistent to the repository since these are nowhere else used.

TimoBbz: Ok but I still think that it’s useful to group the functions to ease the navigation in the file. Do you have any other suggestion than these comments?

backend/capellacollab/projects/crud.py
TimoBbz: Actually all the routes using the ID to fetch a project are unused, that’s also why they are not well finished. I’m removing them.

frontend/src/app/models/choose-init/choose-init.component.ts
```diff
@@ -0,0 +1,26 @@
+// Copyright DB Netz AG and the capella-collab-manager contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { Component, OnInit } from '@angular/core';
+import { ActivatedRoute } from '@angular/router';
```
henrik997, line 5: Mabye we add here: `Params` such that we can describe the return type below.

frontend/src/app/models/choose-init/choose-init.component.ts
```diff
@@ -0,0 +1,26 @@
...
+})
+export class ChooseInitComponent implements OnInit {
+  constructor(
+    public projectService: ProjectService,
+    public modelService: ModelService,
+    private route: ActivatedRoute
+  ) {}
+
+  ngOnInit(): void {
+    this.route.params.subscribe((params) => {
```
henrik997, line 22: We could write here: `params: Params` such that we get no linter message.

frontend/src/app/models/create-coworking-method/create-coworking-method.component.ts
```diff
@@ -0,0 +1,156 @@
...
+import { merge, Observable } from 'rxjs';
+import { GitModel, ModelService } from 'src/app/services/model/model.service';
+import { ProjectService } from 'src/app/projects/service/project.service';
+import {
+  Credentials,
+  GitService,
+  Instance,
+} from 'src/app/services/git/git.service';
+import { Source, SourceService } from 'src/app/services/source/source.service';
+
```
henrik997, line 23: `GitModel`  and `AsyncValidatorFn` can be deleted.

TimoBbz: I removed these useless imports and some others.

backend/capellacollab/models/routes.py
TimoBbz: I removed those useless checks.

frontend/src/app/models/create-coworking-method/create-coworking-method.component.ts
```diff
@@ -0,0 +1,155 @@
...
+            error: (e) => {
+              this.filteredRevisions = { branches: [], tags: [] };
+              subscriber.next({ credentials: { value: e.name } });
+              subscriber.complete();
+            },
+          });
+        }
+      }, 500);
+    });
+  };
```
henrik997, line 52: `validateCredentials` is never used. So, it can be deleted.

backend/capellacollab/models/crud.py
TimoBbz: Replaced with HTTPExceptions.

frontend/src/app/models/create-model/create-model.component.ts
```diff
@@ -0,0 +1,78 @@
...
+        if (slug == new_slug) {
+          return { uniqueSlug: { value: slug } };
+        }
+      }
+      return null;
+    };
+  }
+
+  ngOnInit(): void {
+    this.toolService.get_tools().subscribe((tools) => {});
```
henrik997, line 52: `tool` can be replaced by a wildcard, e.g. `_`

TimoBbz: Actually I can just call subscribe without any callback.

backend/capellacollab/models/crud.py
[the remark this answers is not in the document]
TimoBbz: I renamed the crud functions with more explicit names. However I don’t intend yet to specify “git”, because these functions should not be specialised in one source type.

frontend/src/app/projects/service/project.service.ts
```diff
@@ -3,42 +3,90 @@
```
henrik997, line 3: Please move this file into frontend/src/app/services/project/project.service.ts

frontend/src/app/projects/service/project.service.ts
```diff
@@ -3,42 +3,90 @@
...
+
+  getProject(name: string): Observable<Project> {
+    return this.http.get<Project>(this.BACKEND_URL_PREFIX + name);
+  }
+
   updateDescription(name: string, description: string): Observable<Project> {
-    return this.http.patch<Project>(this.BACKEND_URL_PREFIX + name, {
-      description,
+    let url = new URL(name, this.base_url);
+    console.log(description);
```
henrik997, line 66: Here is still a print statement.

frontend/src/app/services/model/model.service.ts
```diff
@@ -0,0 +1,176 @@
...
+      });
+    });
+  }
+
+  getId(id: number): Observable<Model> {
+    let url = this.base_url;
+    return this.http.get<Model>(url.toString(), { params: { id } });
+  }
+
+  getSlug(slug: string, project_slug: string): Observable<Model> {
```
henrik997, line 118: I would rename `getSlug` to `getProjectBySlug`.

TimoBbz: I renamed this one and some of the model service too.

PR #185, opened 2022-08-16
Local performance

MoritzWeber0:
# Description

This PR improves the local performance on devices with low resources. There is now a separate resource definition for local deployment. 

In addition, there is a new goal for `make` to run the kubernetes dashboard. 

# Testing

Run `make dashboard` to see and test the local dashboard.
Run `make deploy` to test the local resources. All containers were still in the running state.


henrik997 (approved):
Looks great!

PR #201, opened 2022-08-29
Feature: download files

### LINKED ISSUE #48: Allow download and upload of files from/to the container
The user should be able to download and upload files from/to the container, if they are placed in a specific directory. 
This will allow the download of diagrams.

amolenaar:
# Description

Add the capability to download files from a running container.

Resolves #48

* The download mechanism is streaming. It does not store the download somewhere in the middle.
* It currently downloads a zip, so Windows explorer knows how to handle it.
* Download is only on folder level (as implemented by @henrik997). File level downloads should be easy. You'll get a zip file.

# Testing

Backend calls have been unit tested (with mock). Frontend has manually been checked.

# Checklist

- [x] My code follows the guidelines of this project: [CONTRIBUTING.md](/DSD-DBS/capella-collab-manager/blob/staging/CONTRIBUTING.md)

image

### Notes

* Tests have been created for the download feature. However, those tests do not run in CI. I marked them as `skip` for now, until the code base is in a shape to support those tests.


backend/t4cclient/sessions/files/__init__.py
```diff
@@ -0,0 +1,5 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
```
MoritzWeber0, line 5:
```suggestion
```

The license header got inserted twice (one is the old version and one is the new version).

MoritzWeber0:
> * This implementation is pretty bare bones: it blocks while downloading. This could cause problems with many concurrent users or very large repo's.

Indeed, this could be an issue and there doesn't seem to be asynio support for the kubernetes-client python library: https://github.com/kubernetes-client/python/issues/323. This is another argument for using our custom file service inside the session container. 

> * It currently downloads a tar. Not sure if Windows explorer knows how to handle those.

As far as I can see, it uses `zip` now. This should be fine for Windows users.

backend/t4cclient/sessions/files/routes.py
```diff
@@ -86,3 +87,24 @@ def upload_files(
...
+    filename: str,
+    db: Session = Depends(get_db),
+    token=Depends(JWTBearer()),
+) -> UploadFile:
+    check_session_belongs_to_user(get_username(token), id, db)
+
+    return StreamingResponse(
+        OPERATOR.download_file(id, filename),
+        headers={
+            "content-disposition": 'attachment; filename="download.zip"',
```
MoritzWeber0, line 107: The filename should not be `download.zip`. I messes up the Download directory on the file system und you have no idea which download it was later on. We could just use the name of the directory.

backend/tests/test_session_operator_k8s.py
```diff
@@ -0,0 +1,57 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
```
MoritzWeber0, line 5: Here is another occurrence for the duplicated license header.

frontend/src/app/services/load-files/load-files.service.ts
```diff
@@ -29,4 +29,12 @@ export class LoadFilesService {
       this.BACKEND_URL_PREFIX + id + '/files?show_hidden=' + showHiddenFiles
     );
   }
+
+  download(id: string, filename: string): Observable<Blob> {
+    return this.http.request(
+      'GET',
```
MoritzWeber0, line 35:
```suggestion
    return this.http.get(
```

You can just use `http.get` instead.

frontend/src/app/active-sessions/upload-dialog/upload-dialog.component.ts
```diff
@@ -246,6 +248,14 @@ export class UploadDialogComponent implements OnInit, OnDestroy {
       });
   }
 
+  download(filename: string) {
+    this.loadService
+      .download(this.session.id, filename)
+      .subscribe((response: Blob) => {
+        saveAs(response, 'download.zip');
```
MoritzWeber0, line 255: It would be nice to see a progress of the download. I tried it with our largest models, I clicked the button and for about one minute, nothing happened. After one minute, the download was just there. 

One workaround could be to use just a href tag: https://stackoverflow.com/a/61190093, but this would only capture the download progress between the backend and the frontend (not between container and backend). In the end we are at the point of implementing async again.

MoritzWeber0:
The implementation doesn't seem to work with large files unfortunately. I just tried it out with our largest model (~1GB), but the downloaded zip seems to be corrupted: 
```bash
> unzip download\ \(5\).zip
Archive:  download (5).zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of download (5).zip or
        download (5).zip.zip, and cannot find download (5).zip.ZIP, period.
```

frontend/src/app/active-sessions/upload-dialog/upload-dialog.component.ts
amolenaar: I added some fancy progress bars (see screenshot above)

amolenaar:
Version v2 branch is in #274

PR #229, opened 2022-09-09
Rename legacy repositories to projects

MoritzWeber0:
In v1, repositories had another meaning than in v2 and the repositories in v1 are now called projects in v2. Therefore, we need to rename all references.


TimoBbz (commented):
Does this PR intends to account for to-be-deleted files?
If so, the keyword argument `repository` is still used in the module `models`.

MoritzWeber0:
> Does this PR intends to account for to-be-deleted files? If so, the keyword argument `repository` is still used in the module `models`.

After a discussion, the comment was related to an outdated version. Seems to be fixed now.

TimoBbz (approved):

PR #248, opened 2022-09-15
fix: update kubernetes dashboard token generation

dominik003:
# Description

In newer versions of kubectl account tokens are no longer auto-generate for ServiceAccounts. Therefore, make dashboard just outputs an empty token for those versions. This PR fixes this and should be backwards compatible, so also support older kubectl versions


amolenaar:
Nice!

PR #255, opened 2022-09-23
Staging

MoritzWeber0:


frontend/src/app/services/auth/auth.service.ts
```diff
@@ -36,20 +64,45 @@ export class AuthService {
...
+    return this.http
+      .put<PostTokenResponse>(
+        environment.backend_url + '/authentication/tokens',
+        { refreshToken }
+      )
+      .pipe(
+        map((res) => {
+          this.logIn(res.access_token, res.refresh_token);
+          return res;
+        }, first())
```
TimoBbz, line 78: Since the `res` is not altered by the `map` operator, I’d rather use a `tap`.
```typescript
pipe(
  tap(res => {
    this.logIn(res.access_token, res.refresh_token);
  }),
  first()
)
```

Also, is the `first` operator really useful there? It’s only useful if the observable doesn’t complete directly after the first sent value.

PR #263, opened 2022-09-27
Rename helm chart

MoritzWeber0:
# Description

This renames the helm chart to `collab-manager`, which is the name of the platform for a while now. I don't expect any breaking changes. There were no issues when upgrading a existing application with the old name.


MoritzWeber0:
I tried it out in the local cluster and nothing changed. We don't use the chart name in the templates, only the release name, which doesn't change.

amolenaar (approved):
LGTM

PR #264, opened 2022-09-27
Some model stepper improvements

MoritzWeber0:
Some refactoring, fixes and improvement for PR #212.

PR #266, opened 2022-09-29
Allow to abort a running model creation

MoritzWeber0:
# Description

- Allow to abort a running model creation during the project creation. 
- Redirect to project after model creation when called from the project creation context. 
- Some small fixes.

#264 has a higher importance, this PR can be reviewed after #264.

### DIFF
--- a/backend/capellacollab/core/authentication/database/__init__.py
+++ b/backend/capellacollab/core/authentication/database/__init__.py
@@ -6,7 +6,6 @@ import sqlalchemy.orm.session
 from fastapi import Depends, HTTPException
 
 import capellacollab.projects.users.crud as repository_users
-import capellacollab.users.crud as get_user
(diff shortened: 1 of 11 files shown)


TimoBbz:
image
The icon is not at the right height

TimoBbz:
Shouldn’t the dialog to abort model creation be available for the model creation outside of the project creation stepper?

MoritzWeber0:
> image The icon is not at the right height image Here to.

This has been fixed in the underlying branch already.

> Also, shouldn’t the dialog to abort model creation be available for the model creation outside of the project creation stepper?

This is a valid point, but I don't see it in the scope of this PR. I opened an issue for it: https://github.com/DSD-DBS/capella-collab-manager/issues/276

TimoBbz (approved):

PR #275, opened 2022-10-04
Add frontend testing

dominik003:
# Description

The frontend tests were already initialized, but didn't really work due to several problems. The first problem was that the _path-is-absolute_ and _webpack_ dependencies were missing, which are now included. Several other problems were caused by spec files that were forgotten during refactoring, so some imports and other things were wrong and are now updated.
In addition to the two dependencies above, _karma-coverage-istanbul-reporter_ was added to visualize code coverage and see which lines are covered and which are not. Lastly, _playwright_ was added used for E2E testing. The general code coverage statistics (i.e., statement, branch, function, and line coverage) are printed in the console and the advanced visualization can be viewed by opening the _index.html_ file in the _frontend/test-result/istanbul-coverage_ folder.

### Integration testing

This PR contains two integration test suites, one for the Project Service and one for the Create Project component. The tests can be run with `make test`. The PR also disables all tests that are not yet implemented to clearly distinguish which tests are successful and which are not due to a problem.
The project service test mainly ensures that the service calls the backend correctly. This means that it calls the correct http methods, makes the correct number of calls, and verifies that the expected results are returned (i.e., that no unexpected conversions occur after receiving the faked http response).
The component test verifies that the conditional buttons are active only when they should be, that the child components are loaded in each step, and that the cancel and finish buttons correctly lead to the subsequent components. To simplify testing (i.e., to avoid having the same code for retrieving an element or set an elements' value over and over again), several element helper functions have been added.

### End-to-End testing

Unlike integration tests, E2E tests must have a running backend in order to run correctly. This PR includes an E2E test that covers the entire project creation process. This includes authentication at the beginning, followed by the actual project creation and verification that each click leads to the expected page. The E2E test can be run with `npx playwright test` and to additionally get the trace as output (i.e. screenshots of each step and action) one can add `-trace on` as an option.

# Checklist


frontend/tsconfig.json
```diff
@@ -16,7 +16,7 @@
     "target": "es2017",
     "module": "es2020",
     "lib": [
-      "es2018",
+      "es2022",
```
amolenaar, line 19: Why this upgrade? Do we really need it? Does it have impact on the generated frontend code (can it work on last years browsers)?

amolenaar:
We probably want to run the NG tests as part of `test.yaml` and the Playwright tests can be ran from `push.yaml` once the whole application is started.

frontend/tsconfig.json
dominik003: The upgrade to ES2022 is required because `Array.at(index)` is used in the create-model component, which was first introduced in ES2022, and the test runner complains about it. However, I only checked if we can use ES2022 with our current node version and didn't think about possible problems with the browser version. As far as I can see (https://caniuse.com/mdn-javascript_builtins_array_at), most browsers support using Array.at since mid-2021 (except Safari, which supports it since March this year). So I think we should either just add "es2022.array" in addition to "es2018" so that we don't include all ES2022 functions, or just rewrite the "Array.at" part

MoritzWeber0:
Would be nice to have this mentioned in our development documentation. You can find it in `docs/development` :)

docs/development/docs/frontend/testing.md
```diff
@@ -0,0 +1,92 @@
+<!--
+ ~ SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+ ~ SPDX-License-Identifier: Apache-2.0
+ -->
+
+# Frontend Testing
+
+## Integration testing
```
MoritzWeber0, line 8: Is "Unit testing" not the better term for it? It doesn't really test the integration (as it is not connected to the backend)

dominik003: It depends on the test you write. You can also use it to test the integration of different components and services. This article nicely describes the different test types that are possible (https://vsavkin.com/three-ways-to-test-angular-2-components-dcea8e90bd8d). But I agree that only naming it Integration test is confusing, so maybe change it to “Unit & Integration testing“?

MoritzWeber0: "Unit & Integration testing" sounds good.

frontend/tests/project-creation.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  await page.goto('http://localhost:4200/');
+
+  // Check whether you got redirected to the /auth page
+  await expect(page).toHaveURL('http://localhost:4200/auth');
+
+  await page.locator('button:text-matches("Login with .*", "g")').click();
+
+  // Wait for the response from the backend
+  await page.waitForResponse('http://localhost:8000/api/v1/authentication/');
```
MoritzWeber0, line 24:
```suggestion
  await page.waitForResponse(environment.backend_url + '/authentication/');
```

The tests should read the `environment.ts` and use the specified backend URL. This allows us to tests in different environments (e.g. in a cluster or staging environment).
I'd suggest to create a separate environment OR configuration file and use it for the playwright tests. In general, the configurable options should also be configurable for the tests. 

When using a custom playwright environment, you can import it with: 
```
import { environment } from 'src/environments/environment.dev';
```

frontend/playwright.config.ts
```diff
@@ -0,0 +1,99 @@
...
+    //   },
+    // },
+  ],
+
+  /* Folder for test artifacts such as screenshots, videos, traces, etc. */
+  outputDir: 'test-results/playwright',
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
+    command: 'npm run start',
```
MoritzWeber0, line 94:
```suggestion
    command: 'ng serve --configuration playwright --port 4201',
```

We should use a separate configuration here (the configuration has to be created first and use the `environment.playwright.ts` environment file. In addition, it is helpful to start the server on another port (this allows to run the dev frontend and the tests at the same time).

frontend/playwright.config.ts
```diff
@@ -0,0 +1,99 @@
...
+    port: 4200,
```
MoritzWeber0, line 95:
```suggestion
    command: 'ng serve --configuration playwright --port 4201',
    port: 4201,
```

We should use a separate configuration here (the configuration has to be created first and use the `environment.playwright.ts` environment file. In addition, it is helpful to start the server on another port (this allows to run the dev frontend and the tests at the same time).

frontend/tests/project-creation.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+const nearPixelArea = 50; // Specifies how far away an element can be when using near
+
+test('test the full project creation workflow', async ({ page }) => {
+  const testProjectName = `test-project-name-${Math.trunc(
+    Math.random() * 1000
+  )}`;
+  const testProjectDescription = 'test-project-description';
+
+  await page.goto('http://localhost:4200/');
```
MoritzWeber0, line 16: The localhost URL should be configurable, it would allow testing against deployed applications (running on a specific URL).

frontend/tests/project-creation.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // Check whether you got redirected to the /auth page
+  await expect(page).toHaveURL('http://localhost:4200/auth');
+
+  await page.locator('button:text-matches("Login with .*", "g")').click();
```
MoritzWeber0, line 21: This line failed for me, because the "RELEASE NOTES" dialog was popping up. I've reworked the dialog so that it doesn't popup automatically: https://github.com/DSD-DBS/capella-collab-manager/pull/288

MoritzWeber0:
> We probably want to run the NG tests as part of `test.yaml` and the Playwright tests can be ran from `push.yaml` once the whole application is started.

For the `ng test` part, I opened another PR to add Github actions: https://github.com/DSD-DBS/capella-collab-manager/pull/287

frontend/src/app/helpers/spec-helper/element.spec-helper.ts
```diff
@@ -0,0 +1,69 @@
+/*
```
MoritzWeber0, line 1: What do you think about moving this file to the `tests` directory? It is test specific and I'd expect helpers in the `src` directory that are also used in the "production" code (except the `.spec.ts` files).

MoritzWeber0 (approved):

frontend/src/app/helpers/spec-helper/element.spec-helper.ts
dominik003: The problem at the moment is that the `tests` directory is not inside the `/src` directory, which makes sense to me as this way we have a clear separation between the unit/integration tests and the E2E tests. So I personally would prefer to keep it that way. However, we could create another folder inside the `/src` directory where we then put the spec helpers.

MoritzWeber0: In theory, they don't need to be part of the `tests` folder. You can still reference them relatively.

PR #279, opened 2022-10-04
Unique style for t4c instances

MoritzWeber0:
# Description

Apply the existing style to TeamForCapella instances to make them consistent with existing application.


TimoBbz:
Is the full width large instance creation form voluntary? It seems too large for me. Apart from that, there is nothing

MoritzWeber0:
> Is the full width large instance creation form voluntary? It seems too large for me. Apart from that, there is nothing

Fixed with https://github.com/DSD-DBS/capella-collab-manager/pull/279/commits.

PR #281, opened 2022-10-05
T4C instances improvements

MoritzWeber0:
# Description

Some improvements for the T4C instances PR (https://github.com/DSD-DBS/capella-collab-manager/pull/277). See commits for more details.


frontend/src/app/settings/modelsources/t4c-settings/edit-t4c-instance/edit-t4c-instance.component.html
```diff
@@ -80,15 +105,18 @@ <h2 *ngIf="existing"></h2>
...
       </fieldset>
-      <div *ngIf="existing == 'editing'" id="editing-button-wrapper">
+      <div *ngIf="editing" id="editing-button-wrapper">
         <button mat-flat-button color="warn" (click)="cancelEditing()">
           Cancel
         </button>
         <button mat-flat-button type="submit" color="primary">Submit</button>
       </div>
-      <div *ngIf="existing == 'existing'">
+      <div *ngIf="!editing">
```
TimoBbz, line 119: This condition should be `!editing && existing`, else the button appears on creation step.

frontend/src/app/settings/modelsources/t4c-settings/edit-t4c-instance/edit-t4c-instance.component.ts
```diff
@@ -47,7 +52,6 @@ export class CreateT4cInstanceComponent implements OnInit {
     host: new FormControl('', Validators.required),
     port: new FormControl(null as number | null, [
       Validators.required,
-      Validators.pattern(/[\d^0]\d*/),
```
TimoBbz, line 50: This validator was useful to assert that the port did not start with 0, to prevent for example a port looking like 01337.

frontend/src/app/settings/modelsources/t4c-settings/edit-t4c-instance/edit-t4c-instance.component.html
```diff
@@ -43,10 +48,21 @@ <h2 *ngIf="existing"></h2>
         <div class="field-separator"></div>
         <mat-form-field appearance="fill">
           <mat-label>Port</mat-label>
-          <input matInput inputmode="numeric" formControlName="port" />
-          <mat-error *ngIf="form.controls.port.errors"
-            >The port is invalid.</mat-error
+          <input
+            matInput
+            inputmode="numeric"
+            type="number"
```
TimoBbz, line 54: It should be a numeric string but not a number. The number type should only be used for mathematical numbers, not strings that appear to be composed of digits. The `inputmode="numeric"` is here to hint that it’s composed of digits. https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

Another issue is that the number type allows floats (and Angular then truncates then to integers, but still).

A blog post on the matter: https://stackoverflow.blog/2022/09/15/why-the-number-input-is-the-worst-input/

MoritzWeber0: Interesting, thanks for sharing the articles. Actually, I always thought that a port is a number - but even this is not the case: https://stackoverflow.com/a/47992797. But yeah, this is a pretty rare case - I think we can stick with numbers, but I'll remove the `type="number"` annotation.

frontend/src/app/settings/modelsources/t4c-settings/edit-t4c-instance/edit-t4c-instance.component.ts
TimoBbz: I also checked, without this decorator, decimal number (that are no valid port) are accepted.

MoritzWeber0: Yes, decimal numbers should be permitted of cause. I will change the validator pattern to `/^\d*$/`.

backend/capellacollab/settings/modelsources/t4c/routes.py
```diff
@@ -101,5 +101,5 @@ def edit_t4c_instance(
     for key in body.dict():
         value = body.__getattribute__(key)
         if value:
-            instance.__setattr__(key, body.__getattribute__(key))
+            instance.__setattr__(key, value)
```
TimoBbz, line 104: You could even use the walrus operator here if you wanted (introduced in Python 3.8). I find it elegant, but it’s not widespread
```python
if value := body.__getattribute__(key):
    instance.__setattr__(key, value)
```

MoritzWeber0: Absolutely! I like it, but always forget to use it :D

TimoBbz (approved):

PR #283, opened 2022-10-05
Link Git models with general models

### LINKED ISSUE #222: Link models with Git repositories
Models of a Capella Collaboration project should be assignable to a exiting git repository. The global Git instance setting has to be used. 
The instance baseURL should be expandable with a specific path in the instance. We should prevent Path Traversal Attacks. 

image

dominik003:
# Description

This PR provides the functionality to add an existing git repository to a model. When adding a git repository, there are different cases to consider. In general, you can distinguish between the case where at least one base model source is specified in the settings and the case where no model sources are specified. In both cases, you must provide credentials if you want to access a private repository, and then you can select the branch or tag from those available. Below, I describe the two cases for providing a repository url.

### No model sources specified in the settings
In this case, the user can enter any Git repository (i.e. without any restrictions), and the selection field for the base integration url is hidden. The only requirement in this case is that the input has a valid url prefix (i.e. http:// or https://) and does not contain certain sequences that are not allowed (i.e., ".." and "%"). 

### At least one model source specified in the settings
In this case there is an additional selection input where you have to select one of the available git settings. The resulting repository url then has the url from the selected git settings as a prefix (e.g. if only https://github.com is specified in the git settings, you can only access repositories that have this url as a prefix). In the second text entry field, one can then specify the relative path starting from the url from the selected git settings (the relative path must start with "/"). This means that the resulting repository url is the selected git settings url followed by the relative path specified in the text input field. 
To allow the user to also copy and paste URLs, the text input can also be given an absolute url, and the git setting will then automatically be set to the one with the longest matching prefix.

### Summary of the possible behaviors
- the revision is required in all cases
- the revision must match a revision retrieved from the backend
- in any case, it is not allowed to have blacklisted sequences in the url
- if at least one git setting exists, one must select one and the resulting url must have the git setting url as prefix
- if no git setting exists, the selection input should not be visible and required
- when entering an absolute url, the git setting with the longest prefix match should be automatically selected (if available)

Resolves #222

# Checklist


frontend/src/app/projects/models/create-coworking-method/create-coworking-method.component.ts
```diff
@@ -13,58 +20,139 @@ import {
...
-          tags: this.gitService.instance.tags.filter((tag) =>
-            tag.startsWith(value as string)
-          ),
-        };
+    this.gitService.instance.subscribe((revisions) => {
+      this.availableRevisions = revisions;
+      this.form.controls.revision.updateValueAndValidity({
+        onlySelf: false,
+        emitEvent: true,
+      });
```
MoritzWeber0, line 84:
```suggestion
      this.form.controls.revision.updateValueAndValidity();
```

Should be sufficient.

frontend/src/app/projects/models/create-coworking-method/create-coworking-method.component.ts
```diff
@@ -13,58 +20,139 @@ import {
...
+    let urlGroup = this.form.controls.urlGroup;
+    if (
+      urlGroup.invalid ||
+      urlGroup.controls.baseUrl.invalid ||
+      urlGroup.controls.inputUrl.invalid
+    ) {
+      return;
+    }
+
+    let projectName = this.projectService?.project?.name || '';
```
MoritzWeber0, line 118:
```suggestion
    let projectName = this.projectService!.project!.name;
```

If you expect that a project is there, use the `!` instead of the `?`.

frontend/src/app/settings/modelsources/git-settings/git-settings.component.ts
```diff
@@ -33,45 +48,68 @@ export class GitSettingsComponent implements OnInit {
...
+      if (gitSettingName == newInstanceName) {
+        return { uniqueName: { value: gitSettingName } };
+      }
+    }
+    return null;
+  }
+
+  urlValidator(control: AbstractControl): ValidationErrors | null {
+    return null;
+  }
```
MoritzWeber0, line 114: This can be deleted, I assume :)

frontend/src/app/services/git/git.service.ts
```diff
@@ -23,29 +22,28 @@ export interface Instance {
...
-            password: credentials.password,
-          },
-        })
-        .subscribe((value) => {
-          this.instance = value;
-          subscriber.next(value);
-          subscriber.complete();
-        });
-    });
+  loadInstance(
```
MoritzWeber0, line 34: The naming is confusing here. It's not really loading a instance, it's fetching the revisions of a git repository.

frontend/src/app/projects/models/create-coworking-method/create-coworking-method.component.ts
```diff
@@ -73,22 +161,112 @@ export class CreateCoworkingMethodComponent implements OnInit {
...
           this.projectService.project.name,
           this.modelService.model.slug,
-          source
+          {
+            path: this.resultUrl,
+            username: this.form.value.credentials!.username || '',
+            password: this.form.value.credentials!.password || '',
+            revision: this.form.value.revision!,
+            entrypoint: this.form.value.entrypoint || '',
+          } as Source
```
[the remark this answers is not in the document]
dominik003: The problem with this is that the actual path is not part of the form, as it is a combination of the base URL (if required) and the input URL.

frontend/src/app/projects/models/create-coworking-method/create-coworking-method.component.html
```diff
@@ -8,14 +8,93 @@
...
+            <input
+              matInput
+              (ngModelChange)="onInputChange($event)"
+              formControlName="inputUrl"
+            />
+            <mat-error
+              *ngIf="
+                form.controls.urlGroup.controls.inputUrl.errors
+                  ?.urlPrefixError &&
+                form.controls.urlGroup.controls.inputUrl.value
```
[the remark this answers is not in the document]
dominik003: The check is used here because an empty value has no valid prefix and thus the url prefix error is present. However, it does not make sense to print the error in this case, especially because the input field is not required.

frontend/src/app/projects/models/create-coworking-method/create-coworking-method.component.html
```diff
@@ -8,14 +8,93 @@
...
+              "
+            >
+              {{
+                form.controls.urlGroup.controls.inputUrl.errors?.urlPrefixError
+              }}
+            </mat-error>
```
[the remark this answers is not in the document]
dominik003: The input field for the url is not required if you simply use the selected base url for the git source.

backend/capellacollab/extensions/modelsources/git/routes.py
```diff
@@ -161,25 +161,28 @@ def get_revisions(
...
     remote_refs["default"] = git_model.revision
 
-    log.debug("Determined branches: %s", remote_refs["branches"])
-    log.debug("Determined tags: %s", remote_refs["tags"])
     log.debug("Determined default branch: %s", remote_refs["default"])
+
     return remote_refs
+
+
+@router.post("/revisions", response_model=GetRevisionsModel)
```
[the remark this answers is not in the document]
dominik003: From a semantic point of view, it would make the most sense to use a `GET` here. However, the problem is that the route gets the password in the body of the request and the body of a `GET` should have no meaning. Therefore it is best to use a `POST`, but in the future we might want to consider using `QUERY` https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html (if it is then supported in fast API).

frontend/src/app/services/settings/git-settings.service.ts
```diff
@@ -74,7 +92,7 @@ export class GitSettingsService {
   }
```
MoritzWeber0, line 92: Lines 82-91 are unreachable

frontend/src/app/settings/modelsources/git-settings/git-settings.component.ts
```diff
@@ -18,60 +28,92 @@ import { DeleteGitSettingsDialogComponent } from 'src/app/settings/modelsources/
...
+
   constructor(
     private navbarService: NavBarService,
     private gitSettingsService: GitSettingsService,
     public dialog: MatDialog,
     public dialogRef: MatDialogRef<DeleteGitSettingsDialogComponent>
   ) {
     this.navbarService.title = 'Settings / Modelsources / Git';
-    this.instances = [];
+    this.cmpGitSettings = [];
```
MoritzWeber0, line 52: You can directly set it during the declaration above.

frontend/src/app/settings/modelsources/git-settings/git-settings.component.ts
```diff
@@ -18,60 +28,92 @@ import { DeleteGitSettingsDialogComponent } from 'src/app/settings/modelsources/
   templateUrl: './git-settings.component.html',
   styleUrls: ['./git-settings.component.css'],
 })
-export class GitSettingsComponent implements OnInit {
-  public instances: GitSettings[];
-  gitSettingsForm = new FormGroup({
+export class GitSettingsComponent implements OnInit, OnDestroy {
+  public cmpGitSettings: GitSetting[];
```
MoritzWeber0, line 32: What does `cmp` stand for in this context?

backend/capellacollab/sql_models/extensions.py
```diff
@@ -9,10 +9,7 @@
 log = logging.getLogger(__name__)
 
 # Load extension models
-eps = (
-    metadata.entry_points()["capellacollab.extensions.backups"]
-    + metadata.entry_points()["capellacollab.extensions.modelsources"]
-)
+eps = metadata.entry_points()["capellacollab.extensions.backups"]
```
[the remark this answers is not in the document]
dominik003: The only place it has not been removed is the `pyproject.toml`, or are there other places?

frontend/src/app/projects/models/create-coworking-method/create-coworking-method.component.ts
```diff
@@ -3,67 +3,180 @@
...
 
 @Component({
   selector: 'app-create-coworking-method',
   templateUrl: './create-coworking-method.component.html',
   styleUrls: ['./create-coworking-method.component.css'],
 })
-export class CreateCoworkingMethodComponent implements OnInit {
-  @Output() create = new EventEmitter<boolean>();
+export class CreateCoworkingMethodComponent implements OnInit, OnDestroy {
+  @Output() create = new EventEmitter<{ created: boolean }>();
```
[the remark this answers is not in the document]
dominik003: Not sure what happened there, but agree that we should just use a bool.

frontend/src/app/settings/modelsources/git-settings/git-settings.component.ts
dominik003: It stands for component, but I also renamed it because I wasn't happy with the name myself

dominik003:
The additional comments are addressed and resolved in #311, as this branch is far behind and therefore there is no point in working on it.

backend/capellacollab/sql_models/extensions.py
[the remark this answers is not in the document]
MoritzWeber0: Another reference is in `capellacollab/extensions/__init__.py`. The whole file content can be deleted.

PR #209, opened 2022-08-31
Alembic fix migration history

TimoBbz:
# Description

The Alembic migration files were not consistent. This PR unifies the database state given by the models files and the state computed by the migration files on top of the models files of the staging branch.

# Testing

# Checklist

- [x] My code follows the guidelines of this project: [CONTRIBUTING.md](/CONTRIBUTING.md)
- [x] I updated the documentation with the new functionality
- [x] I went through the code and removed all print statements, breakpoints and unused code
- [x] Error handling for all possible scenarios has been implemented
- [x] I've add logging for all relevant operations
- [x] I added the appropriate labels to this PR (enhancement/bug/documentation, effort, priority)

PR #285, opened 2022-10-07
Fix alembic migration history

MoritzWeber0:
# Description

Depends on #267 and extends #209. It fixes some migration issues.

# Testing
With our new test from #267.


MoritzWeber0:
Is continued in #285.

backend/capellacollab/__main__.py
```diff
@@ -31,7 +31,6 @@ def filter(self, record: logging.LogRecord) -> bool:
 
 logging.getLogger("uvicorn.access").addFilter(HealthcheckFilter())
 log.addFilter(HealthcheckFilter())
-logging.getLogger(__name__).setLevel(config["logging"]["level"])
```
MoritzWeber0, line 34: This line is still needed if the Collaboration Manager is started from a context that is not alembic.

backend/capellacollab/alembic/versions/d888f5c30aab_add_model.py
```diff
@@ -101,7 +104,7 @@ def upgrade():
             slug = base_slug
             index = 0
 
-            while slug in existing_slugs:
+            while slug in existing_slugs[project_id]:
```
amolenaar, line 107: What is this doing? A comment is desired.

amolenaar (approved):
One bit of code is unclear. Apart from that it's a nice improvement.

backend/capellacollab/alembic/versions/d888f5c30aab_add_model.py
MoritzWeber0: It is the migration of the old databases. Per project, the slug of the model has to be unique. This was not a constraint in the old version, therefore there are conflicts (and we need to assign new slugs, adding just `-\d` to the existing "slugs").

backend/capellacollab/alembic/env.py
```diff
@@ -15,24 +17,28 @@
...
     config.set_main_option("sqlalchemy.url", cfg["database"]["url"])
 
-# Interpret the config file for Python logging.
-# This line sets up loggers basically.
-if config.attributes.get("configure_logger", True):
-    fileConfig(config.config_file_name)
-
 # Import models
-import capellacollab.sql_models  # isort:skip
-from capellacollab.sql_models import extensions  # isort:skip
```
TimoBbz, line 30: Why not importing the models anymore there?

TimoBbz (approved):

PR #300, opened 2022-10-11
Fix unknown state in sessions

MoritzWeber0:
# Description

This PR fixes the issue with "Unknown" session states. Usually, we fetch the events from kubernetes to determine a detailed state for a Pod. With events, we can get information like "Image pull failed", etc. 

However, events are only stored for a limited time. Therefore, after a specific time, the sessions goes into state "unknown".
I added a fallback to the Pod phase, which should be set always: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
It has less information (e.g. just "Failed", not the reason for the failure), but it's better than nothing and should be pretty stable.

### DIFF
--- a/frontend/src/app/sessions/active-sessions/active-sessions.component.html
+++ b/frontend/src/app/sessions/active-sessions/active-sessions.component.html
@@ -24,9 +24,9 @@
   <mat-card-content class="sessionContent">
     <h3
       class="state"
-      [ngClass]="sessionService.beatifyState(session.state).css"
+      [ngClass]="sessionService.beautifyState(session.state).css"
     >
-      {{ sessionService.beatifyState(session.state).text }}
+      {{ sessionService.beautifyState(session.state).text }}
     </h3>
     <p id="creationTime">
       The session was created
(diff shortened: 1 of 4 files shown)


frontend/src/app/sessions/active-sessions/active-sessions.component.html
```diff
@@ -54,9 +54,7 @@ <h3>{{ sessionService.beatifyState(session.state).text }}</h3>
           mat-flat-button
           color="primary"
           (click)="openReconnectDialog(session)"
-          [disabled]="
-            !(session.state === 'Started' || session.state === 'START_SESSION')
-          "
+          [disabled]="!sessionService.beatifyState(session.state).success"
```
dominik003, line 57: I am not sure if this function should really be called `beatify` and not `beautify`. Also, there is a similar function in easebackup.service.ts that should be deleted if it is deprecated, or the function names should be updated to distinguish between the two.

MoritzWeber0: I have renamed it. I'll rework the whole EASE Backup component in another PR, I'll get rid of it there.

dominik003 (approved):

PR #303, opened 2022-10-11
Introduce ESLint

### LINKED ISSUE #243: Add ESLint pre-commit hook
We should enforce the use of [ESLint](https://eslint.org/) in a pre-commit hook. 
Before implementing the hook, all ESLint issues have to be fixed.

MoritzWeber0:
# Description

- Add ESLint with default Angular configuration, sort imports and sort ngArray enabled.
- I ran `npx eslint --fix`, some file were fixed.

Resolves #243


frontend/src/app/workspace/workspace.component.ts
```diff
@@ -29,18 +29,4 @@ export class WorkspaceComponent implements OnInit {
...
-    // this.projectUserService.subscribe({
-    //   next: (res) => {
-    //     this.repositories = res;
-    //     this.showSpinner = false;
-    //   },
-    //   error: () => {
-    //     this.showSpinner = false;
-    //   },
-    // });
-  }
```
dominik003, line 45: Should this actually be removed because there is still a FIXME in it?

MoritzWeber0: In my opinion, it's a better way to track general issues in Github issues and use FIXME only for local markers, which are removed before merging. Otherwise there is the issue that FIXMEs and TODOs sum up and nobody cares about it anymore.

dominik003 (approved):

PR #308, opened 2022-10-13
Linked Git Models Overview

dominik003:
# Description

With this PR, the existing git models (for a model in a project) are included in the model overview. In addition, the routes to edit an existing git model or to create a new model are included.

# Checklist

- [x] My code follows the guidelines of this project: [CONTRIBUTING.md](/CONTRIBUTING.md)
- [x] I updated the documentation with the new functionality
- [x] I went through the code and removed all print statements, breakpoints and unused code
- [x] Error handling for all possible scenarios has been implemented
- [x] I've add logging for all relevant operations
- [x] I added the appropriate labels to this PR (enhancement/bug/documentation, effort, priority)

PR #311, opened 2022-10-14
Edit Linked Git Models

dominik003:
# Description

This PR extends #308 with the actual functionality to edit and create new git integrations for an existing model. The following functionalities are included:

- Overview of existing git models for a capella model (with all information) (already added in #308)
- Ability to add a new git model to an existing capella model (with the same behavior as adding during model creation)
- Ability to edit an existing git model. Everything can be changed here and non-primary models can be made primary. If no username and password are specified, the password is removed, if a username but no password is specified, the old password is kept, and if both are specified, the username and password are updated

# Checklist


backend/capellacollab/projects/capellamodels/modelsources/git/routes.py
```diff
@@ -70,22 +69,44 @@ def verify_valid_path_sequences(path: str):
...
+        if git_model.id == git_model_id:
+            return
+
+    return HTTPException(
+        status_code=400,
+        detail={
+            "err_code": "git_model_not_exists_on_project_and_model",
+            "reason": "The git model does not exists on the capella model for the project",
+        },
+    )
```
TimoBbz, line 89: I would have used an injectable function that loads a model (https://fastapi.tiangolo.com/tutorial/dependencies/). For example:

```python
def load_capella_model(
  db: Session), project_slug: str, model_slug: str
) -> DatabaseCapellaModel:
  # the code, with a HTTPException if needed
  return capella_model


def load_git_model(
  db: Session, capella_model: DatabaseCapellaModel, git_model_id: int
) -> DatabaseGitModel:
  # the code, with a HTTPException if needed
  return git_model


@router.get()
def route_handling(
  capella_model: DatabaseCapellaModel = Depends(load_capella_model),
  git_model: DatabaseGitModel = Depends(load_git_model)
) :
  # the code
```
I’m not sure of the dependency between the different functions but that’s the idea.

backend/capellacollab/projects/capellamodels/routes.py
```diff
@@ -49,25 +69,16 @@ def list_in_project(
     ]
 
 
-@router.get("/{slug}", response_model=ResponseModel)
+@router.get("/{model_slug}", response_model=ResponseModel)
 def get_model_by_slug(
     project_slug: str,
-    slug: str,
+    model_slug: str,
```
TimoBbz, line 75: No more in use since you load the model and the project in the dependencies.

backend/capellacollab/projects/capellamodels/routes.py
```diff
@@ -114,17 +125,9 @@ def set_tool_details(
     db: Session = Depends(get_db),
     token: JWTBearer = Depends(JWTBearer()),
     project=Depends(get_existing_project),
-):
+    model=Depends(get_existing_model),
```
TimoBbz, line 128: You can remove the `model_slug: str` above for the same reason.

frontend/src/app/helpers/validators/validators.ts
```diff
@@ -24,3 +24,18 @@ export function regExpValidator(
...
+
+    if (value.endsWith('.aird')) {
+      return null;
+    }
+
+    return {
+      cappellaSuffixError: `${value} must end with ".aird" in case of a capella model`,
+    };
+  };
+}
```
TimoBbz, line 41: Could this not just be handled with a RegexpValidator? `\^$|\.aird$\`. should work I think

TimoBbz:
When I:
- Go to a model
- Click on source creation
- Go back
- Click on another source
The source doesn’t load as long as I don’t reload the page. This might be an issue of completed subject.

frontend/src/app/projects/models/create-coworking-method/create-coworking-method.component.ts
```diff
@@ -156,52 +197,102 @@ export class CreateCoworkingMethodComponent implements OnInit {
...
+          this.projectService.project?.slug!,
+          this.modelService.model?.slug!,
+          this.createGitModelFromForm()
         )
         .subscribe(() => {
-          this.create.emit({ created: true });
+          if (this.asStepper) {
+            this.create.emit({ created: true });
+          } else {
+            this.location.back();
```
TimoBbz, line 227: The intended behaviour is not going back but redirecting to the previous page. This leads to unwanted consequences such as the possibility to go forward with the browser (when you would have liked to return to the page with the back arrow of your browser).

Use instead `this.router.navigate(['..'], { relativeTo: this.route });`.

frontend/src/app/projects/project-detail/model-overview/model-detail/model-detail.component.html
```diff
@@ -3,4 +3,43 @@
...
+    >
+      <mat-card matRipple class="mat-card-overview">
+        <div class="header">Integration {{ gitModel.id }}</div>
+        <div class="content">
+          <div>Path: {{ gitModel.path }}</div>
+          <div>Revision: {{ gitModel.revision }}</div>
+          <div>Entry point: {{ gitModel.entrypoint }}</div>
+          <div>Primary: {{ gitModel.primary ? "True" : "False" }}</div>
+          <div>Username: {{ gitModel.username }}</div>
+          <div>Passwort set: {{ gitModel.password ? "True" : "False" }}</div>
```
TimoBbz, line 40: typo.

TimoBbz:
When I enter a private repository as repository, my backend asks me for git credentials. I don’t think this is a desired behavior. Also, there should be a specific error raised when the credentials are wrong when fetching the branches and tags.

You can make some tests with https://github.com/TimoBbz/example_branch_and_tags, it’s private

TimoBbz:
I also think that issues related to unreachable repositories or inexisting branches should not lead to errors but to warnings. One could want to create a branch later, a Git instance could be offline, etc.
But this is more a feature choice than an error, so perhaps is it out of this scope.

dominik003:
> When I:
> 
>     * Go to a model
> 
>     * Click on source creation
> 
>     * Go back
> 
>     * Click on another source
> 
> 
> The source doesn’t load as long as I don’t reload the page. This might be an issue of completed subject.

When I started applying your changes and some more, I also encountered this problem. However, it seems to be fixed now (at least on my side) maybe it had something to do with the `location.back()` instead of using the router.

dominik003:
> When I enter a private repository as repository, my backend asks me for git credentials. I don’t think this is a desired behavior. Also, there should be a specific error raised when the credentials are wrong when fetching the branches and tags.
> 
> You can make some tests with https://github.com/TimoBbz/example_branch_and_tags, it’s private

Somehow I forgot to move `askpass.py` when I moved the function to retrieve revisions. So if you try to access a private repository, you should no longer be prompted to enter your credentials in vs code. However, it is important to note that for a private Github repository, you cannot use your Github password, but must create an access token with repository privileges and use it

dominik003:
> I also think that issues related to unreachable repositories or inexisting branches should not lead to errors but to warnings. One could want to create a branch later, a Git instance could be offline, etc. But this is more a feature choice than an error, so perhaps is it out of this scope.

I totally agree with you, this is actually something I wanted to include as well. The problem here is that we are using python sub-processes to run `ls-remote` which unfortunately gives the same response if the credentials are wrong, the repository doesn`t exist, etc.

TimoBbz:
When I:
- Create an instance
- Get redirected to the list of the models details
- Click on any instance
The instance isn’t loaded.

backend/capellacollab/projects/capellamodels/modelsources/git/crud.py
```diff
@@ -33,58 +36,65 @@ def get_gitmodel_by_id(db: Session, id: int) -> DB_GitModel:
...
 
 
-def add_gitmodel_to_capellamodel(
-    db: Session, capellamodel_id: int, model: PostGitModel
+def delete_model_from_repository(
+    db: Session, capellamodel_id: int, model_id: int
 ):
-    if len(get_gitmodels_of_capellamodels(db, capellamodel_id)):
+    db.query(DB_GitModel).filter(DB_GitModel.id == model_id).filter(
+        DB_GitModel.model_id == capellamodel_id
```
TimoBbz, line 56: Why do you need to filter the model of the git model? The behaviour here is to fail silently in case there is no model matching, I’m not sure that’s the desired behaviour.

dominik003: I haven't actually made any changes here, but somehow (I suspect due to changing the order of some functions) it shows up here. But yes, you are right that this is probably not the desired behavior. Since the function is not needed at the moment and is just a simple crud function, I removed it so we don't just accidentally use it at a later point.

backend/capellacollab/projects/capellamodels/modelsources/git/models.py
```diff
@@ -88,34 +73,16 @@ class DB_GitModel(Base):
...
-            primary=True,
+            path=new_model.path,
+            entrypoint=new_model.entrypoint,
+            revision=new_model.revision,
+            primary=primary,
             model_id=model_id,
-            username=source.username,
-            password=source.password,
+            username=new_model.username,
+            password=new_model.password,
```
TimoBbz, line 87: This could definitely use a structure like this for more concision (That’s not a bug tho).
```python
return cls(
  name="",
  primary=primary,
  model_id=model_id,
  **new_model.dict(),
)
```

backend/capellacollab/projects/capellamodels/modelsources/git/models.py
```diff
@@ -13,62 +13,47 @@
...
+    id: int
+    name: str
     path: str
     entrypoint: str
     revision: str
+    primary: bool
     username: str
-    password: str
-
+    password: bool
```
TimoBbz, line 50: I think the path is not validated there. Also this could inherit from the PostGitModel, like this:
```python
class ResponseGitModel(PostGitModel):
    id: int
    name: str
    primary: bool
    password: bool

    @validator("password", pre=True)
    def transform_password(cls, passw: t.Union[str, bool]) -> bool:
        if isinstance(passw, bool):
            return passw
        return passw is not None and len(passw) > 0

    class Config:
        orm_mode = True
```

dominik003: Do we need to validate the path for the response model? To me, the response model is only ever used to return data and should not need to do any validations at all (here the validator is simply used to transfom the password. Oddly enough, pydantic has no transfomation decorator and simply refers to using the validator).
And how is the case handled where the `PostGitModel` has `str` as the password type and the `ResponseGitModel` has `bool`. Does the response model simply override the boolean value from the post model?

TimoBbz: 1. Yes you’re right
2. I don’t know, I tried and it works but I’m not sure what it does under the hood.

That makes two good reasons to leave it like it is.

TimoBbz (approved):

PR #330, opened 2022-10-24
Refactor role verification and routes

### LINKED ISSUE #310: Use project slug instead of project name for all routes
In some routes, there is still the project name used. We should change it to the slug in the routes.

dominik003:
# Description

This PR replaces the use of crud operations (e.g. to get a project) with the use of injectables (where possible and appropriate). As a result, several crud operations have also been updated so that they do not get a name, slug or ID, but directly the database object they are supposed to work with. To be able to use the injectables, the previous use of the project name has been replaced with the project slug in most places (which also resulted in some frontend route and naming changes).
In addition, most of the role and project role verification within the code has been changed to dependency injection and is now only part of the code when needed (e.g.., in conditions). Finally, there are some minor code improvements.

Resolves #310 

# Checklist


backend/capellacollab/settings/modelsources/git/injectables.py
```diff
@@ -0,0 +1,25 @@
...
+
+from fastapi import Depends, HTTPException
+from sqlalchemy.orm import Session
+
+import capellacollab.settings.modelsources.git.crud as crud
+from capellacollab.core.database import get_db
+from capellacollab.settings.modelsources.git.models import DB_GitSettings
+
+
+def get_existng_git_setting(
```
TimoBbz, line 12: typo

backend/capellacollab/settings/modelsources/git/routes.py
```diff
@@ -1,62 +1,87 @@
...
-            detail={
-                "reason": f"The git instance with id {id} does not exist."
-            },
-        )
-
-
-@router.get("/")
+@router.get(
+    "/",
+    response_model=t.List[GitSettingsGitGetResponse],
```
TimoBbz, line 31: Rather use `list[GitSettingsGitGetResponse]` with `from __future__ import annotations` at the beginning of the file than `t.List[GitSettingsGitGetResponse]`.

TimoBbz:
Some dependencies can be moved directly to the router. For exemple, since every route in the model router checks the project role to be user, the model router itself could check it in its declaration.

Also, there is no need to convert the SQLAlchemy models to Pydantic models, since FastAPI handles the trivial cases with the `response_model` decorator parameter.

backend/capellacollab/projects/routes.py
```diff
@@ -52,127 +54,100 @@
...
-                [
-                    user
-                    for user in project.users
-                    if user.role == ProjectUserRole.USER
-                    and user.permission == ProjectUserPermission.READ
-                ]
-            ),
-        ),
-    )
+    return True
```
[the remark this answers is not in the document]
dominik003: I changed it back to None, since we always return `None` when a database object is deleted.

backend/capellacollab/settings/modelsources/git/routes.py
dominik003: I changed it to `list` here and in another route.

backend/capellacollab/users/routes.py
```diff
@@ -103,7 +100,9 @@ def get_sessions_for_user(
     db: Session = Depends(get_db),
     token=Depends(JWTBearer()),
 ):
-    if user.name != get_username(token) and not is_admin(token, db):
+    if user.name != get_username(token) and not RoleVerification(
+        required_role=Role.ADMIN, verify=False
+    )(token, db):
```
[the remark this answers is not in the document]
dominik003: Done and additionally went over the entire route again and added injectables in several other places as well.

dominik003:
> Some dependencies can be moved directly to the router. For exemple, since every route in the model router checks the project role to be user, the model router itself could check it in its declaration.
> 
> Also, there is no need to convert the SQLAlchemy models to Pydantic models, since FastAPI handles the trivial cases with the `response_model` decorator parameter.

I added the ProjectRoleVerification with the minimal project role `USER` to the project model routes. As a result, I had to move the `include_route` for the git and t4c model routes to the capella model route (but for me, that is also the right place to include them). As a result, only role checks for `MANAGER` and `ADMIN` need to be added now for all routes of the form `/{project_slug}/models/{model_slug}`.
In addition, I removed all legacy `from_orm` calls, since this is automatically applied based on the `response_model`.

backend/capellacollab/projects/capellamodels/routes.py
```diff
@@ -23,57 +22,57 @@
...
     db: Session = Depends(get_db),
-    token=Depends(JWTBearer()),
-) -> t.List[ResponseModel]:
-    verify_project_role(project.name, token, db)
-    return [
-        ResponseModel.from_orm(model)
-        for model in crud.get_all_models_in_project(db, project.slug)
-    ]
+) -> list[DatabaseCapellaModel]:
+    return crud.get_all_models_in_project(db, project)
```
TimoBbz, line 52: You could actually return `project.models`, isn’t it?

backend/capellacollab/projects/capellamodels/routes.py
```diff
@@ -84,58 +83,66 @@ def create_new(
                 "technical": "Slug already used",
             },
         )
-    return ResponseModel.from_orm(model)
+    return model
```
TimoBbz, line 86: The return could be directly in the try:

backend/capellacollab/projects/capellamodels/routes.py
dominik003: This should already be done here.

backend/capellacollab/projects/capellamodels/routes.py
dominik003: True, I moved it directly into the try block

TimoBbz (approved):

PR #337, opened 2022-10-26
Improve deploy experience

MoritzWeber0:
# Description
  - Remove workspaces images
  - Remove mocks
  - Add new option in `values.yaml` for oauth mock
  - Add target "other", which is the default (prevents using local in prod)
  - Add documentation for some variables
  - Remove `service.fqdn` (can be generated from general.host, general.port, general.scheme)
  - Use download Capella functionality from capella dockerimages
  - Hide come commands to the user (makes it cleaner)
  - Combine `kubectl wait` and `kubectl get --watch` with background process (only show  `kubectl get --watch` until all containers are ready)


Makefile
```diff
@@ -27,9 +28,9 @@ frontend:
 	docker build --build-arg CONFIGURATION=local -t t4c/client/frontend -t $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend frontend
 	docker push $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend
 
-capella: capella-download
+capella:
```
amolenaar, line 31: Is downloading capella done in the docker image?

Makefile
```diff
@@ -137,21 +118,26 @@ create-cluster:
...
 	k3d cluster list $(CLUSTER_NAME) 2>&- && k3d cluster delete $(CLUSTER_NAME)
 
+.ONESHELL:
 wait:
 	@echo "-----------------------------------------------------------"
 	@echo "--- Please wait until all services are in running state ---"
 	@echo "-----------------------------------------------------------"
-	@kubectl get -n $(NAMESPACE) --watch pods
-
-.provision-guacamole:
```
amolenaar, line 146: The idea was to `touch .provision-guacamole` at the end of this step. I suppose that has disappeared.

amolenaar (changes requested):
All in all a nice bunch of cleanup.

However, when I set up a local k3d environment I end up with a backend that's unable to communicate with gracamole.

Have you tried on a fresh local container?

Makefile
MoritzWeber0: Yes :) 
It's a new feature of `capella-dockerimages`, so we don't need to do it manually anymore.

Makefile
MoritzWeber0: `touch .provision-guacamole` was a good idea until we dropped the persistent volume for the guacamole database. Every time the database Pod got restarted (which happens on new rollout and can happen after a new deployment when there were changes), the database structure was gone (no persistence). Therefore, I changed it to the "deploy always" mode. When the database structures are already there, postgres skips the creation.

amolenaar (approved):

PR #350, opened 2022-10-27
Make model tool required

dominik003:
# Description

This fixes the problem that the model tool was not marked as required in the frontend. It also includes some code quality improvements.

### DIFF
--- a/frontend/src/app/projects/models/create-model-base/create-model-base.component.ts
+++ b/frontend/src/app/projects/models/create-model-base/create-model-base.component.ts
@@ -35,7 +35,10 @@ export class CreateModelBaseComponent implements OnInit {
   public form = new FormGroup({
     name: new FormControl('', [Validators.required, this.slugValidator()]),
     description: new FormControl(''),
-    tool_id: new FormControl(-1, this.validToolValidator()),
+    toolID: new FormControl<number | undefined>(undefined, [
+      Validators.required,
+      this.validToolValidator(),
+    ]),
   });
 
   constructor(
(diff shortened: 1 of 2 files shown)


frontend/src/app/projects/models/create-model-base/create-model-base.component.ts
```diff
@@ -35,7 +35,10 @@ export class CreateModelBaseComponent implements OnInit {
   public form = new FormGroup({
     name: new FormControl('', [Validators.required, this.slugValidator()]),
     description: new FormControl(''),
-    tool_id: new FormControl(-1, this.validToolValidator()),
+    toolID: new FormControl<number | undefined>(undefined, [
```
TimoBbz, line 38: Changing the name of the field requires to change the model field (or to assign it manually in the submit), and also the field name in the back-end (and finally breaking the python naming convention). I’d advise to keep it as it was for more simplicity.

dominik003: I actually overlooked that the name is required at the moment to transform the form value to a `NewModel`. But in my opinion, a name chain from the backend model to the form name in the frontend reduces maintainability and also violates against our naming scheme in the frontend. 
For me it makes a lot of sense to have the `tool_id` in the `NewModel`, because this way we can use it directly with the backend without transformations. But personally I would not put the naming further into the frontend components. For example, here we could just use:
```
this.modelService.createNewModel(this.projectService.project.slug, {
          name: this.form.value.name,
          description: this.form.value.description,
          tool_id: this.form.value.toolID,
        } as NewModel),
```
,which on one hand is not as simple as the current solution, but on the other hand removes the restriction to the form name.

TimoBbz (approved):

PR #355, opened 2022-10-31
Remove push images feature

MoritzWeber0:
# Description

This makes use of the new "push images" feature of Capella Dockerimages. We can remove the functionality here and reuse the feature from Capella Dockerimages.


Makefile
```diff
@@ -11,12 +11,16 @@ SESSION_NAMESPACE = t4c-sessions
...
 PORT ?= 8080
 
+CAPELLA_DOCKERIMAGES = $(MAKE) -C capella-dockerimages PUSH_IMAGES=1 LOCAL_REGISTRY_NAME=$(LOCAL_REGISTRY_NAME) LOCAL_REGISTRY_PORT=$(REGISTRY_PORT)
+
 # Adds support for msys
 export MSYS_NO_PATHCONV := 1
 
-build: backend frontend docs capella/remote readonly
+build: backend frontend docs
+	$(CAPELLA_DOCKERIMAGES) capella/remote capella/readonly
```
amolenaar, line 20: I would spell it out, and not use an environment variable for an executable with arguments. (for readability mostly)

amolenaar (approved):
Nice bit of cleanup

Makefile
MoritzWeber0: Two reasons against it: 
1. The focus should be on the images that are built during in the step. With the current version, you can see them directly. With the new version, you have to read the whole (long) command first.
2. It's a long command and used in two different places. Would be code duplication. 

What do you think about these two points?

MoritzWeber0: Maybe we can find another name for it like `$(CAPELLA_DOCKERIMAGES_MAKE_COMMAND)`.

amolenaar (commented):
One remark: for local development on K3S cluster, we need those images available. They can't be downloaded from the local docker server, for K3S a separate registry is spawned.

How will this work with the local K3S environment if the images are no longer accessible locally?

**Edit**: this comment is invalid. It was based on the premise that the images are not uploaded to the local docker registry.

MoritzWeber0:
> One remark: for local development on K3S cluster, we need those images available. They can't be downloaded from the local docker server, for K3S a separate registry is spawned.
> 
> How will this work with the local K3S environment if the images are no longer accessible locally?

I don't quite get the point. Images are still available in the k3d cluster - they are still pushed after build. Capella Dockerimages has now the push functionality and we provide the registry with the Make variable.

Makefile
amolenaar: I see, that's defined as part of the `CAPELLA_DOCKERIMAGES` env var. :+1:

PR #372, opened 2022-11-11
Solve askpass issues

TimoBbz:
# Description

Askpass was not taken in account because it’s shebang contained `python` instead of `python3`. I also needed to make it executable


dominik003 (approved):

MoritzWeber0 (approved):
LGTM, also follows the official documentation: https://docs.python.org/3/using/windows.html#customizing-default-python-versions

PR #375, opened 2022-11-11
Fix small migration issues

TimoBbz:
…mpty string.

# Description

On tools creation, models now have a default tool. Also, the default value of git_models.username is set on empty string on migration.

### DIFF
--- a/backend/capellacollab/core/database/migration.py
+++ b/backend/capellacollab/core/database/migration.py
@@ -116,6 +116,7 @@ def create_tools(db):
     tools.create_nature(db, papyrus.id, "SysML 1.1")
 
     for model in models.get_all_models(db):
+        models.set_tool_for_model(db, model, capella)
         models.set_tool_details_for_model(
             db, model, default_version, default_nature
         )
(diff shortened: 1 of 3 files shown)


backend/capellacollab/core/database/migration.py
```diff
@@ -86,16 +87,21 @@ def create_tools(db):
...
-    tools.create_nature(db, capella.id, "model")
+    default_nature = tools.create_nature(db, capella.id, "model")
     tools.create_nature(db, capella.id, "library")
 
     tools.create_nature(db, papyrus.id, "UML 2.5")
     tools.create_nature(db, papyrus.id, "SysML 1.4")
     tools.create_nature(db, papyrus.id, "SysML 1.1")
+
+    for model in models.get_all_models(db):
+        models.set_tool_details_for_model(
```
MoritzWeber0, line 105: This doesn't set the tool (only the version and nature). Still get the same issue with the test database: 
```
pydantic.error_wrappers.ValidationError: 3 validation errors for ResponseModel
response -> 0 -> tool
```

TimoBbz: Fixed normally

MoritzWeber0 (approved):

MoritzWeber0 (approved):

PR #376, opened 2022-11-14
Build and use multiple capella versions

amolenaar:
# Description

Build multiple capella versions as part of the local deployment. 

* Update capella-dockerimages to the latest version.
* Update Makefile to work with the latest version of capella-dockerimages.
* Align naming with what's found in capella-dockerimages.
* Use standard capella images (not t4c/client), since those are built as part of `make deploy` (the new user experience)

# Testing

Clean everything, including docker registry. Rebuild everything and start capella 6.0 through the collab manager

# Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,8 +9,10 @@ RELEASE = dev-t4c-manager
 NAMESPACE = t4c-manager
 SESSION_NAMESPACE = t4c-sessions
 PORT ?= 8080
-CAPELLA_VERSIONS = 5.0.0 5.2.0 6.0.0
-T4C_CLIENT_VERSIONS = 5.2.0 6.0.0
+# List of Capella versions, e.g.: `5.0.0 5.2.0 6.0.0`
+CAPELLA_VERSIONS = 5.2.0
+# List of T4C versions, e.g., `5.2.0 6.0.0`
+T4C_CLIENT_VERSIONS = 5.2.0
 CAPELLA_DOCKERIMAGES = $(MAKE) -C capella-dockerimages PUSH_IMAGES=1 DOCKER_REGISTRY=$(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)
 
 # Adds support for msys
--- a/Makefile
+++ b/Makefile
@@ -9,16 +9,13 @@ RELEASE = dev-t4c-manager
 NAMESPACE = t4c-manager
 SESSION_NAMESPACE = t4c-sessions
 PORT ?= 8080
-
+CAPELLA_VERSIONS = 5.0.0 5.2.0 6.0.0
 CAPELLA_DOCKERIMAGES = $(MAKE) -C capella-dockerimages PUSH_IMAGES=1 DOCKER_REGISTRY=$(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)
 
 # Adds support for msys
 export MSYS_NO_PATHCONV := 1
 
-build: backend frontend docs
-
-build-capella:
-	$(CAPELLA_DOCKERIMAGES) capella/remote capella/readonly
+build: backend frontend docs capella
 
 build-t4c: build
 	$(CAPELLA_DOCKERIMAGES) t4c/client/remote t4c/client/backup
@@ -32,6 +29,14 @@ frontend:
 	node frontend/fetch-version.ts
 	docker build --build-arg CONFIGURATION=local -t t4c/client/frontend -t $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend frontend
 	docker push $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend
+
+capella:
+	for version in $(CAPELLA_VERSIONS)
+	do $(MAKE) capella-$$version
+	done
+
+capella-%: registry
+	$(CAPELLA_DOCKERIMAGES) CAPELLA_VERSION=$* capella/remote capella/readonly
 
 docs:
 	docker build -t capella/collab/docs -t $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/docs docs/user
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -32,6 +31,11 @@ frontend:
 	docker build --build-arg CONFIGURATION=local -t t4c/client/frontend -t $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend frontend
 	docker push $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend
 
+capella: capella-5.0.0 capella-5.2.0 capella-6.0.0
```
MoritzWeber0, line 34: I'd not add it as default - The whole setup is too much for most notebooks. Maybe an Make variable for the CAPELLA_VERSION(S)?

MoritzWeber0 (changes requested):

Makefile
amolenaar: I added a Make variable with to-be built versions.

MoritzWeber0 (commented):
LGTM, but we have two conflicting files. Can you rebase @amolenaar?

backend/capellacollab/core/database/migration.py
```diff
@@ -75,8 +75,7 @@ def create_tools(db):
     registry = config["docker"]["registry"]
     capella = Tool(
         name="Capella",
-        docker_image_template=f"{registry}/t4c/client/remote:$version-latest",
-        docker_image_backup_template=f"{registry}/t4c/client/backup:$version-latest",
+        docker_image_template=f"{registry}/capella/remote:$version-latest",
```
MoritzWeber0, line 78: Any reason for this change? We use the T4C images for the persistent workspaces in both instances. And the deletion of the `docker_image_backup_template` will break the backups integration.

Makefile
```diff
@@ -9,19 +9,16 @@ RELEASE = dev-t4c-manager
 NAMESPACE = t4c-manager
 SESSION_NAMESPACE = t4c-sessions
 PORT ?= 8080
-
+CAPELLA_VERSIONS = 5.0.0 5.2.0 6.0.0
+T4C_CLIENT_VERSIONS = 5.2.0 6.0.0
```
MoritzWeber0, line 13:
```suggestion
# List of Capella versions, e.g.: `5.0.0 5.2.0 6.0.0`
CAPELLA_VERSIONS = 5.2.0
# List of T4C versions, e.g., `5.2.0 6.0.0`
T4C_CLIENT_VERSIONS = 5.2.0
```

The default behavior should not change. The whole building process takes a long time and it's high hurdle for new users.

Makefile
```diff
@@ -9,19 +9,16 @@ RELEASE = dev-t4c-manager
...
 CAPELLA_DOCKERIMAGES = $(MAKE) -C capella-dockerimages PUSH_IMAGES=1 DOCKER_REGISTRY=$(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)
 
 # Adds support for msys
 export MSYS_NO_PATHCONV := 1
 
-build: backend frontend docs
-
-build-capella:
-	$(CAPELLA_DOCKERIMAGES) capella/remote capella/readonly
+build: backend frontend docs capella
```
MoritzWeber0, line 19:
```suggestion
build: backend frontend docs
```

I'd not include it here. At least, we should have a separate target to build images except the Capella Docker images.

Makefile
```diff
@@ -33,6 +30,22 @@ frontend:
 	docker build --build-arg CONFIGURATION=local -t t4c/client/frontend -t $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend frontend
 	docker push $(LOCAL_REGISTRY_NAME):$(REGISTRY_PORT)/capella/collab/frontend
 
+capella:
+	for version in $(CAPELLA_VERSIONS)
+	do $(MAKE) capella-$$version
+	done
+
+capella-%: registry
+	$(CAPELLA_DOCKERIMAGES) CAPELLA_VERSION=$* capella/remote capella/readonly
```
MoritzWeber0, line 39:
```suggestion
capella:
	for version in $(CAPELLA_VERSIONS)
	do $(CAPELLA_DOCKERIMAGES) CAPELLA_VERSION=$$version capella/remote capella/readonly
	done
```

Makes it clearer and keeps it simpler. We shouldn't start building commands around the Capella Docker images targets, they belong to the Capella Docker images Makefile.

MoritzWeber0 (changes requested):

backend/capellacollab/core/database/migration.py
amolenaar: I misread that as the readonly image.

The capella images are built by default. For the T4C images the (new) user has to do more work. I think it makes sense to do the default setup as new-user friendly as possible.

Makefile
amolenaar: > The default behavior should not change. The whole building process takes a long time and it's high hurdle for new users.

It's either that or they get only one working Capella version and the other versions are just dummies, which also does not give a good impression.

backend/capellacollab/core/database/migration.py
MoritzWeber0: On the other hand, it has to changed manually for each production version. The best case would be to detect if it was a Capella deployment with or without T4C. 

I'd rather argue that 90% of the users want to use the Collab Manager with TeamForCapella, so 90% would have to change it. For the other 10% of users, we can add a sentence to the README that it has to be changed.

Makefile
MoritzWeber0: But waiting more than one hour for the first deployment is also not really great. Also something for the README to describe that we build only one version, but it can be changed with the environment variable.

MoritzWeber0 (changes requested):
Everything looks good, only one small change that let's the pipeline fail.

backend/tests/test_sessions_routes.py
```diff
@@ -307,5 +307,5 @@ def test_create_persistent_session_as_user(client, db, user, kubernetes):
     assert kubernetes.sessions
     assert (
         kubernetes.sessions[0]["docker_image"]
-        == "k3d-myregistry.localhost:12345/t4c/client/remote:5.0.0-latest"
+        == "k3d-myregistry.localhost:12345/capella/remote:5.0.0-latest"
```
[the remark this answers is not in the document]
amolenaar: fixed

Makefile
```diff
@@ -9,19 +9,18 @@ RELEASE = dev-t4c-manager
...
 export MSYS_NO_PATHCONV := 1
 
 build: backend frontend docs
 
-build-capella:
-	$(CAPELLA_DOCKERIMAGES) capella/remote capella/readonly
-
-build-t4c: build
-	$(CAPELLA_DOCKERIMAGES) t4c/client/remote t4c/client/backup
+build-t4c: build t4c-client
```
MoritzWeber0, line 23: We can also remove this target, I assume.

amolenaar: I suppose...

MoritzWeber0 (approved):

PR #378, opened 2022-11-14
Persistent session per tool/version

### LINKED ISSUE #298: Create persistent volume claims per tool/version
There has to be one persistent volume per (username, tool, version). Migration has to be possible (can be done manually). Capella workspaces depend on the version.
E.g. on instance xy, we'd like to keep the existing workspaces, but can manually rename the persistent volume claims in the cluster (to map them on the used Capella version).

Right now, the name is determined in `_get_claim_name` and is the username with some removed characters. If possible, we could add labels with tool and version and select the proper persistent volume by the labels. If not, we have to add the tool and version to the volume name (should be still human readable to allow the renaming).

amolenaar:
# Description

Create a persistent volume per tool/version. 

In this installment, only one persistent session can be active at a time. I think that's safe as a starting point.

Resolves #298

Needs #376 

# Testing

Manual

# Checklist

### DIFF
--- a/backend/capellacollab/sessions/operators/k8s.py
+++ b/backend/capellacollab/sessions/operators/k8s.py
@@ -19,6 +19,8 @@ import kubernetes.config
 import kubernetes.stream.stream
 
 from capellacollab.config import config
+from capellacollab.projects.capellamodels.models import Version
+from capellacollab.tools.crud import get_image_for_tool_version
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


MoritzWeber0 (approved):
LGTM, but we need to migrate existing persistent volumes. Can you share a `kubectl` command that we can run to migrate the persistent volumes?

amolenaar:
Let's close this for now. It look like we can get away with one shared persistent volume.

PR #406, opened 2022-11-17
Fix login

### LINKED ISSUE #401: Redirect back to origin route after required authentication
When authenticating, the user should be redirected to the route the user is coming from. Right now, we already redirect to the index page. In addition, the redirect to the SSO provider should happen automatically without manual action.

amolenaar:
# Description

## Current behavior

* When refreshing a page you'll end up on the login page (`/auth`).
* After login you'll end up on the default page ('/').

## New behavior

* When refreshing, you're automatically forwarded to the SSO provider page. This normally results in instant authentication.
* When redirected back, you'll be forwarded to the page you were on before reloading.

Resolves #401

# Testing

Locally, with oauth mock (the mock requires you to provide a new name).

# Checklist

### DIFF
--- a/frontend/src/app/general/auth/auth/auth.component.ts
+++ b/frontend/src/app/general/auth/auth/auth.component.ts
@@ -18,7 +18,7 @@ export class AuthComponent implements OnInit {
   @Input()
   set autoLogin(value: boolean) {
     if (value) {
-      this.webSSO();
+      this.authService.webSSO();
     }
   }
 
@@ -41,8 +41,6 @@ export class AuthComponent implements OnInit {
   }
 
   webSSO() {
-    this.authService.getRedirectURL().subscribe((res) => {
-      window.location.href = res.auth_url;
-    });
+    this.authService.webSSO();
   }
 }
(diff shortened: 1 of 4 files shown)


frontend/src/app/general/auth/auth/auth.component.ts
```diff
@@ -41,8 +41,6 @@ export class AuthComponent implements OnInit {
   }
 
   webSSO() {
-    this.authService.getRedirectURL().subscribe((res) => {
-      window.location.href = res.auth_url;
-    });
+    this.authService.webSSO();
   }
```
MoritzWeber0, line 45: I assume this can be replaces by calling `this.authService.webSSO()` directly.

amolenaar: I wanted to leave it in, since `webSSO()` is invoked from the html template as well.

MoritzWeber0: I'd make it public then and replace it in the HTML template as well.

MoritzWeber0 (approved):

frontend/src/app/general/auth/auth/auth.component.ts
amolenaar: I feel it would be better to use the component as the abstraction level, though. Referencing services directly is a first  step towards the creation of [train wrecks](https://wiki.c2.com/?TrainWreck). It always starts with one...

PR #423, opened 2022-11-25
Replace projects_name with project_id for ProjectUserAssociation

### LINKED ISSUE #319: Change `projects_name` to `project_slug`
We should consistently change `projects_name` to `project_name` as you did for `get_user_of_project`. But this can also be done from my side when I revise the uses of the project name.

_Originally posted by @dominik003 in https://github.com/DSD-DBS/capella-collab-manager/pull/312#discussion_r997191158_
      

Additionally, we should change the `project_name` to `project_slug` where applicable.

TimoBbz:
# Description

This PR removes the last uses of projects_name as a foreign key, to replace with project_id. It also refactors the ProjectUserAssociation crud to use SQLAlchemy 2.0 syntax and object-level comparisons.

Resolves #319


backend/capellacollab/projects/routes.py
```diff
@@ -57,7 +55,7 @@ def get_projects(
     if RoleVerification(required_role=Role.ADMIN, verify=False)(token, db):
         return crud.get_all_projects(db)
 
-    return [project.projects for project in user.projects]
+    return [project.project for project in user.projects]
```
amolenaar, line 58: Why `project.project`? Shouldn't `user.project` contain a list of projects?

backend/capellacollab/users/models.py
```diff
@@ -34,14 +41,14 @@ class PatchUserRoleRequest(BaseModel):
...
     __tablename__ = "users"
 
-    id = Column(Integer, primary_key=True, index=True)
-    name = Column(String, unique=True, index=True)
-    role = Column(Enum(Role))
-    projects = relationship(
+    id: int = Column(Integer, primary_key=True, index=True)
+    name: str = Column(String, unique=True, index=True)
+    role: Role = Column(Enum(Role))
+    projects: list["ProjectUserAssociation"] = relationship(
```
amolenaar, line 47: Is SQLAlchemy not able to abstract over ProjectUserAssociation? It's only an a link for the n:m relationship between projects and users.

amolenaar (approved):
:shipit:

backend/capellacollab/projects/routes.py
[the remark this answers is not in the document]
TimoBbz: You’re right, I could rather use `association`:
```suggestion
    return [association.project for association in user.projects]
```

backend/capellacollab/users/models.py
TimoBbz: As far as I know, it doesn’t, and since `ProjectUserAssociation` has properties we want to access, it’s better to keep it like this.

MoritzWeber0 (approved):

PR #424, opened 2022-11-25
Introduce project renaming

### LINKED ISSUE #396: provide means to rename a project
In the current UI we are missing a feature that would enable project name change. As projects were migrated from the earlier instance, they inherited names of the models which might be misleading in some cases

TimoBbz:
# Description

Projects can now be renamed. The slug stays unchanged.

- [x] Validation on the front-end
- [x] Validation on the back-end
- [x] Update toaster

Resolves #396

### DIFF
--- a/backend/capellacollab/projects/crud.py
+++ b/backend/capellacollab/projects/crud.py
@@ -33,6 +33,7 @@ def update_project(
 ) -> DatabaseProject:
     if patch_project.name:
         project.name = patch_project.name
+        project.slug = slugify(patch_project.name)
     if patch_project.description:
         project.description = patch_project.description
     db.commit()
--- a/backend/capellacollab/projects/crud.py
+++ b/backend/capellacollab/projects/crud.py
@@ -9,7 +9,7 @@ import typing as t
 from slugify import slugify
 from sqlalchemy.orm import Session
 
-from capellacollab.projects.models import DatabaseProject
+from capellacollab.projects.models import DatabaseProject, PatchProject
 
 
 def get_project_by_name(db: Session, name: str) -> DatabaseProject:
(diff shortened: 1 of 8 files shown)


backend/capellacollab/projects/crud.py
```diff
@@ -28,10 +28,14 @@ def get_all_projects(db: Session) -> t.List[DatabaseProject]:
...
 
-def update_description(
-    db: Session, project: DatabaseProject, description: str
+def update_project(
+    db: Session, project: DatabaseProject, patch_project: PatchProject
 ) -> DatabaseProject:
-    project.description = description
+    if patch_project.name:
+        project.name = patch_project.name
+        project.slug = slugify(patch_project.name)
```
amolenaar, line 36: Do we want to update the slug when the name changes? Shouldn't slugs be constant (system) identifiers?

frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.html
```diff
@@ -6,21 +6,45 @@
...
+        </mat-form-field>
+      </fieldset>
+      <fieldset>
+        <mat-form-field appearance="fill">
+          <mat-label>Description</mat-label>
+          <textarea
+            formControlName="description"
+            matInput
+            rows="3"
+            style="resize: none"
```
amolenaar, line 34: No harm done if we support resizing.

amolenaar (changes requested):
IIRC we decided to not change the slug when a project is renamed. Renaming the slug will automatically break the UI for all users on that slug/path.

backend/capellacollab/projects/crud.py
MoritzWeber0: It's like renaming a project on Gitlab - they have two options: Updating the name and the slug (path). For now, let's just leave it as it is. In the future, we can consider adding a separate route. 
Project renaming is rare anyway.

frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.html
MoritzWeber0: Kind of, the tiles in the overview are not expandable, so too long descriptions will break them - but this is something we should fix.

MoritzWeber0 (approved):

MoritzWeber0 (approved):

PR #412, opened 2022-11-18
Manage pure::variants license.

TimoBbz:
# Description

Included the pure::variant license management in the database, backend and frontend.
TODO:
- [x] Frontend validation
- [x] Backend validation
- [x] Database validation
- [x] Frontend style
- [x] Frontend permissions
- [x] Bridge with the Docker container
- [x] Toaster for update

Resolves #258

# Testing

How were the changes tested?

# Checklist

- [ ] My code follows the guidelines of this project: [CONTRIBUTING.md](/CONTRIBUTING.md)
- [ ] I updated the documentation with the new functionality
- [ ] I went through the code and removed all print statements, breakpoints and unused code
- [ ] Error handling for all possible scenarios has been implemented
- [ ] I've add logging for all relevant operations
- [ ] I added the appropriate labels to this PR (enhancement/bug/documentation, effort, priority)

PR #425, opened 2022-11-28
Pure::variants support

MoritzWeber0:
This PR contains the first part of the `pure::variants` implementation. It was created as separate PR to improve efficiency. 

It includes the following features: 
- Set a `pure::variants` license server URL in the frontend. It is stored in the backend and mounted as environment variable into containers
- Preparation work for license server key mounting. The call is not implemented yet, so there is no way to use the license key mounting yet. 

In addition, it is missing usage restrictions. With the implementation of this PR, the license server is added to each persistent session (but not the license key, so there is no security issue).


amolenaar:
@MoritzWeber0 Don't you want to merge https://github.com/DSD-DBS/capella-collab-manager/pull/412 first?

MoritzWeber0:
> @MoritzWeber0 Don't you want to merge #412 first?

This PR has #412 as base. Mounting of license keys is quite important for #412 (otherwise users can not use the pure::variants sessions as there is no license).

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -109,8 +125,10 @@ def start_persistent_session(
                 "RMT_PASSWORD": password,
                 "FILESERVICE_PASSWORD": password,
                 "T4C_USERNAME": username,
+                "PURE_VARIANTS_LICENSE_SERVER": pure_variants_license_server,
```
amolenaar, line 128: Shouldn't this dict be replaced by environment?

backend/capellacollab/sessions/routes.py
```diff
@@ -281,12 +282,17 @@ def request_persistent_session(
...
-        password=rdp_password,
-        docker_image=docker_image,
-        t4c_license_secret=t4c_license_secret,
-        t4c_json=t4c_json,
+        get_username(token),
+        rdp_password,
+        docker_image,
+        t4c_license_secret,
+        t4c_json,
+        pv_license_env,
```
amolenaar, line 295: I prefer the named arguments: there are to many to remember in  order.

backend/capellacollab/settings/integrations/purevariants/models.py
```diff
@@ -0,0 +1,46 @@
...
+
+log = logging.getLogger(__name__)
+
+
+def validate_license_url(value: t.Optional[str]):
+    if value:
+        try:
+            requests.Request("GET", value).prepare()
+        except RequestException:
+            log.info("Floating license validation failed", exc_info=True)
```
amolenaar, line 25: We do not need a stack trace here.

frontend/src/app/helpers/skeleton-loaders/form-field-skeleton-loader/form-field-skeleton-loader.component.html
```diff
@@ -4,14 +4,15 @@
...
-    class="test"
-    appearance="circle"
-    [theme]="{
-      'border-radius': '5px',
-      height: '52px',
-      width: '448px',
-      border: '1px solid white'
-    }"
-  ></ngx-skeleton-loader>
+  <div class="margin">
```
amolenaar, line 7: Having this skeleton loading thingy is nice, I suppose, but isn't it a bit over the top? Does it really take that long to load the license server URL.

IMHO less is more, in this case.

frontend/src/app/settings/integrations/pure-variants/pure-variants.component.html
```diff
@@ -0,0 +1,28 @@
...
+ -->
+
+<div class="flex-center">
+  <mat-card class="section-card">
+    <h2>pure::variants configuration</h2>
+    <form (submit)="onSubmit()" [formGroup]="form">
+      <fieldset>
+        <app-form-field-skeleton-loader
+          *ngIf="loading"
+        ></app-form-field-skeleton-loader>
```
amolenaar, line 13: Doesn't seem necessary to me.

frontend/src/app/settings/integrations/pure-variants/pure-variants.component.ts
```diff
@@ -0,0 +1,57 @@
...
+        this.form.controls.licenseServerURL.patchValue(
+          res.license_server_url || ''
+        );
+      });
+  }
+
+  onSubmit(): void {
+    this.loading = true;
+    this.pureVariantsService
+      .setLicenseServerURL(this.form.value.licenseServerURL!)
```
amolenaar, line 47:
```suggestion
    if (this.form.value.licenseServerURL)
      this.pureVariantsService
        .setLicenseServerURL(this.form.value.licenseServerURL)
```

amolenaar (approved):
A few (small) changes to wrap it up.

backend/capellacollab/sessions/operators/k8s.py
MoritzWeber0: Yes, absolutely.

backend/capellacollab/settings/integrations/purevariants/models.py
MoritzWeber0: It's nice for finding the root cause of a failed validation, e.g. we want to add a license server URL, but the validation fails. With the stacktrace in the logs, we can identify the reason more easily.

frontend/src/app/helpers/skeleton-loaders/form-field-skeleton-loader/form-field-skeleton-loader.component.html
MoritzWeber0: Depends on the network connection and the routing in the cluster. Locally, it loads immediately. On the server side it can take some seconds. We have some components without loading animation and it feels really "laggy". I always think that there is no license linked and after some time, the license pops up.

frontend/src/app/settings/integrations/pure-variants/pure-variants.component.html
MoritzWeber0: User experience :)

frontend/src/app/settings/integrations/pure-variants/pure-variants.component.ts
MoritzWeber0: I've disabled the button when the form is invalid. The validator checks if the licenseServerURL is invalid, so we can assert that it is valid at this point.

frontend/src/app/helpers/skeleton-loaders/form-field-skeleton-loader/form-field-skeleton-loader.component.html
MoritzWeber0: And the important thing is: It keeps loading when the request fails. Otherwise, we would need to handle this case separately.

PR #433, opened 2022-11-28
Adjust git url validation

### LINKED ISSUE #419: Allow `%` characters in Git URLs
We should have to allow `%` characters in Git URLs (e.g. `%20` is valid), but still have a detection for path traversal attacks.

dominik003:
# Description

So far, the method to prevent the input of malicious URLs has been to use a blacklist of sequences that basically prohibit the use of `%` or `..`. However, this method is too restrictive since there are urls that use for example `%20`. This PR revises the url checking by removing the sequence blacklist and now always resolves the given url first (e.g. `https://test.com/test/%2e%2e%2f` -> `https://test.com/test/../` -> `https://test.com/`) before checking if there exists a git setting with a url that is a prefix of the given resolved url. This way we no longer disallow special characters in the URL, but still only allow URLs that are a prefix of a given git setting.

Resolves #419

# Testing

First, entering a URL with a special character (`%20`) works when no git settings are specified, but since I do not have access to a repository using special characters in the url, I cannot test the actual creation of the git source. Second, if a Git setting is specified (e.g. `https://github.com/DSD-DBS/capella-collab-manager%20`) and this is entered in the absolute URL field, the correct base integration URL is selected and one can also further extend the url with a relativ path, but again I cannot test whether the actual creation works.

# Checklist

- [x] @MoritzWeber0 Testing the creation of Git sources using a repository with special characters


frontend/src/app/projects/models/add-git-source/add-git-source.component.spec.ts
```diff
@@ -5,20 +5,20 @@
 
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 
-import { CreateCoworkingMethodComponent } from './create-coworking-method.component';
+import { AddGitSourceComponent } from './add-git-source';
 
-xdescribe('CreateCoworkingMethodComponent', () => {
-  let component: CreateCoworkingMethodComponent;
-  let fixture: ComponentFixture<CreateCoworkingMethodComponent>;
+xdescribe('AddGitSourceComponent', () => {
```
amolenaar, line 10: Is it possible to make this test executable?

frontend/src/app/projects/models/add-git-source/add-git-source.html
```diff
@@ -72,21 +70,12 @@ <h2>
...
-              class="error"
-              *ngIf="form.controls.urls.errors?.urlSafetyError && resultUrl"
-            >
-              {{ form.controls.urls.errors?.urlSafetyError }}
-            </mat-card-content>
-            <mat-card-content
+            <mat-error
               class="error"
-              *ngIf="form.controls.urls.errors?.urlPrefixError && resultUrl"
+              *ngIf="this.form.controls.urls.errors?.urlPrefixError"
```
amolenaar, line 75: `this` should not be needed here, right?

amolenaar (approved):
Looks good.

The naming of functions and variables improved.

frontend/src/app/projects/models/add-git-source/add-git-source.component.spec.ts
dominik003: I will leave it as is for now so we can merge this PR, but we should discuss this next week as it is a more general issue

MoritzWeber0 (changes requested):
- [ ] Not directly related to this PR, but `There exists no git setting being a prefix of the resolved url` should be renamed to `The resolved URL doesn't match with one of the allowed Git instances`.
- [ ] When the prefix doesn't match anymore, there is not error message in the frontend.
- [x] Deletion of Git instances doesn't work anymore

I can confirm that URLs with `%20` work again.

dominik003:
> 

Done

dominik003:
> * [x]  Not directly related to this PR, but `There exists no git setting being a prefix of the resolved url` should be renamed to `The resolved URL doesn't match with one of the allowed Git instances`.
> 
>     * [x]  When the prefix doesn't match anymore, there is not error message in the frontend.
> 
>     * [x]  Deletion of Git instances doesn't work anymore
> 
> 
> I can confirm that URLs with `%20` work again.

Done

Issue #24, opened 2022-03-01
Allow deletion of projects and models


MoritzWeber0:
Project managers should be able to delete models in a project.
The deletion of models should only work if: 
- There is not linked Git repository or T4C model (has to be unlinked first)

Project managers should be able to "unlink" T4C models and Git repositories from a generic model.
The unlinking of a T4C model or Git repository should only work if: 
- There is no backup defined

Administrators should be able to delete projects if: 
- There is no model defined

PR #444, opened 2022-12-02
Delete model

amolenaar:
# Description

Delete models when no git/t4c models are attached.

Resolves part of #24

One issue: it still shows up the in "project-wrapper".

# Testing

How were the changes tested?

# Checklist

### DIFF
--- a/backend/capellacollab/projects/toolmodels/routes.py
+++ b/backend/capellacollab/projects/toolmodels/routes.py
@@ -114,6 +114,20 @@ def patch_capella_model(
     return crud.update_model(db, model, body.description, version, nature)
 
 
+@router.delete(
+    "/{model_slug}",
+    dependencies=[
+        Depends(ProjectRoleVerification(required_role=ProjectUserRole.MANAGER))
+    ],
+    tags=["Projects - Models"],
+)
+def patch_capella_model(
+    model: DatabaseCapellaModel = Depends(get_existing_capella_model),
+    db: Session = Depends(get_db),
+) -> DatabaseCapellaModel:
+    crud.delete_model(db, model)
+
+
 def get_tool_by_id_or_raise(db: Session, tool_id: int) -> Tool:
     try:
         return tools_crud.get_tool_by_id(tool_id, db)
--- a/backend/capellacollab/projects/toolmodels/routes.py
+++ b/backend/capellacollab/projects/toolmodels/routes.py
@@ -116,16 +116,18 @@ def patch_capella_model(
 
 @router.delete(
     "/{model_slug}",
+    status_code=204,
     dependencies=[
         Depends(ProjectRoleVerification(required_role=ProjectUserRole.MANAGER))
     ],
     tags=["Projects - Models"],
 )
-def patch_capella_model(
+def delete_capella_model(
     model: DatabaseCapellaModel = Depends(get_existing_capella_model),
     db: Session = Depends(get_db),
 ) -> DatabaseCapellaModel:
-    crud.delete_model(db, model)
+    if not (model.git_models or model.t4c_models):
+        crud.delete_model(db, model)
 
 
 def get_tool_by_id_or_raise(db: Session, tool_id: int) -> Tool:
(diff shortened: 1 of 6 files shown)


backend/capellacollab/projects/capellamodels/routes.py
```diff
@@ -114,6 +114,20 @@ def patch_capella_model(
...
+    dependencies=[
+        Depends(ProjectRoleVerification(required_role=ProjectUserRole.MANAGER))
+    ],
+    tags=["Projects - Models"],
+)
+def patch_capella_model(
+    model: DatabaseCapellaModel = Depends(get_existing_capella_model),
+    db: Session = Depends(get_db),
+) -> DatabaseCapellaModel:
+    crud.delete_model(db, model)
```
dominik003, line 128:
```suggestion
):
    if not (len(model.git_models) or len(model.t4c_models)):
        crud.delete_model(db, model)
```
Here I would add a backend validation (in addition to the frontend check). In case there are still git/t4c models attached, we could also throw an http exception.

frontend/src/app/services/model/model.service.ts
```diff
@@ -76,6 +76,12 @@ export class ModelService {
     );
   }
 
+  deleteModel(project_slug: string, model: Model): Observable<any> {
+    return this.http.delete(
```
dominik003, line 80:
```suggestion
  deleteModel(project_slug: string, model: Model): Observable<void> {
    return this.http.delete<void>(
```
Since we do not expect a return value, I would use `void` here.

frontend/src/app/projects/models/model-description/model-description.component.html
```diff
@@ -51,9 +51,19 @@
...
-        <button mat-flat-button color="primary" type="submit">
-          Update model
-        </button>
+        <div>
+          <button
+            mat-flat-button
+            color="warn"
+            (click)="deleteModel()"
+            *ngIf="canDelete"
+          >
```
dominik003, line 60:
```suggestion
          <button
            mat-flat-button
            color="warn"
            (click)="deleteModel()"
            *ngIf="canDelete"
            type="button"
          >
```
The default type of the button here is `submit` and therefore the `onSubmit()` function is called when the cancel button is clicked.

In addition, the delete button is not visible if there are still git/t4c models. I think it would be better to use a solution here that shows that you are normally able to delete a model, but that this is currently not possible due to existing git/t4c models. Here are some solutions I could think of:
- disable the button and add a tooltip with something like *Please remove all Git and T4C models first before deleting this model*
- disable the button and add an information text above it that prompts the user to remove the models first.
- do the same thing Moritz did with deleting tool versions (and other things): Always leave the button enabled, but if you click on it and there are still dependencies, show them in the error message

dominik003 (changes requested):

dominik003:
> @dominik003 Something like this:
> 
> image

This looks great :thumbsup:

backend/capellacollab/projects/toolmodels/routes.py
```diff
@@ -114,6 +114,22 @@ def patch_capella_model(
...
+    status_code=204,
+    dependencies=[
+        Depends(ProjectRoleVerification(required_role=ProjectUserRole.MANAGER))
+    ],
+    tags=["Projects - Models"],
+)
+def delete_capella_model(
+    model: DatabaseCapellaModel = Depends(get_existing_capella_model),
+    db: Session = Depends(get_db),
+) -> DatabaseCapellaModel:
```
dominik003, line 128:
```suggestion
):
```

dominik003 (approved):
Looks go to me :+1: (There is only one type hint that can be removed for the deletion route)

MoritzWeber0 (approved):
Improvements for the future: 

- [ ] Add tests for tool model routes: https://github.com/DSD-DBS/capella-collab-manager/issues/458
- [ ] To display the reason for a disabled button, I like the tooltip approach: https://material.angular.io/components/tooltip/overview

frontend/src/app/projects/models/model-description/model-description.component.html
MoritzWeber0: > I do not like this, since it clutters the interface with things you cannot do anyway. 

Isn't this an argument for the tooltip? The tooltip only shows the information if you hover over the `Delete` button. Without the tooltip, I see the information about the deletion - even if I just want to update the model, which may be confusing. 
You want to update the model and it shows: `You can not delete the model because of linked Git / T4C models`. 

The tooltip message only appears when you intend to click the Delete button.

Issue #404, opened 2022-11-17
Updating T4C models doesn't work


MoritzWeber0:
When having an existing T4C model, one can not update it. It results in a ValidationError. In general, it makes sense to replace the used query parameters with path parameters and a body.

PR #449, opened 2022-12-12
TeamForCapella model update fix

MoritzWeber0:
This PR fixes the model update bug (#404). When looking into the code, it was hard to understand what is going on in the code. Thus, I also reduced some complexity and removed some code, which was not important.


frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
```diff
@@ -0,0 +1,200 @@
...
+})
+export class ManageT4CModelComponent implements OnInit, OnDestroy {
+  @Input() asStepper?: boolean;
+  @Output() create = new EventEmitter<boolean>();
+
+  editing = false;
+  loading = false;
+
+  t4cInstances: T4CInstance[] | undefined = undefined;
+  get selectedT4CInstance(): T4CInstance | null | undefined {
```
dominik003, line 48: Why do you need the undefined case here? When I needed to handle lists of resources in some components, I just assigned the empty list as the default value, since this doesn't really break anything and you have reduced complexity by not having to consider the extra case (e.g. you could remove the `null` return type for `get selectedT4CInstance`).
The same also applies for the `t4cRepositories`

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
```diff
@@ -0,0 +1,200 @@
...
+        repository.id == this.form.value.t4cRepositoryId
+    );
+  }
+
+  t4cModel?: T4CModel;
+
+  public form = new FormGroup({
+    t4cInstanceId: new FormControl<null | number>(null, [
+      Validators.required,
+      notNullValidator(),
```
dominik003, line 67: Why do we need these two validators and the `required` validator is not enough?

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
```diff
@@ -0,0 +1,200 @@
...
+      });
+  }
+
+  fetchT4CInstances(): Observable<T4CInstance[]> {
+    return this.t4cInstanceService.listInstances().pipe(
+      tap((instances) => {
+        this.t4cInstances = instances;
+        this.form.controls.t4cInstanceId.enable({ emitEvent: false });
+        this.form.controls.t4cInstanceId.patchValue(
+          this.form.controls.t4cInstanceId.value
```
dominik003, line 106: This part is not really intuitive, couldn't we use `this.form.controls.t4cInstanceId.updateValueAndValidity()` for the same behavior?

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
```diff
@@ -0,0 +1,200 @@
...
+    return this.t4cModelService._t4cModel.pipe(
+      tap((model: T4CModel | undefined) => {
+        if (model) {
+          this.editing = true;
+          this.t4cModel = model;
+
+          this.form.controls.t4cInstanceId.setValue(
+            model.repository.instance.id,
+            { emitEvent: true }
+          );
```
dominik003, line 145:
```suggestion
          this.form.controls.t4cInstanceId.setValue(
            model.repository.instance.id
          );
```
True is the default case here.

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
```diff
@@ -0,0 +1,200 @@
...
+            }
+
+            this.toastService.showSuccess(
+              'TeamForCapella successfully created',
+              ''
+            );
+          });
+      }
+    }
+  }
```
dominik003, line 195:
```suggestion
  onSubmit() {
    if (this.form.invalid) {
      return;
    }
    if (this.t4cModelService.t4cModel) {
      this.t4cModelService
        .patchT4CModel(
          this.projectService.project!.slug,
          this.modelService.model!.slug,
          this.t4cModelService.t4cModel.id,
          this.form.value as SubmitT4CModel
        )
        .subscribe((model) => {
          this.t4cModelService._t4cModel.next(model);
          this.patchFormWithExistingT4CModel();

          this.toastService.showSuccess(
            'TeamForCapella successfully updated',
            ''
          );
        });
    } else {
      this.t4cModelService
        .createT4CModel(
          this.projectService.project!.slug,
          this.modelService.model!.slug,
          this.form.value as SubmitT4CModel
        )
        .subscribe((_) => {
          if (this.asStepper) {
            this.create.emit(true);
          } else {
            this.router.navigate(['../..'], { relativeTo: this.route });
          }

          this.toastService.showSuccess(
            'TeamForCapella successfully created',
            ''
          );
        });
    }
  }
```
This reduces the nesting level here, which makes it a little easier to read.

frontend/src/app/projects/models/model-source/t4c/service/t4c-model.service.ts
```diff
@@ -79,14 +68,18 @@ export class T4CModelService {
   ): Observable<T4CModel> {
     return this.http.patch<T4CModel>(
       `${this.urlFactory(project_slug, model_slug)}/${t4c_model_id}`,
-      body
+      {
+        t4c_instance_id: body.t4cInstanceId,
+        t4c_repository_id: body.t4cRepositoryId,
+        name: body.name,
+      }
```
dominik003, line 75: This could possibly be moved to a function, but since it's only used within this service right now and only twice, it's also fine to leave it that way.

dominik003 (commented):
I've added some comments that I don't think are directly related to or from this PR, but I always prefer to add suggestions/comments when I see something that can potentially be improved.

backend/capellacollab/projects/toolmodels/modelsources/t4c/routes.py
```diff
@@ -34,44 +31,30 @@
...
+@router.get("", response_model=list[ResponseT4CModel])
 def list_t4c_models(
-    project: DatabaseProject = Depends(get_existing_project),
     model: DatabaseCapellaModel = Depends(get_existing_capella_model),
-    repository: t.Optional[DatabaseT4CRepository] = Depends(
-        get_optional_existing_instance_repository
-    ),
-    db: Session = Depends(database.get_db),
-    token=Depends(JWTBearer()),
+    db_session: Session = Depends(database.get_db),
```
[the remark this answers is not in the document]
MoritzWeber0: Switched back for `db_session`. 

> And I would change `database.get_db` to `get_db` to match most other places.

The Google style guide says something different. We should rather rename it in the other places.
"Use import statements for packages and modules only, not for individual classes or functions."

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
```diff
@@ -180,9 +178,7 @@ def delete_t4c_repository(
 )
 def start_t4c_repository(
     instance: DatabaseT4CInstance = Depends(get_existing_instance),
-    repository: DatabaseT4CRepository = Depends(
-        get_existing_instance_repository
-    ),
+    repository: DatabaseT4CRepository = Depends(get_existing_t4c_repository),
 ) -> None:
     interface.start_repository(instance, repository.name)
     return None
```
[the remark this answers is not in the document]
MoritzWeber0: This was necessary in an older `fastAPI` version. I've removed it.

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
```diff
@@ -194,9 +190,7 @@ def start_t4c_repository(
 )
 def stop_t4c_repository(
     instance: DatabaseT4CInstance = Depends(get_existing_instance),
-    repository: DatabaseT4CRepository = Depends(
-        get_existing_instance_repository
-    ),
+    repository: DatabaseT4CRepository = Depends(get_existing_t4c_repository),
 ) -> None:
     interface.stop_repository(instance, repository.name)
     return None
```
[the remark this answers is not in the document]
MoritzWeber0: Removed.

frontend/src/app/helpers/validators/validators.ts
```diff
@@ -24,3 +24,12 @@ export function regExpValidator(
...
 }
+
+export function notNullValidator(): ValidatorFn {
+  return (control: AbstractControl): ValidationErrors | null => {
+    if (control.value === null) {
+      return { notNullValidator: true };
+    }
+    return null;
+  };
+}
```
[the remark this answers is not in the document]
MoritzWeber0: Good question, I have no idea anymore. Removed it.

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
MoritzWeber0: Having an empty list as default value is bad style imho. An empty list means there is no T4C instance - which is different to the undefined case, which means "not loaded yet" / "not defined yet". In some parts I use this to indicate the loading animation -> If it is undefined, it shows the loading animation, otherwise not.

But I just noticed that `selectedT4CInstance` is not used. Removed it.

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
MoritzWeber0: Removed it.

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
MoritzWeber0: I think it's not needed at all. Removed it.

frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.ts
MoritzWeber0: I like. Applied.

frontend/src/app/projects/models/model-source/t4c/service/t4c-model.service.ts
MoritzWeber0: Let's keep it for now.

dominik003 (approved):

PR #455, opened 2022-12-13
Access control and auditing

### LINKED ISSUE #398: Access control and auditing
We need means to audit user access history. "manage users" perspective seems to be the right place for it.

What we should see there for an individual user is:

* when the user got access to the system
* last login date and time
* what projects and models the user has access to at the moment
* complete access rights history that shall include entries like:
   * <date time>: User <name> was granted <access level> to the <project / model> by <access sponsor>; reason for authorization: <free text that explains why the user shall have access to the project and that specific level of access>
   * <data time>: User <name> access authorization was revoked by <user who revoked authorization> due to <reason>

dominik003:
# Description

This PR adds access control and auditing so that any action related to user permissions/roles is recorded and can be retrieved via the user settings (for a specific user) or via the *Events* button in the *Profile* selection (showing all events). In addition, the user's creation date and last login time are saved and can be viewed in the user settings by selecting a specific user.
In the following I will explain which events are created and when they are created.

## User Events
- when a user is created by a simple login (without involvement of another user), an event *User created* is created and the executor is displayed as *System* (*System* is always displayed in the frontend if the *executor* is not defined)
- when an administrator creates a user, an event *User created* is generated with the administrator as the executor and the specified reason
- when an administrator upgrades the role of a user to administrator or vice versa a *Assigned Role Admin/User* event is created
- when a user is deleted, all events where he is either the affected user or the executor are deleted, and therefore no event is created

## Project User Events
- when a user is added to a project there are two different cases
	1. when a user is added as *Manager* only a *Added to Project* and *Assigned Project Role Manager* events are created (since managers by default have read/write access)
	1. when a user is added as a *User* in addition to the *Added to Project* and *Assigned Project Role User*, also a *Assigned Permission ReadOnly/ReadWrite* is created
- when a user is removed from a project a *Removed from Project* event is created
- any other role or permission change is captured by the corresponding event (i.e., *Assigned Role/Permission XY*)

Resolves #398 

# Testing

All of the above scenarios can be run manually, and one can see the generated events either in the user settings or via *Events* (found by clicking *Profile*). In addition, this PR adds several tests that check if the expected events are generated when using the corresponding routes.

# Checklist


backend/tests/test_event_creation.py
```diff
@@ -0,0 +1,372 @@
...
+            "reason": reason,
+        },
+    )
+
+    assert response.status_code == 200
+
+    events = get_events_by_user_id(db, user.id)
+    assert len(events) == 3
+
+    for event, expected_event_type in zip(
```
amolenaar, line 162: Try to avoid `for` loops and `if` statements in tests. You want tests to be fully deterministic and easy to read.

You can only test the most interesting fields, since the other fields have been covered in other tests as well.

backend/tests/test_event_creation.py
```diff
@@ -0,0 +1,372 @@
...
+    event = events[0]
+
+    assert event.event_type == expected_role_event_type
+    assert event.executor_id == executor.id
+    assert event.reason == reason
+    assert event.project_id == project.id
+    assert event.user_id == user.id
+
+
+def get_events_by_username(
```
amolenaar, line 351: Nice. "Helper" functions should belong at the bottom of the file :+1: .

amolenaar (approved):
I have some doubts about introducing `projects/users/core.py`. I think the naming is to generic and maybe it just belongs to the crud module.

Apart from that, a few suggestions to improve the tests.

It's good to merge, though.

backend/tests/test_event_creation.py
```diff
@@ -0,0 +1,372 @@
...
+    DatabaseUserHistoryEvent,
+    EventType,
+)
+from capellacollab.users.models import DatabaseUser, Role
+
+reason: str = "TestReason"
+
+
+def test_create_admin_user_by_system(db):
+    user: DatabaseUser = get_user_by_name(db, "your-username")
```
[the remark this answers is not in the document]
dominik003: Yes, I normally use uuids for names, but this is the name of the default admin that is created when the backend starts. (i.e., I don't create it here, I just get the default admin). But I agree with you that using static names like this is not so good, and this test could be replaced with one where we mirror the user creation via login (since this is much more work, I will keep the current test for now).

backend/tests/test_event_creation.py
```diff
@@ -0,0 +1,372 @@
...
+    assert response.status_code == 200
+    assert len(get_events_by_username(db, unique_username)) == 1
+    assert len(get_executed_events_by_user_id(db, executor.id)) == 1
+
+    user_id = int(response.json()["id"])
+    response = client.delete(f"/api/v1/users/{user_id}")
+
+    assert response.status_code == 204
+    assert len(get_events_by_username(db, unique_username)) == 0
+    assert len(get_executed_events_by_user_id(db, executor.id)) == 0
```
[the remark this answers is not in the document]
dominik003: Yes, agreed! I found another use and changed it there too

backend/tests/test_event_creation.py
dominik003: I also didn't like to use a loop here but I didn't find a cleaner solution to check all fields. But good point that I don't really have to check all the fields, so now I just check all fields only for the first event and for the other two I just check their event type.

backend/tests/test_event_creation.py
dominik003: It's actually only followed by two other helper crud functions, so its already kinda at the buttom.

dominik003:
> I have some doubts about introducing `projects/users/core.py`. I think the naming is to generic and maybe it just belongs to the crud module.
> 
> Apart from that, a few suggestions to improve the tests.
> 
> It's good to merge, though.

I didn't want to put it in `curd.py`, because 2/3 functions are simple helper functions that outsource some logic (and not containing any crud operations) and the other one contains more logic than we want in `crud.py` (from my point of view). But I renamed it now to `util.py`, which is still kind of generic, but maybe a bit more descriptive.

Applied your other comments to improve the tests :+1:

MoritzWeber0 (changes requested):
I wanted to try it out, but it seems like a database migration is missing completely. 

At least two columns have been added. Please fix this, otherwise login is not possible anymore.

backend/capellacollab/users/crud.py
```diff
@@ -48,6 +48,16 @@ def update_role_of_user(
     return user
 
 
+def update_last_login(
+    db: Session, user: DatabaseUser, last_login: datetime | None = None
+) -> DatabaseUser:
+    if not last_login:
+        last_login = datetime.now()
+    user.last_login = last_login
```
MoritzWeber0, line 56: This doesn't work when a new user signs in (the user is not created and therefore None).

```
AttributeError: 'NoneType' object has no attribute 'last_login'
```

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.html
```diff
@@ -134,22 +134,34 @@ <h2>Add user</h2>
...
+            >{{ permission.value }}</mat-option
+          >
+        </mat-select>
+        <mat-error>You have to select a permission!</mat-error>
+      </mat-form-field>
+    </div>
+    <br />
+    <mat-form-field>
+      <mat-label>Reason</mat-label>
+      <input matInput formControlName="reason" />
```
MoritzWeber0, line 156:
```suggestion
      <textarea matInput formControlName="reason"></textarea>
```

Reasons are usually a bit longer, it's better to be able to expand the field.

MoritzWeber0 (approved):
When the two improvements are implemented, it can be merged. Nice implementation, thanks @dominik003!

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.html
dominik003: Done

PR #457, opened 2022-12-13
Stopping the session loop when exited

dominik003:
# Description

This PR fixes the problem that you can't exit the Backed with `CTRL + C`.

### DIFF
--- a/backend/capellacollab/sessions/idletimeout.py
+++ b/backend/capellacollab/sessions/idletimeout.py
@@ -45,6 +45,7 @@ async def terminate_idle_sessions_in_background(interval=60):
                 await run_in_threadpool(terminate_idle_session)
             except BaseException:
                 log.exception("Could not handle idle sessions")
+                return
 
     asyncio.ensure_future(loop())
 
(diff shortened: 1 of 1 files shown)


dominik003:
> It still doesn't terminate properly - the background process doesn't stop. I get these messages in the terminal after stopping
> 
> ```
> ERROR:capellacollab.sessions.idletimeout:Could not collect idle sessions from Prometheus
> ```

When exactly is this happening and does it only occur once? After stopping, I get these messages in the terminal
```
INFO:uvicorn.error:Finished server process [84334]
ERROR:capellacollab.sessions.idletimeout:Could not handle idle sessions
...
asyncio.exceptions.CancelledError
INFO:     Stopping reloader process [84332]
```

and then nothing happens (so no more messages). My understanding is that the loop simply adds a single task to the thread pool and then waits the specified interval before adding the next task. So theoretically there could be another error message `Could not collect idle sessions from Prometheus`, but not more than one.

backend/capellacollab/sessions/idletimeout.py
```diff
@@ -45,6 +45,7 @@ async def loop():
                 await run_in_threadpool(terminate_idle_session)
             except BaseException:
                 log.exception("Could not handle idle sessions")
+                return
```
MoritzWeber0, line 48:
```suggestion
            except asyncio.exceptions.CancelledError:
                return
            except BaseException:
                log.exception("Could not handle idle sessions")
```

I'd rather handle it separately - then we also don't get the error message printed during exit. And in the case of an unknown error - the session termination still continues - otherwise it would stop completely.

MoritzWeber0 (changes requested):

PR #459, opened 2022-12-14
Use routes for session termination

MoritzWeber0:
TeamForCapella users don't get deleted otherwise.


dominik003 (commented):
Looks good as a fix, but I don`t like calling a route function directly. So I would probably move the logic that is in the route to a separate function (probably in `sessions.py`) and then call that function instead. Also, is there a reason why the `crud` functions are in `database.py` and we don`t have `crud.py`, and why do we have a separate `model.py` and `schema.py` as well?

MoritzWeber0:
> Looks good as a fix, but I don't like calling a route function directly. So I would probably move the logic that is in the route to a separate function (probably in `sessions.py`) and then call that function instead. Also, is there a reason why the `crud` functions are in `database.py` and we don`t have `crud.py`, and why do we have a separate `model.py`and`schema.py` as well?

`database` is renamed to `crud` and the terminate session code is moved to `core.py`. The `model.py` and `schema.py` are used in many places, so I'd handle it in another PR.

dominik003 (approved):

amolenaar (approved):
I have some doubts on the module name `core`. It's very generic and does not conceal it's intent (like `util`, `helper`, `misc`).

MoritzWeber0:
> I have some doubts on the module name `core`. It's very generic and does conceal it's intent (like `util`, `helper`, `misc`).

I've renamed `core` to `util`.

PR #461, opened 2022-12-14
Fix loading of git and t4c models

MoritzWeber0:
In the Git and T4C model overview, the loading of models is not done properly. The Git models and T4C models are only reset when there is a new update from the backend. When switching between the view fast enough, you can see Git and T4C models from other tool models. This PR fixes this behaviour and adds the skeleton loader.

### DIFF
--- a/frontend/src/app/helpers/skeleton-loaders/mat-card-overview-loader/mat-card-overview-loader.component.ts
+++ b/frontend/src/app/helpers/skeleton-loaders/mat-card-overview-loader/mat-card-overview-loader.component.ts
@@ -14,6 +14,8 @@ export class MatCardOverviewLoaderComponent implements OnInit {
   @Input() loading = true;
   @Input() reservedCards = 1;
 
+  @Input() rows: number | undefined = undefined;
+
   _cardNumbersArray: number[] = [];
   set cardsNumber(value: number) {
     this._cardNumbersArray = [...Array(value).keys()];
@@ -33,8 +35,14 @@ export class MatCardOverviewLoaderComponent implements OnInit {
   resize(width: number, height: number) {
     // Margin is 1vw (left + right) and 1vh (top + bottom)
 
-    const cardsPerColumn = (0.98 * width) / 425;
-    const cardsPerRow = (0.98 * height - 120) / 275;
+    let cardsPerColumn = 0;
+    if (this.rows) {
+      cardsPerColumn = this.rows;
+    } else {
+      cardsPerColumn = (0.98 * height - 120) / 275;
+    }
+
+    const cardsPerRow = (0.98 * width) / 425;
 
     this.cardsNumber =
       Math.trunc(cardsPerColumn) * Math.trunc(cardsPerRow) - this.reservedCards;
--- a/frontend/src/app/helpers/skeleton-loaders/mat-card-overview-loader/mat-card-overview-loader.component.ts
+++ b/frontend/src/app/helpers/skeleton-loaders/mat-card-overview-loader/mat-card-overview-loader.component.ts
@@ -35,13 +35,7 @@ export class MatCardOverviewLoaderComponent implements OnInit {
   resize(width: number, height: number) {
     // Margin is 1vw (left + right) and 1vh (top + bottom)
 
-    let cardsPerColumn = 0;
-    if (this.rows) {
-      cardsPerColumn = this.rows;
-    } else {
-      cardsPerColumn = (0.98 * height - 120) / 275;
-    }
-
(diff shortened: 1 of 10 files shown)


frontend/src/app/helpers/skeleton-loaders/mat-card-overview-loader/mat-card-overview-loader.component.ts
```diff
@@ -33,8 +35,14 @@ export class MatCardOverviewLoaderComponent implements OnInit {
...
     // Margin is 1vw (left + right) and 1vh (top + bottom)
 
-    const cardsPerColumn = (0.98 * width) / 425;
-    const cardsPerRow = (0.98 * height - 120) / 275;
+    let cardsPerColumn = 0;
+    if (this.rows) {
+      cardsPerColumn = this.rows;
+    } else {
+      cardsPerColumn = (0.98 * height - 120) / 275;
+    }
```
dominik003, line 43:
```suggestion
    let cardsPerColumn = this.rows ? this.rows : (0.98 * height - 120) / 275;
```
That more or less a personal preference, that I really like conditional assignments in these cases

frontend/src/app/projects/models/model-detail/model-detail.component.ts
```diff
@@ -22,9 +22,8 @@ import { ProjectService } from 'src/app/services/project/project.service';
   styleUrls: ['./model-detail.component.css'],
 })
 export class ModelDetailComponent implements OnInit, OnDestroy {
-  public gitModels: Array<GetGitModel> = [];
-  private _t4cModels = new BehaviorSubject<T4CModel[] | undefined>(undefined);
-  public t4cModels: T4CModel[] = [];
+  public gitModels: Array<GetGitModel> | undefined = undefined;
+  public t4cModels: T4CModel[] | undefined = undefined;
```
dominik003, line 26:
```suggestion
  public gitModels?: GetGitModel[] = undefined;
  public t4cModels?: T4CModel[] = undefined;
```
I would definitely change the `array<...>` to `...[]`, since we typically use the `[]` notation (even directly below it). 
The `xy | undefined = ...` to `?: xy = ...` is more or less a preference thing, so both variants are valid imo.

frontend/src/app/projects/models/model-detail/model-detail.component.ts
```diff
@@ -41,17 +40,13 @@ export class ModelDetailComponent implements OnInit, OnDestroy {
...
-
     this.t4cModelsSubscription = this.t4cModelService
       .listT4CModels(
         this.projectService.project!.slug,
         this.modelService.model!.slug
       )
       .subscribe((models) => {
-        this._t4cModels.next(models);
+        this.t4cModels = models;
       });
```
dominik003, line 50: I can't make a suggestion here (because of the deleted line), but I would change this to a simple one-liner `.subscribe((models) => (this.t4cModels = models));`

frontend/src/app/projects/project-detail/model-overview/model-detail/git-model.service.ts
```diff
@@ -16,8 +16,10 @@ export class GitModelService {
 
   constructor(private http: HttpClient) {}
 
-  private _gitModel = new Subject<GetGitModel>();
-  public _gitModels = new BehaviorSubject<Array<GetGitModel>>([]);
+  private _gitModel = new Subject<GetGitModel | undefined>();
+  public _gitModels = new BehaviorSubject<Array<GetGitModel> | undefined>(
+    undefined
+  );
```
dominik003, line 22:
```suggestion
  public _gitModels = new BehaviorSubject<GetGitModel[] | undefined>(undefined);
```
First I would remove the `Array<...>` since  we usually use the `[]` notation.

Also, I'm a little confused that the `_gitModels` behavior subject is public, since the main idea here is to use the `gitModels` observable outside of the service. There is only one place where we use `_gitModels` directly, but the initial idea was to use a local `gitModels`: GetGitModel[] = undefined` in components and set it by subscribing to the public `gitModels` observable and using `loadGitModels()`.

MoritzWeber0: > but the initial idea was to use a local `gitModels: GetGitModel[] = undefined` in components and set it by subscribing to the public `gitModels` observable and using `loadGitModels()`.

When we just want to get the value of `_gitModels`, I dislike the idea of replicating the whole `subscribe`, `unsubscribe` and a separate variable (which has the same value of the `gitModel` anyway then. But as we have no consistent decision on that, I'd skip it for now.

dominik003 (approved):

PR #471, opened 2022-12-16
Fix Github rate limiting during fetching of release notes in the pipeline

MoritzWeber0:


.github/workflows/push.yml
```diff
@@ -45,6 +45,8 @@ jobs:
       - name: Fetch version
         if: ${{ matrix.name == 'frontend' }}
         run: |
+          export GITHUB_USERNAME=GITHUB ACTIONS
```
amolenaar, line 48:
```suggestion
          export GITHUB_USERNAME="GITHUB ACTIONS"
```

amolenaar (changes requested):
Are those env vars really used?

MoritzWeber0:
> Are those env vars really used?

They are used in `frontend/fetch-version.ts`.

amolenaar (approved):

Issue #407, opened 2022-11-17
Change wording of "New version" message


MoritzWeber0:
In the bottom left corner of the application, one get's notified of new versions with the message "New version available". This wording is not optimal, because one expects a not deployed version. It should be changed to "New version. What's new?".

PR #472, opened 2022-12-19
Simplify release notes

amolenaar:
Simplify the release notes. Just link to a page in the docs (`/docs/release-notes`).

Removed Markdown component and configuration, since that was only used for rendering the release notes fetched from GH releases.


frontend/src/app/general/metadata/version/version.component.html
```diff
@@ -3,22 +3,21 @@
...
+    *ngIf="versionService.changedVersion && versionService.version"
+  >
+    New version available (<span *ngIf="versionService.oldVersion"
+      >{{ versionService.oldVersion }}
+      <mat-icon class="mat-icon-position top left"
+        >arrow_right_alt</mat-icon
+      ></span
+    >{{ versionService.version.git.tag }})!
+  </div>
+  <a href="/docs/release-notes/" target="_blank">
```
dominik003, line 18: I would probably also place the `<a>' around the *New version available ...* `<div>` because otherwise you can't click on it.

dominik003 (changes requested):
Looks good except for two small things but how do we deal with the current versions that are not yet included in the release notes? Maybe we can put the old style release notes for the recent versions in the `release-notes.md` first or what is your option?

docs/user/docs/release-notes.md
```diff
@@ -3,39 +3,41 @@
  ~ SPDX-License-Identifier: Apache-2.0
```
MoritzWeber0, line 3: I'd add a link to our Github releases page.

MoritzWeber0:
This should be included as well: https://github.com/DSD-DBS/capella-collab-manager/issues/407

PR #477, opened 2022-12-20
Upload `pure::variants` license in integration settings

MoritzWeber0:
Upload a `pure::variants` license file in the settings. The license is stored as `Secret` in the underlying kubernetes cluster.

### DIFF
--- a/backend/capellacollab/settings/integrations/purevariants/routes.py
+++ b/backend/capellacollab/settings/integrations/purevariants/routes.py
@@ -5,11 +5,13 @@ from __future__ import annotations
 
 import typing as t
 
-from fastapi import APIRouter, Depends
+import fastapi
 from sqlalchemy.orm import Session
 
 from capellacollab.core.authentication.database import RoleVerification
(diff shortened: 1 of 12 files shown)


backend/capellacollab/settings/integrations/purevariants/routes.py
```diff
@@ -5,20 +5,22 @@
 
 import typing as t
 
-from fastapi import APIRouter, Depends
+import fastapi
```
amolenaar, line 8: Why import fastapi, and not import from fastapi, as we do in all other modules?

MoritzWeber0:
```zsh
Use import statements for packages and modules only, not for individual classes or functions.
```

This is a rule in the Google styleguide and the team agreed to follow the Google styleguide. I personally also prefer the module imports, but if you don't agree, feel free to start a discussion in our cookiecutter repository: https://github.com/DSD-DBS/cookiecutter-dbs-python/discussions

PR #478, opened 2022-12-20
Delete project

amolenaar:
Delete projects once there are no models linked.

Also add a patch to remove a model from the frontend once deleted.

### DIFF
--- a/backend/capellacollab/projects/routes.py
+++ b/backend/capellacollab/projects/routes.py
@@ -140,9 +140,12 @@ def delete_project(
     project: DatabaseProject = Depends(get_existing_project),
     db: Session = Depends(get_db),
 ):
-    if not project.models:
-        users_crud.delete_users_from_project(db, project)
-        crud.delete_project(db, project)
+    if project.models:
+        raise HTTPException(
+            409, {"reason": "The project still has models assigned to it"}
+        )
+    users_crud.delete_users_from_project(db, project)
+    crud.delete_project(db, project)
 
 
 router.include_router(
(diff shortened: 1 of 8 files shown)


backend/capellacollab/projects/routes.py
```diff
@@ -140,8 +140,8 @@ def delete_project(
     project: DatabaseProject = Depends(get_existing_project),
     db: Session = Depends(get_db),
 ):
-    crud.delete_project(db, project)
-    return None
+    if not project.models:
+        crud.delete_project(db, project)
```
dominik003, line 144:
```suggestion
    if not (project.models or project.users):
        crud.delete_project(db, project)
```
Without this check, we get an error when we delete a project that has users assigned to it. It should then also not be possible to click on the button in the frontend (as you did with the models).

frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.html
```diff
@@ -39,13 +39,26 @@ <h2 id="title">Metadata</h2>
...
-      [disabled]="!form.valid"
-      (click)="updateDescription()"
-      mat-raised-button
-      id="submit"
-    >
-      Submit
-    </button>
+    <p *ngIf="!canDelete">
+      The project can't be deleted since it has linked models.
+    </p>
```
dominik003, line 44: We already had the discussion how to show the user the reason why he can't click on the delete button in https://github.com/DSD-DBS/capella-collab-manager/pull/444. But I think there was an open point from @MoritzWeber0 and he preferred the tooltip variant (and that's what we are using right now for the model deletion). So I would consistently use one variant and currently also prefer the tooltip variant more.

dominik003 (changes requested):

PR #479, opened 2022-12-20
Allow deletion of t4c models from the frontend

amolenaar:
Add a "Unlink" button to remove a T4C model.

There is already a backend route to _delete_ a t4c model. I left that untouched.

I could not test this completely, since I have no running T4C server.

### DIFF
--- a/frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.html
+++ b/frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.html
@@ -72,7 +72,7 @@
           Choose primary source
         </button>
         <button
-          *ngIf="!asStepper"
+          *ngIf="editing"
           (click)="unlinkT4CModel()"
           mat-flat-button
           color="warn"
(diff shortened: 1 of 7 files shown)


frontend/src/app/projects/models/model-source/t4c/manage-t4c-model/manage-t4c-model.component.html
```diff
@@ -71,6 +71,15 @@
...
         </button>
+        <button
+          *ngIf="!asStepper"
+          (click)="unlinkT4CModel()"
+          mat-flat-button
+          color="warn"
+          type="button"
+        >
+          Unlink
+        </button>
```
dominik003, line 82:
```suggestion
        <button
          *ngIf="editing"
          (click)="unlinkT4CModel()"
          mat-flat-button
          color="warn"
          type="button"
        >
          Unlink
        </button>
```
We are also able to create a new t4c instance on the `Configure model sources` page, where `asStepper` is false. I think using `editing` instead works well, since it should only be true if we are actually editing an existing t4c model.

dominik003 (changes requested):
I was not able to test the deletion because the backend route `delete` is missing (it should be placed in `projects/toolmodels/modelsources/t4c/routes.py`). The route should contain the required backend validation. I think we just need to check if there are existing pipelines for the t4c model as you did for the git models,

Issue #320, opened 2022-10-18
Move services into context directories in the frontend


MoritzWeber0:
We have many services in the `services` directory, which is a bit confusing and requires jumping between different places. Instead, components and services should be grouped according to their context, e.g. the `AuthService` should be moved from `services/auth` to `general/auth`

There should be no `services` directory anymore.

MoritzWeber0:
Most of the services have been removed over time; the left-overs are general services

PR #481, opened 2022-12-21
Restrictions on model level

MoritzWeber0:
In some cases, we want to restrict specific functionality on a model level. 
In our specific case, we want to limit the use of the `pure::variants` server for specific models. Only whitelisted models should get access to the server. While it's not possible to limit it inside of persistent sessions, only users which have access to the specific model get the `pure::variants` credentials injected. 

As part of this PR, I moved the modelService into the models context directory (#320).

One might wonder why I created another table with a 1:1 relation to models for just one column. This has several reasons: 
- In the `CapellaModel` json, I wanted to have a seperate field `restrictions` (because I'm also handling it as separate resource in the frontend). I've looked into fixed that by validators. It was not really native and brought in some complexity.
- The restrictions subresource has another permission for updates. Therefore, it's a lot clearer to add another resource and patch route, which can be limited to administrators only. Otherwise, there would be some advanced handling in the patch route of models, which is not where I'd expect it. 
- It's modular. Code related to restrictions is placed in the restrictions module. When we want to get rid of it in the future, it's easier to delete.
- I expect that more restrictions will pop up in the future.


backend/capellacollab/core/database/models.py
```diff
@@ -1,12 +1,15 @@
 # SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
 # SPDX-License-Identifier: Apache-2.0
 
+# pylint: skip-file
```
amolenaar, line 4: Why skip this file? It's no different from any other model file, right?

backend/capellacollab/projects/toolmodels/restrictions/injectables.py
```diff
@@ -0,0 +1,22 @@
...
+
+from .. import crud
+from ..injectables import get_existing_capella_model
+from ..models import DatabaseCapellaModel
+from .models import ToolModelRestrictions
+
+
+def get_model_restrictions(
+    model: DatabaseCapellaModel = Depends(get_existing_capella_model),
+) -> ToolModelRestrictions:
```
amolenaar, line 21: Just wondering: where do we check if the user has access to this model?

backend/capellacollab/projects/toolmodels/restrictions/models.py
```diff
@@ -0,0 +1,38 @@
...
+
+
+class ToolModelRestrictions(BaseModel):
+    # If true, access to the specific resource is granted for a model.
+    # If false, the access is not allowed.
+
+    allow_pure_variants: bool
+
+    class Config:
+        orm_mode = True
```
amolenaar, line 25: Still I'm wondering if this justified a separate table. We could also have a field `capabilities`/`features` or something like that, as an array of extra stuff to enable for the model.

backend/capellacollab/sessions/routes.py
```diff
@@ -364,6 +354,47 @@ def request_persistent_session(
     return response
 
 
+def determine_pure_variants_configuration(
+    db: Session, user: DatabaseUser, tool: Tool, warnings: list[Message]
```
amolenaar, line 358: It would be nicer, more functional, if `warnings` is returned from the function.

Idealy all parameters are input. Try to avoid in/out parameters

amolenaar (approved):
I see a few improvements.

I think with those 1:1 relations between database entities, the model gets more complicated than it should.

However, this is something we can improve upon when it starts to hurt.

backend/capellacollab/core/database/models.py
MoritzWeber0: This file imports all modules for models and every line is therefore an unused import. It's simpler to just skip the whole file instead of skipping each file manually.

backend/capellacollab/projects/toolmodels/restrictions/models.py
MoritzWeber0: I described it in the PR description. Have you seen it? 
The option with postgres arrays is something that I didn't consider, but it's a better solution indeed.

backend/capellacollab/projects/toolmodels/restrictions/injectables.py
MoritzWeber0: The routes which uses the injectable require the ADMIN permission. Admins have access to all models.

backend/capellacollab/sessions/routes.py
MoritzWeber0: Good point. I've changed it :)

Issue #485, opened 2022-12-28
Use http instead of telnet connection type for capella 6.x.x backups


dominik003:
Since version 6.0.0, the `telnet` connection type previously used in our backup containers is deprecated and should no longer be used in the newer version. Therefore, we need to adjust the backup container environment to use the `telnet` connection type only for capella version 5.x.x and the new `http` type for capella version 6.x.x (and all successor versions). In case of the `http` type we have to specify `HTTP_LOGIN`, `HTTP_PASSWORD` and `HTTP_PORT` as additional environment variables for the backup container. The `HTTP_PORT` can be taken from the already existing Experimental REST API field, and for the login and password we need to add two additional fields.

PR #486, opened 2022-12-28
Add connection type for backup containers

dominik003:
This PR fixes the problem that for capella 6.0.0 our backups stopped working. This is done by statically setting the backup connection type to `telnet` and setting `ENABLED_CDO` to `true` for the t4c server. In a follow-up PR, we should change the connection type to `http` for capella 6.x.x, since `telnet` is deprecated (as described in #485).

This PR must first be merged with https://github.com/DSD-DBS/capella-dockerimages/pull/82 and the `ENABLED_CDO` flag for the t4c server must be set to `true`.


MoritzWeber0 (approved):
LGTM, but we should bump the Capella Dockerimages submodule version before merging.

PR #489, opened 2022-12-29
Remove configuration abstraction and include configuration validation

dominik003:
This removes the ConfigDict and ConfigList abstraction and now simply uses `yaml.safe_load()` to load the configuration. In addition, the configuration is now validated against a json schema to make sure everything is well typed.


backend/config/config_schema.json
```diff
@@ -0,0 +1,178 @@
...
+      "type": "object",
+      "properties": {
+        "registry": {
+          "type": "string"
+        }
+      }
+    },
+    "operators": {
+      "type": "object",
+      "deprecated": true,
```
MoritzWeber0, line 14: Should `operators` be deprecated?

backend/config/config_schema.json
```diff
@@ -0,0 +1,178 @@
...
+        },
+        "docker": {
+          "type": "object",
+          "deprecated": true,
+          "properties": {
+            "mountVolume": { "type": "string" },
+            "portRange": { "type": "string" },
+            "containerHost": { "type": "string" }
+          }
+        },
```
MoritzWeber0, line 28: Not just deprecated, the feature is removed, so we should also remove it from the schema.

backend/config/config_schema.json
dominik003: I marked it as deprecated since #440 removes the unnecessary abstraction since we no longer use Docker.

backend/config/config_schema.json
MoritzWeber0: Yes, it shoud be removed there as well. Do you know if there are ways to generate the `config_template` out of the schema (so that we don't have to keep to files up to date)?

backend/config/config_schema.json
MoritzWeber0: But it's also not supported anymore, right? Deprecated means for me: Still works, but please change it. When it doesn't work anymore, we should not accept it.

backend/config/config_schema.json
dominik003: There should be tools for that, but I'm not sure if it's not better to adjust it manually. I mean, we don't really have configuration changes that often (and adding individual attributes to the schema should be easy too). Besides, when maintaining it manually, we can extend the schema from time to time and increase the expressiveness (e.g. I specified an enum for the `storageAccessMode` or the `capabilities`).

backend/capellacollab/config/config_schema.json
```diff
@@ -0,0 +1,168 @@
+{
+  "type": "object",
```
MoritzWeber0, line 2:
```suggestion
  "type": "object",
  "additionalProperties": false
```

backend/capellacollab/config/loader.py
```diff
@@ -11,34 +12,55 @@
...
-    pathlib.Path("/etc/capellacollab/config.yaml"),
+    pathlib.Path("/etc/capellacollab") / "config.yaml",
 ]
 
-fallback_locations: list[pathlib.Path] = [
+config_fallback_locations: list[pathlib.Path] = [
     pathlib.Path(__file__).parents[2] / "config" / "config_template.yaml",
 ]
 
+json_schema_locations = [
```
MoritzWeber0, line 26: This is exactly one file and the path can be hardcoded (you can remove some logic, e.g., the for loop). 

Proposal: 

```py
def load_config_schema() -> dict:
    return json.load((pathlib.Path(__file__).parents[0] / "config_schema.json").open())
```

backend/capellacollab/config/__init__.py
```diff
@@ -3,133 +3,19 @@
...
-    __iadd__ = _unsupported
-    __imul__ = _unsupported
-
-
-config_tmp = loader.load_yaml()
-if config_tmp:
-    config = ConfigDict(config_tmp)
-else:
-    config = ConfigDict({})
+validate(config, config_schema)
```
MoritzWeber0, line 21: The stack trace shouldn't print the whole schema.

backend/capellacollab/config/config_schema.json
```diff
@@ -0,0 +1,162 @@
+{
```
MoritzWeber0, line 1: additionalProperties needs to be added to all layers.

MoritzWeber0 (commented):
Looks good! Thanks for the implementation. Only three (small) comments.

backend/capellacollab/config/config_schema.json
MoritzWeber0, line 1: We have to mark required properties with the `required` flag. Otherwise, it doesn't complain about missing properties.

MoritzWeber0:
Comments were addressed here: #534. When #534 is approved and merged, this PR can be merged either.

MoritzWeber0 (approved):

PR #516, opened 2023-01-13
Add enable loki flag to values.yaml

### LINKED ISSUE #507: Add enableLoki flag to values.yaml
We should add a flag that allows to disable/enable loki. This flag should ensure that we only need the dependent loki chart (and minio used from the loki chart), and also we add everything related to promtail (configmap, sidecarcontainers, etc.) only when it is set.

dominik003:
This allows easy disabling of all loki related Kubernetes objects, which should significantly increase deployment speed locally when logging stuff is not needed.

Resolves #507


backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -507,13 +505,14 @@ def _create_deployment(
...
+        if loki_enabled:
+            volumes.append(
+                client.V1Volume(
+                    name="prom-config",
+                    config_map=client.V1ConfigMapVolumeSource(name=name),
+                )
             )
-        )
+
         promtail_volume_mounts.append(
```
MoritzWeber0, line 516: This shouldn't be executed if loki_enabled is false, right?

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -786,14 +786,18 @@ def _create_config_map(
                 "promtail.yaml": yaml.dump(
                     {
                         "server": {
-                            "http_listen_port": promtail_server_port,
+                            "http_listen_port": cfg["promtail"]["serverPort"],
```
MoritzWeber0, line 789: Not related to this PR, but the function name is a bit confusing. I'd rename it `_create_promtail_configmap` as it's really promtail specific.

helm/templates/backend/backend.deployment.yaml
```diff
@@ -28,6 +28,9 @@ spec:
         - name: config
           configMap:
             name: {{ .Release.Name }}-backend
+        - name: logs
+          emptyDir: {}
+        {{ if .Values.loki.enabled }}
```
MoritzWeber0, line 33: Shouldn't the `name: logs` part also be conditional? If yes, it has to be changed for all occurrences.

MoritzWeber0 (approved):

backend/capellacollab/sessions/operators/k8s.py
dominik003: I refactored this part since there was also an issue that we do not capture read only session logs. So now there is  a more clear separation and everything promtail related is now placed inside the `if loki_enabled` block.

backend/capellacollab/sessions/operators/k8s.py
dominik003: Yes totally agree with that.  Applied your naming suggestion!

helm/templates/backend/backend.deployment.yaml
dominik003: Yes in theory this also needs to be conditional but I thought this would get a bit messy. So you would prefer adjusting all loki related things (e.g., also adjust the guacamole container volume mounts or the guacd command)?

MoritzWeber0: In general: Yes. For the entrypoint, there is another change needed. Right now, we redirect stdour and stderr into the file - which doesn't print them in the kubernetes logs anymore. It would be better to still have it in the kubernetes logs. When this is implemented, we don't need a separation. 

More information: https://stackoverflow.com/a/692407

MoritzWeber0 (approved):

PR #521, opened 2023-01-13
Restrict read-only sessions

amolenaar:
Restrict read-only sessions to one per (project + tool version).


MoritzWeber0:
I've planned the review for today or tomorrow.

backend/tests/test_sessions_routes.py
```diff
@@ -274,6 +277,34 @@ def test_no_readonly_session_as_user(client, db, user, kubernetes):
     assert not kubernetes.sessions
 
 
+def test_create_readonly_session_as_user(client, db, user, kubernetes):
```
MoritzWeber0, line 280: Some pylint issues: 

function already defined line 211pylint(function-redefined)
Redefining name 'user' from outer scope (line 194)pylint(redefined-outer-name)
Redefining name 'kubernetes' from outer scope (line 181)pylint(redefined-outer-name)

backend/tests/test_sessions_routes.py
```diff
@@ -274,6 +277,34 @@ def test_no_readonly_session_as_user(client, db, user, kubernetes):
...
+    _tool, version = next(
```
MoritzWeber0, line 281:
```suggestion
    _, version = next(
```

`_tool` is not used.

backend/tests/test_sessions_routes.py
MoritzWeber0, line 280:
```suggestion
def test_create_multiple_readonly_sessions_as_user(client, db, user, kubernetes):
```

I'd choose a more describing name here.

MoritzWeber0 (approved):

backend/tests/test_sessions_routes.py
amolenaar: I left it as `_tool`, so you know what it actually was. I think that makes it more clear to the user. `_` doesn't give you any information.

PR #523, opened 2023-01-16
Rename namespaces and helm release

MoritzWeber0:


dominik003 (commented):
There are two places where `t4c-manager` is still used which I would probably also remove in this PR. First, inside the `values.yaml` for `backend.authentication.oauth.endpoints.wellknown` (either remove the value since you set it during the helm command or replace it with the new one) and second inside the `sessions.networkpolicy.yaml` (the name of it).
In addition, `t4c-sessions` is also still "used" at four different places where I would also exchange all the "usages" by `collab-sessions`

MoritzWeber0:
> There are two places where `t4c-manager` is still used which I would probably also remove in this PR. First, inside the `values.yaml` for `backend.authentication.oauth.endpoints.wellknown` (either remove the value since you set it during the helm command or replace it with the new one) and second inside the `sessions.networkpolicy.yaml` (the name of it). In addition, `t4c-sessions` is also still "used" at four different places where I would also exchange all the "usages" by `collab-sessions`

Good catch, especially the network policy! Done.

dominik003 (approved):

PR #532, opened 2023-01-17
Install different psycopg2 library depending on environment

MoritzWeber0:


dominik003 (approved):
Running locally and `make deploy` both work. So looks good to me!

PR #542, opened 2023-01-23
Move role verificator to auth module

dominik003:
This PR moves RoleVerficiation and ProjectRoleVerification from the database submodule of authentication to injectables.py. In addition, the current import uses of the two classes have been adapted to the correct format. Finally, the unused deprecated functions have been removed.


backend/capellacollab/core/authentication/injectables.py
```diff
@@ -0,0 +1,112 @@
...
+    def __call__(
+        self,
+        token=fastapi.Depends(jwt_bearer.JWTBearer()),
+        db=fastapi.Depends(database.get_db),
+    ) -> bool:
+        role = users_crud.get_user_by_name(
+            db=db, username=helper.get_username(token)
+        ).role
+        if (
+            role == users_models.Role.USER
```
amolenaar, line 33:
```suggestion
            role != users_models.Role.ADMIN
```

With this statement I do not have to know there are only 2 roles in the system.

backend/capellacollab/core/authentication/injectables.py
```diff
@@ -0,0 +1,112 @@
...
+
+        project = projects_crud.get_project_by_slug(db, project_slug)
+        project_user = projects_users_crud.get_user_of_project(
+            db, project, user
+        )
+
+        # Check role
+        if not project_user or self.roles.index(
+            project_user.role
+        ) < self.roles.index(self.required_role):
```
amolenaar, line 85: ... or use a line similar to this in `RoleVerification.__call__` too.

amolenaar (approved):
One small improvement regarding roles. I think the code can be decoupled from the actual role.

backend/capellacollab/core/authentication/injectables.py
dominik003: Well spotted. I think this is sufficient for now, as it at least prevents the case where we add another user role that would then have come through this check.

backend/capellacollab/core/authentication/injectables.py
dominik003: Applied your suggestion from above.

PR #543, opened 2023-01-23
Breadcrumbs for page navigation

amolenaar:
Clean up the NavBar service. We no longer need to maintain the title and subtitle there.

In fact, the Navbar Service is not a navbar service at all. Instead it's a service that configures if certain static parts are available on the page. Therefore I renamed it to PageLayoutService and moved it up the directory structure accordingly.


frontend/src/app/general/breadcrumbs/breadcrumbs.service.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  public updatePlaceholder(placeholders: Data) {
+    this.placeholders = {
+      ...this.placeholders,
+      ...placeholders,
+    };
+    this.updateBreadcrumbs(this.router);
+  }
+
+  private updateBreadcrumbs(router: Router) {
```
amolenaar, line 44: Somehow this does not work when using `this.router`. updateBreadcrumbs is called from the constructor and it looks like `this.router` is not yet defined.

image

MoritzWeber0: The following code snippet works for me: 

```py
  constructor(private router: Router) {
    this.router.events
      .pipe(filter((event) => event instanceof NavigationEnd))
      .subscribe(() => this.updateBreadcrumbs());
  }

  public updatePlaceholder(placeholders: Data) {
    this.placeholders = {
      ...this.placeholders,
      ...placeholders,
    };
    this.updateBreadcrumbs();
  }

  private updateBreadcrumbs() {
    const root = this.router.routerState.snapshot.root;
    this._breadcrumbs.next(breadcrumbs(root, this.placeholders));
  }
```

frontend/src/app/projects/models/model-source/git/manage-git-model/manage-git-model.component.ts
```diff
@@ -169,6 +171,7 @@ export class ManageGitModelComponent implements OnInit, OnDestroy {
             );
           }
         );
+        this.breadCrumbsService.updatePlaceholder({ gitModel: undefined });
```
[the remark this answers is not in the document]
MoritzWeber0: And you'll have to reset the placeholder here too. 

After
```
this.gitModelService.clear();
```

a new line 

```
this.breadCrumbsService.updatePlaceholder({ gitModel: undefined });
```

could be added.

frontend/src/app/projects/models/model-wrapper/model-wrapper.component.ts
```diff
@@ -40,10 +43,15 @@ export class ModelWrapperComponent implements OnInit, OnDestroy {
...
       });
+
+    this.breadcrumbSubscription = this.modelService._model.subscribe((model) =>
+      this.breadcrumbService.updatePlaceholder({ model })
+    );
   }
 
   ngOnDestroy(): void {
     this.subscription?.unsubscribe();
+    this.breadcrumbSubscription?.unsubscribe();
```
MoritzWeber0, line 54:
```suggestion
    this.breadcrumbSubscription?.unsubscribe();
```

Unsubscribing is not enough, you also need to reset the placeholder.

frontend/src/app/projects/project-wrapper/project-wrapper.component.ts
```diff
@@ -49,7 +51,10 @@ export class ProjectWrapperComponent implements OnInit, OnDestroy {
         )
       )
       .subscribe({
-        next: (project: Project) => this.projectService._project.next(project),
+        next: (project: Project) => {
+          this.breadcrumbsService.updatePlaceholder({ project });
+          return this.projectService._project.next(project);
```
MoritzWeber0, line 56:
```suggestion
         this.projectService._project.next(project);
```

The signature of `next` has `void` as return type. Is there a reason why you return something here?

frontend/src/app/projects/project-wrapper/project-wrapper.component.ts
```diff
@@ -49,7 +51,10 @@ export class ProjectWrapperComponent implements OnInit, OnDestroy {
         )
       )
       .subscribe({
-        next: (project: Project) => this.projectService._project.next(project),
+        next: (project: Project) => {
+          this.breadcrumbsService.updatePlaceholder({ project });
```
MoritzWeber0, line 55: Please reset the placeholder in `ngOnDestroy`

frontend/src/app/app-routing.module.ts
```diff
@@ -122,84 +160,114 @@ const routes: Routes = [
...
+                data: { breadcrumb: 'Git' },
                 children: [
                   {
                     path: '',
+                    data: { breadcrumb: undefined },
                     component: GitSettingsComponent,
                   },
                   {
                     path: 'instances/:id',
+                    data: { breadcrumb: 'Edit' },
```
MoritzWeber0, line 249: For consistency, I'd show the name instead. We do that for the other router too.

frontend/src/app/app-routing.module.ts
```diff
@@ -122,84 +160,114 @@ const routes: Routes = [
...
                     component: T4CSettingsComponent,
                   },
                   {
                     path: 'create',
+                    data: { breadcrumb: 'New' },
                     component: EditT4CInstanceComponent,
                   },
                   {
                     path: 'instance/:instance',
+                    data: { breadcrumb: 'Edit' },
```
MoritzWeber0, line 270: For consistency, I'd show the name instead. We do that for the other router too.

MoritzWeber0 (changes requested):
A really good improvement. Is feels good that the `navBarService` has been removed.
I have some comments, mostly about data inconsistency of the breadcrumbs data, but these are only small changes.

frontend/src/app/projects/project-wrapper/project-wrapper.component.ts
amolenaar: It's a `void` method? Okay. i just returned because the original did too.

frontend/src/app/general/breadcrumbs/breadcrumbs.service.ts
amolenaar: Why didn't it work with me :confused:

MoritzWeber0 (approved):

amolenaar:
:partying_face:

PR #550, opened 2023-01-24
Bump minimum Python version to 3.11

### LINKED ISSUE #549: Bump minimum Python version to 3.11 (backend)

MoritzWeber0:
This PR contains the following changes: 

- Updated all references from Python 3.9/3.10 to 3.11
- Set minimum Python version to 3.11 for all used libraries
- Replace `t.List` with `list`
- Replace `t.Optional[x]` with `x | None`
- Replace `t.Union[x, y]` with `x | y`
- Replace `t.Dict` with `dict`
- Remove `from __future__ import annotations` where possible. Unfortunately, PEP 563 has been put on hold indefinitely and was not introduced in Python 3.11, so we still need the import in some cases.

Resolves #549

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v2
         with:
-          python-version: "3.10"
+          python-version: "3.11"
       - run: pip install mkdocs-material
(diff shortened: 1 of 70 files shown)


amolenaar (approved):
:shipit: 

pyupgrade did not have any additional suggestions.

PR #574, opened 2023-01-31
Jupyter notebook support

amolenaar:
Add support for Jupyter notebooks

- [x] Since jupyter works for persistent sessions only, it's not needed to configure a read-only and backup image -> make readonly and backup non-required fields.
- [x] This setup works with (basic) k8s ingress routes. For openshift we need a different type of route.
- [x] host is "localhost"
- [x] No Metrics at the moment (can't automaticallty terminate idle session)
- [x] Harden the image (no ssh)
- [x] Memory and CPU usage are now similar to Capella's
- [x] Move jupyter image to capella-dockerimages repo -> https://github.com/DSD-DBS/capella-dockerimages/pull/108
- [x] Remove openshift routes on container shutdown

Changes include:

- Add a "Jupyter notebook" integration
- readonly and backup docker images are optional
- Jupyter notebooks are exposed via Ingress rules on k8s
- On openshift clusters, jupyter endpoints are exposed via a custom domain

Postponed/to be picked up separately:

- [ ] Create an abstraction for deploying guacamole jupiter and _your favorite tool_. -> later, maybe
- [ ] The jupyter setup is pretty basic. It installs py-capellambse. We may want to provide more tooling out of the box -> let's see. Separate PR anyway.
- [ ] Only one persistent session can be running per user at any time. You can't run both a capella and a jupyter session. That would be nice, I suppose. -> one per tool, separate PR
- [ ] Logging may need some configuration

Requires: https://github.com/DSD-DBS/capella-dockerimages/pull/108


helm/templates/backend/backend.serviceaccount.yaml
```diff
@@ -35,6 +35,9 @@ rules:
 - apiGroups: [""]
   resources: ["secrets"]
   verbs: ["create", "delete"]
+- apiGroups: ["networking.k8s.io"]
+  resources: ["ingresses"]
+  verbs: ["create", "delete"]
```
dominik003, line 40: Didn't try it yet locally or in the cluster but you have to also distinguish the openshift/ non openshift case here. So just like in our routing templates you can do the following:
```
{{- if eq .Values.cluster.kind "Kubernetes" }}
- apiGroups: ["networking.k8s.io"]
  resources: ["ingresses"]
  verbs: ["create", "delete"]
{{- else }}
- apiGroups: ["route.openshift.io/v1"]
  resources: ["route"]
  verbs: ["create", "delete"]
{{- end }}
```

images/jupyter/Dockerfile
```diff
@@ -0,0 +1,43 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
```
dominik003, line 1: Please remove this and add make command for `capella-docker-images` in the Makefile.

images/jupyter/docker-entrypoint.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
dominik003, line 1: Please remove this as it will be contained in the capella-docker-images

backend/capellacollab/sessions/crud.py
```diff
@@ -56,6 +56,8 @@ def create_session(db: Session, session: DatabaseSession):
     db.add(session)
     db.commit()
     db.refresh(session)
+    session.state = "New"
+    session.last_seen = "UNKNOWN"
```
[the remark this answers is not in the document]
amolenaar: I see. This is indeed a dirty trick to avoid a type case, and appearently it works, since `GetSessionsResponse` acts more like an interface, than an actual (type) class.

backend/capellacollab/sessions/routes.py
```diff
@@ -361,29 +359,57 @@ def request_persistent_session(
...
-        tool_name=tool.name,
-        version_name=version.name,
-        password=rdp_password,
-        docker_image=docker_image,
-        t4c_license_secret=t4c_license_secret,
-        t4c_json=t4c_json,
-        pure_variants_license_server=pv_license_server_url,
-        pure_variants_secret_name=pure_variants_secret_name,
-    )
+    if tool.integrations and tool.integrations.jupyter:
```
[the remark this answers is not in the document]
amolenaar: That would tie our implementation to a user configurable (human readable) tool name. That's not a good idea.

backend/tests/test_sessions_routes.py
```diff
@@ -38,7 +38,9 @@
     get_natures,
     get_versions,
 )
-from capellacollab.tools.models import Tool
+from capellacollab.tools.integrations.crud import update_integrations
+from capellacollab.tools.integrations.models import PatchToolIntegrations
+from capellacollab.tools.models import Tool, Version
```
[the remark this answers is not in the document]
amolenaar: What do you mean?

helm/templates/backend/backend.serviceaccount.yaml
amolenaar: Nice find!

MoritzWeber0:
If not done yet: Please test it in our OpenShift development cluster. 

- [ ] Test functionality in OpenShift development cluster

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -132,10 +159,50 @@ def start_persistent_session(
...
+        }
+
+        session_parameters = self._start_session(
+            image=docker_image,
+            username=username,
+            session_type="persistent",
+            tool_name=tool_name,
+            version_name=version_name,
+            environment=environment,
+            ports={"http": 8888},
```
dominik003, line 192: You use the port `8888` at five different places here. So I think we should add a `JUPYTER_PORT` at the top of the k8s operator and then simply use this here. This makes it a lot easier to make changes to it later (e.g., if we want to make it configurable or if we want to change the port for some reason)

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -185,23 +257,41 @@ def _start_session(
...
+        if self.openshift:
+            if dep_status := self._delete_openshift_route(_id):
+                log.info(
+                    "Deleted route %s with status %s", _id, dep_status.status
+                )
+        if dep_status := self._delete_ingress(_id):
+            log.info(
+                "Deleted ingress %s with status %s", _id, dep_status.status
+            )
+
```
dominik003, line 294:
```suggestion
        if self.openshift:
            if dep_status := self._delete_openshift_route(_id):
                log.info(
                    "Deleted route %s with status %s", _id, dep_status.status
                )
        else:
            if dep_status := self._delete_ingress(_id):
                log.info(
                    "Deleted ingress %s with status %s", _id, dep_status.status
                )
```
Without this, we would always produce an error on openshift

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -442,10 +532,19 @@ def _export_attrs(
...
+            port = {ports["rdp"]}
+        elif "http" in ports:
+            port = {ports["http"]}
+        else:
+            port = {ports.values()}
+
         return {
             "id": deployment.to_dict()["metadata"]["name"],
-            "ports": {3389},
+            "ports": port,
```
dominik003, line 547: What is the purpose of this? As far as I can see `"ports"` is only used in one place to get the rdp port. So why do we need the other things here?

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -668,28 +776,91 @@ def _create_service(
...
+                                )
+                            ]
+                        ),
+                    )
+                ]
+            ),
+        )
+        return client.NetworkingV1Api().create_namespaced_ingress(
+            namespace, ingress
+        )
```
dominik003, line 833: Please move the `NetworkingV1Api` to the `__init__` (e.g., `self.v1_networking = client.NetworkingApi(api_client=self.client)`), so we do not create a new one each time we create a resource

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -668,28 +776,91 @@ def _create_service(
...
+                "port": {
+                    "targetPort": port_number,
+                },
+                "tls": {
+                    "termination": "edge",
+                    "insecureEdgeTerminationPolicy": "Redirect",
+                },
+            },
+        }
+        dyn_client = DynamicClient(self.client)
```
dominik003, line 858: Please also move this client to the `__init__`. If you only really need the `v1_routes` you could also add `self.v1_routes` instead of the dynamic client

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -849,6 +1020,26 @@ def _delete_service(self, name: str) -> client.V1Status | None:
             log.exception("Error deleting service with name: %s", name)
             return None
 
+    def _delete_ingress(self, name: str) -> client.V1Status | None:
+        try:
+            return client.NetworkingV1Api().delete_namespaced_ingress(
```
dominik003, line 1025:
```suggestion
            return self.v1_networking.delete_namespaced_ingress(
```

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -22,6 +22,8 @@
 import yaml
 from kubernetes import client
 from kubernetes.client import exceptions
+from openshift.dynamic import DynamicClient
+from openshift.dynamic.exceptions import ResourceNotFoundError
```
dominik003, line 26: Please adjust the imports here:
- `from openshift import dynamic` and then use `dynamic.DynamicClient`
- `from openshift.dynamic import exceptions as dynamic_exceptions` and then use `dynamic_exceptions.ResourceNotFoundError`

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -849,6 +1020,26 @@ def _delete_service(self, name: str) -> client.V1Status | None:
...
+            log.exception("Error deleting ingress with name: %s", name)
+            return None
+
+    def _delete_openshift_route(self, name: str) -> client.V1Status | None:
+        try:
+            dyn_client = DynamicClient(self.client)
+            v1_routes = dyn_client.resources.get(
+                api_version="route.openshift.io/v1", kind="Route"
+            )
+            return v1_routes.delete(name=name, namespace=namespace)
```
dominik003, line 1038: Please use the `self.xy` that you should initialise in the `__init__` here

dominik003 (changes requested):
Finished the review of the k8s operator but as the backend is currently not starting I can't really test the jupyter sessions. So please fix the backend issue first.

backend/capellacollab/sessions/operators/k8s.py
amolenaar: Discussed with @dominik003. Let's leave it for now. We need more cleanup in the k8s operator anyway.

backend/capellacollab/sessions/operators/k8s.py
amolenaar: Let's not move it, since it breaks non-openshift clusters.

amolenaar:
> If not done yet: Please test it in our OpenShift development cluster.
> 
>     * [x]  Test functionality in OpenShift development cluster

Works.

MoritzWeber0 (commented):
- [x] Add release notes
- [x] Restructure / Add user documentation. Right now, it'd focused on Capella sessions.

MoritzWeber0 (commented):
Great addition, our list of supported tools starts growing 🥳

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -668,28 +782,90 @@ def _create_service(
...
         )
         return self.v1_core.create_namespaced_service(namespace, service)
 
+    def _create_ingress(self, id, path: str, port_number: int):
+        ingress = client.V1Ingress(
+            api_version="networking.k8s.io/v1",
+            kind="Ingress",
+            metadata=client.V1ObjectMeta(
+                name=id,
+                annotations={"ingress.kubernetes.io/ssl-redirect": "false"},
```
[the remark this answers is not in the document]
amolenaar: Nice catch. I missed fixing this (temporary) issue.

backend/capellacollab/sessions/routes.py
```diff
@@ -361,30 +359,59 @@ def request_persistent_session(
...
+            db,
+            schema.WorkspaceType.PERSISTENT,
+            session,
+            owner,
+            rdp_password,
+            tool,
+            version,
+            None,
+            t4c_password,
+        )
```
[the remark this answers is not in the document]
amolenaar: rdp or guacamole? So far we've used `guacamole` as the term for this kind of connection.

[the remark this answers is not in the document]
MoritzWeber0: `guacamole` is also good.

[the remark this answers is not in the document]
amolenaar: This change cleaned up more code than I initially expected. :+1:

MoritzWeber0 (approved):
Is ready for merge. However, I'd wait until https://github.com/DSD-DBS/capella-dockerimages/pull/108 is merged.

Thanks again for the implementation, @amolenaar! 😃

dominik003 (approved):

dominik003:
:shipit:

PR #578, opened 2023-01-31
Restructure project and model service

dominik003:
Based on #571


frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.ts
```diff
@@ -6,20 +6,15 @@
 import {
```
MoritzWeber0, line 6: There is a small issue, to reproduce: 

1. Open a project from the project overview
2. The "Submit" button is disabled
3. One wants to change the description
4. The "Submit" button is still disabled, so it can not be submitted.

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.ts
```diff
@@ -59,9 +63,19 @@ export class ProjectUserSettingsComponent implements OnChanges {
...
+    private toastService: ToastService,
+    private projectService: ProjectService
   ) {}
 
+  ngOnInit(): void {
+    this.projectService.project
+      .pipe(untilDestroyed(this), filter(Boolean))
+      .subscribe((project) => {
+        this.projectSlug = project!.slug;
+        this.projectName = project!.name;
```
MoritzWeber0, line 75: In this case, I'd just save the project itself. Makes it a lot easier if we want to extend it in the future, otherwise this method will grow for each new attribute.

frontend/src/app/projects/project-wrapper/project-wrapper.component.ts
```diff
@@ -33,60 +32,25 @@ export class ProjectWrapperComponent implements OnInit, OnDestroy {
...
+      .pipe(untilDestroyed(this))
+      .subscribe((project) =>
+        this.breadcrumbsService.updatePlaceholder({ project })
+      );
   }
 
   ngOnDestroy(): void {
-    this.projectSubscription?.unsubscribe();
-    this.modelsSubscription?.unsubscribe();
-    this.projectUserSubscription?.unsubscribe();
```
MoritzWeber0, line 87: The `projectUserSubscription`, `modelsSubscription` and `projectSubscription` are still defined, but don't seem to be used.

MoritzWeber0 (changes requested):
I couldn't test everything. As this is a major restructuring, could go through all of the workflows and test if they still work?

frontend/src/app/projects/project-wrapper/project-wrapper.component.ts
dominik003: I searched for all usages of subscriptions and changed all of them to use untilDestroyed(this)`, so also removed these unused subscriptions here.

frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.ts
dominik003: Yes, this is a "common" problem that I also encountered when simplifying the `edit-git-instance` thing. The problem here is that the slug validator throws an error because the project slug already exists, so the form is not valid. The simplest and also a good solution to this problem is to pass a project to the asynchronous validator, which is then excluded/ignored. I have applied this and now it works.

dominik003:
I tested the following work flows:
- start a capella t4c session (only issue is that `No session token was generated for your session.` was displayed but not sure if that is the usual case (never actually paid attention to this in the local dev environment)
- created a test mode, linked it to a git lab instance (with integration), and loaded the diagram cache
- create a project with a description, added one user during the creation, and added one empty (uninitialised) model during this step

MoritzWeber0 (approved):

PR #590, opened 2023-02-06
Move the fileservice into the Collab Manager base

MoritzWeber0:
Move the fileservice into the backend - Python code is executed via `kubectl exec`.
In addition, it fixes the upload & download of files.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -4,13 +4,27 @@
 ARG BASE_IMAGE=python:3.11-bullseye
 FROM $BASE_IMAGE
 
+SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
+ENV SHELL=/bin/bash
+
 RUN apt-get update && \
     apt-get upgrade --yes && \
     apt-get install --yes \
     ca-certificates \
     unzip \
     libpq-dev \
+    curl \
     && rm -rf /var/lib/apt/lists/*
+
+# Install kubectl if not available in base image
+# kubectl is needed for the list files endpoint
+RUN dpkg -s kubectl || kubectl_installed="no"; \
+    if [[ "$kubectl_installed" == "no" ]]; \
+    then \
+    curl -fsSLo /usr/share/keyrings/kubectl.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg; \
+    echo "deb [signed-by=/usr/share/keyrings/kubectl.gpg] https://apt.kubernetes.io/ kubernetes-xenial main" | tee /etc/apt/sources.list.d/kubernetes.list; \
+    apt-get update && apt-get install -y kubectl && rm -rf /var/lib/apt/lists/*; \
+    fi
 
 EXPOSE 8000
 COPY . /tmp/backend
--- a/backend/capellacollab/sessions/operators/__init__.py
+++ b/backend/capellacollab/sessions/operators/__init__.py
@@ -4,15 +4,14 @@
 
 from capellacollab.sessions.operators.k8s import KubernetesOperator
 
+operator = None
 
-class OperatorLoader:
-    operator = None
 
-    @classmethod
-    def load_operator(cls):
-        cls.operator = KubernetesOperator()
-        cls.operator.load_config()
+def load_operator():
+    global operator
+    operator = KubernetesOperator()
+    operator.load_config()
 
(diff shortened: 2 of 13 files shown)


backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -885,6 +854,92 @@
...
+                    "--stdin",
+                    pod_name,
+                    "--container",
+                    _id,
+                    "--",
+                    "python",
+                    "-",
+                    directory,
+                    json.dumps(show_hidden),
+                ],
```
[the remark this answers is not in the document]
MoritzWeber0: `show_hidden` has type bool. The type is enforced by pydantic.

backend/Dockerfile
```diff
@@ -4,14 +4,28 @@
...
+# Install kubectl if not available in base image
+# kubectl is needed for the list files endpoint
+RUN dpkg -s kubectl || kubectl_installed="no"; \
+    if [[ "$kubectl_installed" == "no" ]]; \
+    then \
+    curl -fsSLo /usr/share/keyrings/kubectl.gpg https://packages.cloud.google.com/apt/doc/apt-key.gpg; \
+    echo "deb [signed-by=/usr/share/keyrings/kubectl.gpg] https://apt.kubernetes.io/ kubernetes-xenial main" | tee /etc/apt/sources.list.d/kubernetes.list; \
+    apt-get update && apt-get install -y kubectl && rm -rf /var/lib/apt/lists/*; \
+    fi
+
```
amolenaar, line 28: Why not "just" install it?

backend/capellacollab/sessions/operators/__init__.py
```diff
@@ -4,8 +4,15 @@
 
 from capellacollab.sessions.operators.k8s import KubernetesOperator
 
-OPERATOR = KubernetesOperator()
+
+class OperatorLoader:
+    operator = None
+
+    @classmethod
```
amolenaar, line 11: Why the class? Just use a module variable. One class with a class method is not pythonic IMHO.

backend/capellacollab/sessions/operators/helper.py
```diff
@@ -0,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
+
+import inspect
+import typing as t
+
+
+def get_source_of_python_function(function: t.Callable):
```
amolenaar, line 8: I'd move this "helper" into k8s.py.

"helper" is a bad name for any module.

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -885,6 +857,92 @@ def _get_pods(self, label_selector: str) -> list[client.V1Pod]:
...
```
amolenaar, line 926: Hmm.. At this point we could just as well use `ls` :smile: (but let's not change that now)

amolenaar (approved):
Nice improvement. I like it that we finally are rid of the OPERATOR global.

I suppose we can make an injectable from `get_operator()` at some point.

I added a few suggestions.

backend/Dockerfile
MoritzWeber0: In our DB environment pipelines, we don't have internet access. The install command would fail (`https://apt.kubernetes.io/` is not reachable). But it is already installed in our Base image, we can skip the installation.

backend/capellacollab/sessions/operators/helper.py
MoritzWeber0: The `k8s.py` has 1030 lines already and is too large imho. Also, it's not related to the k8s operator itself - it just returns the source of a python function.

backend/capellacollab/sessions/operators/k8s.py
MoritzWeber0: Parsing ls is a lot more pain then just running Python. We went there, whitespaces, symlinks, remote code execution (when running `ls` on a directory called `$(date)` for example.

backend/capellacollab/sessions/operators/__init__.py
MoritzWeber0: Fixed.

PR #600, opened 2023-02-08
Simplify git settings component

dominik003:
This PR first simplifies the settings component and in addition adds a name validation for the edit git instance component (until now we could actually have two instances with the same name)

### DIFF
--- a/frontend/src/app/settings/modelsources/git-settings/edit-git-settings/edit-git-settings.component.ts
+++ b/frontend/src/app/settings/modelsources/git-settings/edit-git-settings/edit-git-settings.component.ts
@@ -4,10 +4,10 @@
  */
 
 import { Component, OnDestroy, OnInit } from '@angular/core';
-import { FormControl, FormGroup, Validators } from '@angular/forms';
+import { FormBuilder, Validators } from '@angular/forms';
 import { ActivatedRoute, Router } from '@angular/router';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
-import { filter } from 'rxjs';
+import { filter, map } from 'rxjs';
 import { BreadcrumbsService } from 'src/app/general/breadcrumbs/breadcrumbs.service';
 import {
   GitInstance,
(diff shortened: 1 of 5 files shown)


frontend/src/app/settings/modelsources/git-settings/git-settings.component.html
```diff
@@ -4,34 +4,36 @@
...
-  <div class="title" *ngIf="availableGitSettings.length">
-    Existing instances
-  </div>
-  <div class="instances" *ngIf="availableGitSettings.length">
-    <mat-card *ngFor="let instance of availableGitSettings">
-      <div class="mat-title">
-        <div class="instance-head">
-          {{ instance.name }}
-          <span class="mat-subtitle">{{ instance.type }}</span>
+  <div *ngIf="(this.gitSettingsService.gitSettings | async)?.length !== 0">
```
MoritzWeber0, line 7:
```suggestion
  <div *ngIf="(this.gitSettingsService.gitSettings | async)?.length">
```

This should do the same.

frontend/src/app/settings/modelsources/git-settings/edit-git-settings/edit-git-settings.component.ts
```diff
@@ -6,7 +6,7 @@
 import { Component, OnDestroy, OnInit } from '@angular/core';
```
MoritzWeber0, line 6: image

Doesn't seem to work. It happens when I try to edit any instance.

MoritzWeber0 (changes requested):

frontend/src/app/settings/modelsources/git-settings/edit-git-settings/edit-git-settings.component.ts
```diff
@@ -21,18 +21,19 @@ import {
...
-  gitSettingsForm = new FormGroup({
-    type: new FormControl('', Validators.required),
-    name: new FormControl('', Validators.required),
-    url: new FormControl('', Validators.required),
-    apiURL: new FormControl(''),
+  gitSettingsForm = this.fb.group({
+    type: this.fb.control('', Validators.required),
+    name: this.fb.control('', Validators.required),
+    url: this.fb.control('', Validators.required),
+    apiURL: this.fb.control(''),
```
MoritzWeber0, line 28: What's the difference between the `FormBuilder` and `FormGroup` / `FormControl`. I prefer the latter one, because no injectable is needed, so it's one line less and we use it everywhere else in the code.

dominik003: The first advantage of this approach is that we can decrease the imports at the top as it usually takes several lines to import the `FormGroup` and `FormControl` while by using the `FormBuilder` this can be reduced to just one single line.
An additional advantage of this approach is the reduction of duplicated/redundant code which this code snippet here does not yet reflect. If you look at the last commit you can see that by using the FormBuilder you can remove all the 'new FormControl(...)` or here the `this.fb.control(...)`. So for me using the FormBuilder (that also is just there to provide a convenient method to define forms) is better than directly using the 'FormGroup` and `FormControl`.

MoritzWeber0 (approved):
When the conflicts are solved, it can be merged.

PR #625, opened 2023-02-16
Disable terminate sessions locally

dominik003:


backend/capellacollab/sessions/operators/__init__.py
```diff
@@ -2,15 +2,15 @@
...
+import functools
 
-operator = None
+from capellacollab.sessions.operators.k8s import KubernetesOperator
 
 
 def load_operator():
-    global operator
-    operator = KubernetesOperator()
+    get_operator()
```
MoritzWeber0, line 11: We can get rid of the `load_operation` function. Just call `get_operator` instead.

MoritzWeber0 (approved):

backend/capellacollab/sessions/operators/__init__.py
dominik003: Removed the `load_operator` and added a comment to `__main__` that the `get_operator` is needed there to load the kubernetes config at startup

PR #630, opened 2023-02-20
Upgrade Guacamole to 1.5.0

amolenaar:
Pin the version, to avoid future surprises.

### DIFF
--- a/guacamole/Dockerfile
+++ b/guacamole/Dockerfile
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
 # SPDX-License-Identifier: Apache-2.0
 
-FROM guacamole/guacamole:1.4.0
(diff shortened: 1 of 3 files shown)


dominik003:
Its kind of weird that you now have to use `/bin/sh` and are no longer able to use `/bin/bash` I am wondering whats the reason for this is. But thats nothing we can change I assume.
Just a pre-comment on this: I think we don't really use the `guacamole: docker.io/guacamole/guacamole:1.5.0` image as we use our own guacamole image that is placed inside of `collab-manager/guacamole/Dockerfile` and this file still has the old `1.4.0` tag set. So we must adjust the tag there and check whether building the image still works.

amolenaar:
Also the guacamole installation folder has changed. We should indeed also update our customized images.

MoritzWeber0:
Marked as draft until the customised images are updated.

MoritzWeber0 (approved):

PR #638, opened 2023-02-22
Change frontend session structure

dominik003:
Resolves: #595


MoritzWeber0 (approved):

dominik003:
Tested the following session functionalities locally:
- start a capella persistent session (session was correctly started, I was able to connect to the session, and the session was displayed in the session overview)
- file browser for a started session works and I am able to download stuff
- readonly session start works

PR #650, opened 2023-02-28
 Conditionally set the ingress-class-name

amolenaar:
This is required for "normal" clusters configured with an Nginx ingress. Our local (k3d) cluster does not need it, since it's using Traefik.

I added "target" to the configuration. It's not ideal, but we need to know about this property when we launch a jupyter session.

The construct is similar to https://github.com/DSD-DBS/capella-collab-manager/blob/main/helm/templates/routing/nginx.ingress.yaml#L16-L18.


backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -810,6 +810,9 @@ def _create_ingress(self, id, path: str, port_number: int):
                 name=id,
             ),
             spec=client.V1IngressSpec(
+                ingress_class_name=None
+                if cfg["target"] == "local"
+                else "nginx",
```
MoritzWeber0, line 815: Instead deriving "nginx" from the target, which is not self-explaining, I'd directly pass the ingressClassName into the backend. In addition, it gives us the flexibility to use different ingressClassNames ("nginx" may have another name in different instances).

amolenaar:
On hindsight, it looks like it's working perfectly fine without this.

MoritzWeber0:
Feature is still needed - reopening pull request.

helm/config/backend.yaml
```diff
@@ -7,6 +7,9 @@ docker:
 
 k8s:
   namespace: {{ .Values.backend.k8sSessionNamespace }}
+  {{- if eq .Values.target "azure" }}
+  ingressClassName: nginx
+  {{- end }}
```
MoritzWeber0, line 12: I'd rather just take it from the `values.yaml` and overwrite it there. A good place would be `cluster.namespaces.sessions.ingressClassName`.

The ingressClassName is a flexible name / string. It can be named nginx (as it is in our case), but can have any other name. To support more Kubernetes instances, let's have it configurable.

MoritzWeber0 (changes requested):

helm/config/backend.yaml
amolenaar: I did not want to change Helm's `values.yml`. However putting it in the `sessions` section _does_ make sense :sweat_smile: .

MoritzWeber0: In this case, it would make sense. And when it's set to `null` per default, it doesn't have any effect on existing deployments.

MoritzWeber0 (approved):

PR #659, opened 2023-03-30
Blob authentication error handling

MoritzWeber0:
This fixes a bug where the authentication handling was erroring if a blob is used as response type in requests. The blob response has to be transformed to the correct format first.


amolenaar (approved):
:shipit:

PR #663, opened 2023-03-30
Add Gitlab CI template for image builder

MoritzWeber0:


ci-templates/gitlab/image-builder.yml
```diff
@@ -0,0 +1,108 @@
...
+    value: "0"
+    description: "Build the guacamole image?"
+  REVISION:
+    value: "main"
+    description: "Revision of Github repository"
+  DOCKER_BUILD_ARGS: "--no-cache"
+  DOCKER_BUILDKIT: "1"
+
+default:
+  image: $DOCKER_REGISTRY/base
```
amolenaar, line 24: Should we use a more commonly used base image (e.g. `debian`)?

amolenaar (changes requested):

ci-templates/gitlab/image-builder.yml
MoritzWeber0: It's as trivial, as the `docker`, `node`, `python`, `git` and `npm` have to be installed. And we don't want to install it in each pipeline run. I did a bit of research, but haven't found any matching image. Do you know an image where these tools are already preinstalled?

amolenaar: Not really. Let's keep it like this for now, then. If somebody from the community runs into this, he/she may come up with a proper solution.

amolenaar (approved):

PR #664, opened 2023-03-30
Add required resources to README

MoritzWeber0:


README.md
```diff
@@ -107,33 +107,63 @@ running in a few minutes.
...
 
-   ```sh
-   helm upgrade production -n <namespace> -f deployments/yourinstance.values.yaml helm
+1. If you want to uninstall the management portal, you can run the following comment:
+
+```sh
+helm uninstall production -n <namespace> helm
+```
+
+1. The previous command doesn't clean the sessions namespace.
```
dominik003, line 156: It seems like this does not result in

1. If you want ...
2. The previous ...

But in 1. and 1. as the shell command is not part of the actual enumeration. So you have to add the correct indentation here

dominik003 (approved):
Just one small issue otherwise looks good!

PR #671, opened 2023-04-03
fix(frontend): Make docsURL optional in env

MoritzWeber0:
A small fix which makes the docsURL environment optional. In addition, it adds a trailing slash to the docs, otherwise it doesn't load.

### DIFF
--- a/frontend/src/app/projects/project-detail/model-overview/model-complexity-badge/model-complexity-badge.component.ts
+++ b/frontend/src/app/projects/project-detail/model-overview/model-complexity-badge/model-complexity-badge.component.ts
@@ -70,7 +70,10 @@ export class ModelComplexityBadgeComponent implements OnChanges {
   }
 
   openModelComplexityBadgeDocs() {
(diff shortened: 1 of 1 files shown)


frontend/src/app/projects/project-detail/model-overview/model-complexity-badge/model-complexity-badge.component.ts
```diff
@@ -70,7 +70,10 @@ export class ModelComplexityBadgeComponent implements OnChanges {
   }
 
   openModelComplexityBadgeDocs() {
-    const docsURL = environment.docsURL || '/docs';
-    window.open(docsURL + '/projects/models/complexity_badge', '_blank');
+    let docsURL = '/docs';
+    if ('docsURL' in environment) {
+      docsURL = environment.docsURL as string;
+    }
```
amolenaar, line 76: The original line should do the same, right? If docsURL is not defined, it returns `undefined`, which will trigger the or-lhs clause: `/docs`.

PR #718, opened 2023-05-23
Add pagination and sorting for pipeline runs

MoritzWeber0:
The number of pipeline runs can reach a high number in a short time, therefore it's essential to have pagination. However, pagination is not handled by the user, instead, new pipeline runs are lazy loaded while scrolling.

Depends on https://github.com/DSD-DBS/capella-collab-manager/pull/647


backend/capellacollab/projects/toolmodels/backups/runs/crud.py
```diff
@@ -52,6 +54,17 @@ def get_scheduled_and_running_pipelines(
...
 
+def get_pipeline_runs_for_pipeline(
+    db: Session, pipeline: pipeline_models.DatabaseBackup
+):
+    return fastapi_pagination.ext.sqlalchemy.paginate(
+        db,
+        select(models.DatabasePipelineRun)
+        .where(models.DatabasePipelineRun.pipeline == pipeline)
+        .order_by(models.DatabasePipelineRun.id.desc()),
+    )
```
dominik003, line 65: First, I would not provide the whole pipeline here but rather just the `pipeline_id`. So then having,
`def get_pipeline_runs_by_pipeline_id(db: Session, pipeline_id: int)`.

In addition, IMO we should not return the`fastapi_pagination.ext.sqlalchemy.paginate(...)` type from within our `crud.py` like this. I have a few suggestions how we could do it:
1. We could keep the pagination functions in our `curd.py` and rename the function name to `get_pipeline_runs_for_pipeline_id_paginated`
2. We could create a separate `pagination.py` where we then put this function in without any name changes.
3. Do not use the sqlalchemy extension but rather put the resulting list/sequence from the select in the "normal" pagination function

backend/capellacollab/projects/toolmodels/backups/runs/routes.py
```diff
@@ -184,3 +185,6 @@ def get_logs(
         )
 
     return helper.filter_logs(logs, masked_values + masked_values_generated)
+
+
+fastapi_pagination.add_pagination(router)
```
dominik003, line 190: Is it required to have this here at the bottom? Otherwise, I would move it to the router definition at the top.

dominik003 (commented):
LGTM only two small comments

dominik003 (approved):

backend/capellacollab/projects/toolmodels/backups/runs/crud.py
MoritzWeber0: > First, I would not provide the whole pipeline here but rather just the `pipeline_id`. So then having, `def get_pipeline_runs_by_pipeline_id(db: Session, pipeline_id: int)`.

https://github.com/DSD-DBS/capella-collab-manager/pull/717/files#diff-7702246c9b82e90105dd8e55cd049dcafd3f6db16587ecf3909f19c85c5a00a7

> In addition, IMO we should not return the`fastapi_pagination.ext.sqlalchemy.paginate(...)` type from within our `crud.py` like this. I have a few suggestions how we could do it:
> 
> 1. We could keep the pagination functions in our `curd.py` and rename the function name to `get_pipeline_runs_for_pipeline_id_paginated`

This sounds like the best solution for me. 

> 2. We could create a separate `pagination.py` where we then put this function in without any name changes.

As we have only one paginated function, this doesn't really make sense yet. And it belongs to the crud module, I'd say.


> 3. Do not use the sqlalchemy extension but rather put the resulting list/sequence from the select in the "normal" pagination function

This is very inefficient, esp. with a growing number of pipeline runs.

backend/capellacollab/projects/toolmodels/backups/runs/routes.py
MoritzWeber0: I tried it out, but unfortunately, it doesn't work. The error is: 

```py
fastapi_pagination/api.py", line 64, in resolve_params
    raise RuntimeError("Use params or add_pagination") from None
RuntimeError: Use params or add_pagination
```

So it doesn't recognise the pagination registration anymore.

MoritzWeber0:
Because of major restructurings in the pipelines base branch, I had to rework many parts of this PR. 

The move to BehaviourSubjects produced some difficulties with pagination, which are solved. Can please review it again @dominik003?

backend/capellacollab/projects/toolmodels/backups/runs/crud.py
dominik003: Then let's use the first version and simply rename the function. It might be useful to include this in our development documentation, probably in the style guide, so that we use this scheme consistently when we use pagination.

frontend/src/app/projects/models/backup-settings/job-run-overview/job-run-overview.component.ts
```diff
@@ -30,26 +43,91 @@ export class JobRunOverviewComponent implements OnInit {
...
+            projectSlug,
+            modelSlug,
+            pipelineID,
+            1,
+            this.pageSize
           )
         )
       )
       .subscribe();
   }
```
dominik003, line 72: As far as I can see this can be simplified a lot (probably I am missing something about the pagination). Here would be my simplified version:
```typescript
combineLatest([
      this.projectService.project.pipe(filter(Boolean)),
      this.modelService.model.pipe(filter(Boolean)),
      this.pipelineService.pipeline.pipe(filter(Boolean)),
    ])
      .pipe(untilDestroyed(this))
      .subscribe(([project, model, pipeline]) =>
        this.pipelineRunService.loadPipelineRuns(
          project.slug,
          model.slug,
          pipeline.id,
          1,
          this.pageSize
        )
      );
```

Changes I applied here:
- Do not use `map` here as IMO the mapping here makes this way harder to read
- Move the `switchMap` part into the `subscribe` as nothin is done with the result of `loadPipelineRuns` (and also `loadPipelineRuns` should not return anything -> having `void` as return type)

frontend/src/app/projects/models/backup-settings/job-run-overview/job-run-overview.component.ts
```diff
@@ -18,7 +25,13 @@ import { ProjectService } from 'src/app/projects/service/project.service';
...
 @UntilDestroy()
-export class JobRunOverviewComponent implements OnInit {
+export class JobRunOverviewComponent implements OnInit, AfterViewInit {
+  pageSize = 25;
+  pageSizeArray = [...Array(this.pageSize).keys()];
+
+  @ViewChildren('page', { read: ElementRef })
+  pageElements?: QueryList<ElementRef>;
+
   constructor(
```
dominik003, line 35: What is your opinion on the order of service in the constructor? I always tried to have the public services first followed by the private ones. By following this, we can easily and directly identify public and private services.

frontend/src/app/projects/models/backup-settings/job-run-overview/job-run-overview.component.ts
```diff
@@ -30,26 +43,91 @@ export class JobRunOverviewComponent implements OnInit {
...
+            pipelineID,
+            1,
+            this.pageSize
           )
         )
       )
       .subscribe();
   }
 
+  ngAfterViewInit(): void {
```
dominik003, line 74: Similar to the `ngOnInit` I would change this to the following to increase readability. In addition, there was an `untilDestroy(this)` missing.

```typescript
combineLatest([
      this.projectService.project.pipe(filter(Boolean)),
      this.modelService.model.pipe(filter(Boolean)),
      this.pipelineService.pipeline.pipe(filter(Boolean)),
    ])
      .pipe(untilDestroyed(this))
      .subscribe(([project, model, pipeline]) => {
        this.observeVisibleJobs(project.slug, model.slug, pipeline.id);
      });
```

frontend/src/app/projects/models/backup-settings/pipeline-runs/service/pipeline-run.service.ts
```diff
@@ -25,10 +26,21 @@ export class PipelineRunService {
   }
 
   _pipelineRun = new BehaviorSubject<PipelineRun | undefined>(undefined);
-  pipelineRun = this._pipelineRun.asObservable();
-
-  _pipelineRuns = new BehaviorSubject<PipelineRun[] | undefined>(undefined);
-  pipelineRuns = this._pipelineRuns.asObservable();
+  public readonly pipelineRun$ = this._pipelineRun.asObservable();
+
+  _pipelineRunPages = new BehaviorSubject<PageWrapper<PipelineRun>>({
```
dominik003, line 31: I would add `private` here for both behavior subjects

frontend/src/app/projects/models/backup-settings/pipeline-runs/service/pipeline-run.service.ts
```diff
@@ -92,17 +113,58 @@ export class PipelineRunService {
   loadPipelineRuns(
     projectSlug: string,
     modelSlug: string,
-    pipelineID: number
-  ): Observable<PipelineRun[]> {
+    pipelineID: number,
+    page: number,
+    size: number
+  ): Observable<Page<PipelineRun> | undefined | 'loading'> {
```
dominik003, line 119: Both `load...` functions here should have `void` as return type and instead of using `tap` you can then directly put this into the `subscribe` block. Please check the usages of both functions and removed the `subscribe()` from them

frontend/src/app/projects/models/backup-settings/pipeline-runs/service/pipeline-run.service.ts
```diff
@@ -92,17 +113,58 @@ export class PipelineRunService {
...
+    if (pageWrapper.total !== undefined) {
+      // Do nothing, is already initialized
+      return;
+    }
+
+    const firstPage = pageWrapper.pages[0];
+    if (firstPage === undefined || firstPage === 'loading') {
+      throw new TypeError(
+        'first page is undefined, but should be of type Page<PipelineRun>.'
+      );
```
dominik003, line 157: This error message is not directly matching the condition here. I would rather state `first page is undefined or loading, ...` or separate the cases here and throw unique errors for them.

frontend/src/app/projects/models/backup-settings/pipeline-runs/service/pipeline-run.service.ts
```diff
@@ -92,17 +113,58 @@ export class PipelineRunService {
...
+    }
+
+    pageWrapper.total = firstPage.total;
+    for (const page of [...Array(pageWrapper.total).keys()]) {
+      if (pageWrapper.pages[page] === undefined) {
+        pageWrapper.pages[page] = undefined;
+      }
+    }
```
dominik003, line 165: What is the purpose of this loop, so why do we need it? Because right now this sets `pageWrapper.pages[page]` to undefined only if it is already undefined.

dominik003 (changes requested):

frontend/src/app/projects/models/backup-settings/job-run-overview/job-run-overview.component.ts
MoritzWeber0: Sounds good.

Issue #572, opened 2023-01-31
Database columns are mostly nullable


dominik003:
Currently, most of our database columns do not have `nullable=False` set, although we normally assume that they are not null. Therefore, we should adjust our database classes to take this into account. However, since these changes require a database migration, we should postpone this until we have decided whether or not to switch to another database.

PR #719, opened 2023-05-23
Migrate to sql alchemy 2.0

### LINKED ISSUE #307: Migrate SQLAlchemy queries to 2.0 style
More information here: https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-orm-usage

dominik003:
## Summary

This PR comprises major adjustments to our codebase to ensure compatibility with SQLAlchemy 2.0. This compatibility is achieved through the refactoring of our migration scripts, SQLAlchemy models, CRUD operations, and tests. Additionally, we've improved code readability and simplicity, while also ensuring our SQLAlchemy dependency is at version 2.0 or higher.

Resolves #307 #572

### Migration Scripts Adjustments

- **Removal of downgrade scripts**: We've removed downgrade scripts from the existing migration scripts, the Alembic template, and the Alembic tests' downgrade stage.
- **SQLAlchemy 2.0 adjustments**: Made necessary adjustments for SQLAlchemy 2.0 compatibility. This includes wrapping raw SQL statements inside `sa.text(...)`, removing the legacy `db.query(...)` calls, and adjusting all of our models.
- **Non-nullable fields**: A SQLAlchemy 2.0 migration script has been provided, which ensures that many of our fields are non-nullable and addresses issues that arose due to these changes.

### Model Adjustments

- **Introduction of `Mapped[<type>]`**: This change allows for the automatic detection of database column types.
- **Evaluation of field nullability**: Based on the new typing provided by SQLAlchemy, each field's nullability status was evaluated and modified accordingly.
- **Preparation for future enhancements**: In a follow-up PR, we could enhance our database classes by turning them into [dataclasses](https://docs.sqlalchemy.org/en/20/orm/dataclasses.html).

### CRUD Operations Adjustments

- **Transition to `db.execute(...)`**: We've moved from the legacy `db.query(...)` method to the new `db.execute(...)` standard offered by SQLAlchemy 2.0.
- **Consistent CRUD functions**: We have improved the consistency of our CRUD functions. Specifically, the `db: Session` object is now uniformly the first argument across all functions. Additionally, all 'get' CRUD functions will return the requested object or `None`. If an object is not found and an error should be raised, it is advised to use a `_or_raise` CRUD function that raises an SQL exception or handle the situation outside of the CRUD file, possibly by throwing an HTTP Exception.

### Test Adjustments

- **SQLAlchemy 2.0 compatibility**: We have made necessary changes to our tests to ensure they are compatible with SQLAlchemy 2.0 and especially our adjustments to the nullability of fields

### Refactoring and Improvements

- **Code enhancements**: The codebase has been refined for readability and simplicity, which involved various modifications and adjustments, primarily to our models and CRUD functions.

### Dependency Update

- **SQLAlchemy version**: SQLAlchemy version has been bumped to 2.0 or higher.


MoritzWeber0:
> Solves #307 #572

Solves is not a valid keyword, use I've edited the description to Resolved :) 
https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
```diff
@@ -193,11 +165,88 @@ def stop_t4c_repository(
...
+
+    db_repos_dict = {repo.name: repo for repo in db_repos}
+    server_repos_dict = {repo["name"]: repo for repo in server_repos}
+
+    db_repos_names = set(db_repos_dict.keys())
+    server_repos_names = set(server_repos_dict.keys())
+
+    exist_db_and_server_repo_names = server_repos_names & db_repos_names
+    exist_db_not_server_repo_names = db_repos_names - server_repos_names
+    exist_server_not_db_repo_names = server_repos_names - db_repos_names
```
MoritzWeber0, line 216: This is a lot harder to read than just using the value directly in the loop. 

`server_repos_names - db_repos_names` is pretty intuitive when you know the set operations in Python, which I expect as basic knowledge. Reading the variable name takes a few seconds to understand what it means in my brain.

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
```diff
@@ -193,11 +165,88 @@ def stop_t4c_repository(
...
+    server_repos : dict
+        The dictionary of repository data from the server, where the keys are repository names and the values are repository data.
+    instance : DatabaseT4CInstance
+        The database instance associated with the repositories.
+
+    Returns
+    -------
+    list[T4CRepository]
+        The sorted list of updated database repository objects, sorted by repository id.
+    """
```
MoritzWeber0, line 206: Nice docstring! :)

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
MoritzWeber0: And if it's not clear, you can add a short comment.

backend/capellacollab/core/database/migration.py
```diff
@@ -162,15 +158,14 @@ def create_t4c_instance_and_repositories(db):
...
     for t4c_model in t4c_crud.get_t4c_models(db):
-        repository = settings_repositories_models.CreateT4CRepository(
-            name=t4c_model.name,
-        )
         t4c_repository = repositories_crud.create_t4c_repository(
-            repository, default_instance, db
+            db=db, repo_name=t4c_model.name, instance=default_instance
         )
-        t4c_crud.set_repository_for_t4c_model(db, t4c_model, t4c_repository)
+        t4c_model.repository = t4c_repository
```
MoritzWeber0, line 166:
```
        t4c_model.repository = repositories_crud.create_t4c_repository(
            db=db, repo_name=t4c_model.name, instance=default_instance
        )
```

You can combine both lines.

backend/capellacollab/projects/toolmodels/backups/injectables.py
```diff
@@ -20,12 +19,11 @@ def get_existing_pipeline(
...
-    except NoResultFound:
+    if not (backup := crud.get_pipeline_by_id(db, pipeline_id)):
         raise HTTPException(
-            404,
+            status.HTTP_404_NOT_FOUND,
             {
                 "reason": f"The pipeline with the id {pipeline_id} of the model with th id {model.id} was not found.",
             },
         )
+    return backup
```
MoritzWeber0, line 29:
```py
    if backup := crud.get_pipeline_by_id(db, pipeline_id):
        return backup
    
    raise HTTPException(
        status.HTTP_404_NOT_FOUND,
        {
            "reason": f"The pipeline with the id {pipeline_id} of the model with th id {model.id} was not found.",
        },
    )

```

Turn it around.

MoritzWeber0 (changes requested):
Thanks for the nice improvement and that we can finally move to SQLAlchemy 2.0. 
My main concern is the import-style, which has been used for the new `sqlalchemy` imports, but you'll find a more detailed comment in the conversations. Otherwise, it looks really good.

backend/capellacollab/notices/crud.py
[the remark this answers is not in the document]
MoritzWeber0: I've applied some improvements in a new PR, feel free to take it over: https://github.com/DSD-DBS/capella-collab-manager/pull/732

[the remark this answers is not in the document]
dominik003: Created #732, which fixes many of our import issues. Due to the amount of imports, it focuses on consistent adjustment of the most common imports

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
dominik003: IMO, we should never have to create comments for things like this, as you can do the same through a well-described function or a variable name. Also, these three lines here are quite important to the method, and I would prefer to keep them as separate variables. However, I agree that the current names are a bit confusing. My main goal was to use a consistent naming scheme, but I think it's better to make the names more precise and shorter.

So my suggestions here would be to first move each of these lines directly above the for loop where they were used instead of having them in three consecutive lines. In addition I would suggest the following renaming:
- `exist_db_and_server_repo_names` -> `common_repo_names_in_db_and_server `
- `exist_db_not_server_repo_names` -> `unique_repo_names_in_db `
- `exist_server_not_db_repo_names` -> `unique_repo_names_in_server `

What do you think about this?

backend/capellacollab/projects/toolmodels/backups/injectables.py
dominik003: Nice!

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
MoritzWeber0: I agree, we should remove the `exist_` prefix.

MoritzWeber0 (approved):

PR #758, opened 2023-06-13
Use Angular Material button for "Open issue on Github"

MoritzWeber0:
We don't want fetch external JavaScript from Github for security reasons


amolenaar:
Nice cleanup.

PR #764, opened 2023-06-14
Proper loading animations for project detail page

MoritzWeber0:
Loading of projects feels really slow in production and some calls in the frontend were done in wrappers instead of services.

The new loading animation: 

https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/499630cd-ad15-4ec5-b500-4dbbd2182a56

### DIFF
--- a/frontend/src/app/projects/project-detail/project-users/add-user-to-project/add-user-to-project.component.ts
+++ b/frontend/src/app/projects/project-detail/project-users/add-user-to-project/add-user-to-project.component.ts
@@ -4,7 +4,6 @@
  */
 
 import { Component } from '@angular/core';
-import { FormGroupDirective } from '@angular/forms';
 import {
   AbstractControl,
   AsyncValidatorFn,
@@ -85,15 +84,13 @@ export class AddUserToProjectComponent {
     };
   }
 
-  addUser(formDirective: FormGroupDirective): void {
+  addUser(): void {
     this.projectService.project.pipe(filter(Boolean)).subscribe((project) => {
       if (this.addUserToProjectForm.valid) {
         const formValue = this.addUserToProjectForm.value;
 
-        let permission = formValue.permission;
-        if (formValue.role == 'manager') {
-          permission = 'write';
-        }
+        const permission =
+          formValue.role === 'manager' ? 'write' : formValue.permission;
         this.projectUserService
           .addUserToProject(
             project.slug,
@@ -103,9 +100,7 @@ export class AddUserToProjectComponent {
             formValue.reason as string
           )
           .subscribe(() => {
-            formDirective.resetForm();
             this.addUserToProjectForm.reset();
-            this.projectUserService.loadProjectUsers();
             this.toastService.showSuccess(
               `User added`,
               `User '${formValue.username}' has been added to project '${project?.name}'`
--- a/frontend/src/app/projects/project-detail/project-users/project-user-settings.component.ts
+++ b/frontend/src/app/projects/project-detail/project-users/project-user-settings.component.ts
@@ -3,7 +3,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { Component, OnChanges, OnInit, SimpleChanges } from '@angular/core';
+import { Component, OnInit } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { filter } from 'rxjs';
@@ -23,7 +23,7 @@ import { Project, ProjectService } from '../../service/project.service';
   templateUrl: './project-user-settings.component.html',
   styleUrls: ['./project-user-settings.component.css'],
 })
-export class ProjectUserSettingsComponent implements OnInit, OnChanges {
+export class ProjectUserSettingsComponent implements OnInit {
   public project?: Project;
 
   search = '';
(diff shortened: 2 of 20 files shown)


frontend/src/app/projects/project-detail/project-users/add-user-to-project/add-user-to-project.component.ts
```diff
@@ -0,0 +1,118 @@
...
+
+  addUser(formDirective: FormGroupDirective): void {
+    this.projectService.project.pipe(filter(Boolean)).subscribe((project) => {
+      if (this.addUserToProjectForm.valid) {
+        const formValue = this.addUserToProjectForm.value;
+
+        let permission = formValue.permission;
+        if (formValue.role == 'manager') {
+          permission = 'write';
+        }
```
dominik003, line 96:
```suggestion
        const permission = formValue.role == 'manager' ? 'write' : formValue.permission;
```

amolenaar:
Looks good. The implementations seems quite simple. We may want a skeleton loader for the user manager at some point.

amolenaar (approved):

frontend/src/app/projects/project-detail/project-users/add-user-to-project/add-user-to-project.component.ts
```diff
@@ -0,0 +1,118 @@
...
+            project.slug,
+            formValue.username as string,
+            formValue.role as SimpleProjectUserRole,
+            permission as string,
+            formValue.reason as string
+          )
+          .subscribe(() => {
+            formDirective.resetForm();
+            this.addUserToProjectForm.reset();
+            this.projectUserService.loadProjectUsers(project);
```
dominik003, line 108: IMO this should not be part of this method here. I would add a `.pipe(tap(() => this.loadProjectUsers(project_slug)))` to the `addUserToProject` function. Unfortunately, the `loadProjectUsers` function currently expects a whole `Project` but as it just needs the slug, I would adjust the function to simply require the project slug.

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.ts
```diff
@@ -170,7 +53,7 @@ export class ProjectUserSettingsComponent implements OnInit, OnChanges {
     this.projectUserService
       .deleteUserFromProject(this.project.slug, user.id, reason)
       .subscribe(() => {
-        this.refreshProjectUsers();
+        this.projectUserService.loadProjectUsers(this.project!);
```
dominik003, line 56: Similar to the `addUser` function, I would also move the `loadProjectUsers` to the `changeRoleOfProjectUser` function. As I can see we have several `this.projectUserService.loadProjectsUsers(...)` in this component which I would all move to the projectsUsersService

dominik003 (changes requested):
The changes look good, nice improvements for loading the project. However, I have some improvements for the code you probably just copied (it's still good to improve it even if you didn't actually write it here), and I have some doubts about moving too much logic to the user service instead of creating a proper wrapper component for it.

dominik003:
I just saw that there is already a review on this PR. I just looked in my open PR list and started the review, so now we have a total of two reviews here 😄 It might be a good idea to unassign reviewers when they are no longer needed.

frontend/src/app/projects/project-detail/project-users/add-user-to-project/add-user-to-project.component.ts
```diff
@@ -0,0 +1,118 @@
...
+      ) {
+        permission?.setErrors(null);
+        return null;
+      }
+      permission?.setErrors({ permissionInvalid: true });
+      return { permissionInvalid: true };
+    };
+  }
+
+  addUser(formDirective: FormGroupDirective): void {
```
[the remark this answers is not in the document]
MoritzWeber0: Is removed.

frontend/src/app/projects/project-detail/project-users/add-user-to-project/add-user-to-project.component.ts
MoritzWeber0: Done.

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.ts
MoritzWeber0: Done

frontend/src/app/projects/project-detail/project-users/service/project-user.service.ts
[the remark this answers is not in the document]
MoritzWeber0: Each wrapper needs a custom route, which we don't have for project users. Project users are loaded when you call the `/project/<slug>` route, so the wrapper would not be called. 

I don't want to have it the ProjectWrapper (the project wrapper should only care about projects and not all subcomponents, this is against the principle of context-scoped components/services).

Alternatively, we can move the logic to all related project user components, but this would involve a lot of code duplication.

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.ts
```diff
@@ -243,11 +126,25 @@ export class ProjectUserSettingsComponent implements OnInit, OnChanges {
...
 
-  getProjectUsersByRole(role: SimpleProjectUserRole): ProjectUser[] {
-    return this.projectUsers.filter(
+  getProjectUsersByRole(
+    projectUsers: ProjectUser[] | undefined | null,
+    role: string
+  ): ProjectUser[] | undefined {
+    if (projectUsers === undefined || projectUsers === null) {
+      return undefined;
+    }
```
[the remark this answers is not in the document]
MoritzWeber0: `async` can produce null (I don't know in which case, but thecompiler complains about it).

dominik003 (approved):

PR #767, opened 2023-06-15
Add Tailwind CSS

MoritzWeber0:
Tailwind CSS is great for fast prototyping. 99% of the CSS is already covered. In addition, we have many unused CSS classes which have not been deleted when the HTML disappeared, Tailwind solves the issue.


amolenaar (approved):
Looks good. Tailwind makes for some nice simplifications.

PR #766, opened 2023-06-15
Add responsive design / support for mobile phones

MoritzWeber0:
Ever wanted to run Capella on your smartphone? Then the PR is made for you. 

This PR adds some fundamentals for responsive design. It adds a side-nav-bar and removes many fixed-width components to support small screens.

Not all components are responsive yet, but the most important ones are already supported.

Preview of the Capella Collaboration Manager running on smartphone:

https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/0c2d4a67-ae59-4ba1-8016-ccb509009243

This is how Capella looks on a small screen: 
image

Resolves #765

PR #776, opened 2023-06-20
Make all components responsive

MoritzWeber0:
The important components were made responsive in https://github.com/DSD-DBS/capella-collab-manager/pull/766

However, it was focused on the components that the user can see. This PR also makes the administrator components responsive.

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -58,7 +58,7 @@ const routes: Routes = [
           },
           {
             path: 'create',
-            data: { breadcrumb: 'New Project' },
+            data: { breadcrumb: 'create' },
             component: CreateProjectComponent,
           },
         ],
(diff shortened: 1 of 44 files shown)


frontend/src/app/projects/models/model-description/model-description.component.ts
```diff
@@ -121,4 +123,8 @@ export class ModelDescriptionComponent implements OnInit {
       },
     });
   }
+
+  navigateBack() {
+    this.location.back();
+  }
```
dominik003, line 129: I would use something like `this.router.navigate(['../../..'], { relativeTo: this.route })` here as `this.location.back()` can have some side effects, so this one is more robust

frontend/src/app/settings/modelsources/git-settings/edit-git-settings/edit-git-settings.component.ts
```diff
@@ -77,4 +79,8 @@ export class EditGitSettingsComponent implements OnInit, OnDestroy {
         this.router.navigate(['../..'], { relativeTo: this.route })
       );
   }
+
+  navigateBack() {
+    this.location.back();
+  }
```
dominik003, line 85: Same as above: I would use `this.router.navigate(['../..'], { relativeTo: this.route })` instead

dominik003 (approved):

PR #782, opened 2023-06-21
Update development instructions

Paula-Kli:
This PR updates a few instructions for setting up the development environment.

Biggest change: 

- the option of setting up the OAuth Mock with a self signed certificate is removed

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,7 +18,7 @@ your code follows our style guidelines outlined below.
 ## General
 
 This project consists of several services. Here is the architecture of the services:
-![Capella Collab Manager architecture](doc/architecture.png)
+![Capella Collab Manager architecture](docs/architecture.png)
 
 To get an overview of the services, it is also worth taking a look at the Helm Chart,
 which can be found in the `helm` folder.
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -70,12 +70,8 @@ first.
...
 First of all, navigate to the `mocks/oauth` directory.
 
-You can handle the SSL-support in two different ways:
-
-### Option 1: Use insecure connections
-
-This is the default and recommended option for local development.
-Do not use it in production!
+This is the recommended option for local development.
+Do not use insecure connections it in production!
```
MoritzWeber0, line 74:
```suggestion
```

I'd remove the two lines. Not only insecure connections, but the OAuth mock in general should not be used in production.

"This is the recommended option for local development." doesn't make much sense anymore as we have only one option left.

Instead, I would add `Do not use the OAuth mock in production!` after `For local development, we have an OAuth mock that needs to be started first.`

MoritzWeber0 (commented):
Good first PR, thanks! I have only one small comment.

MoritzWeber0 (approved):

PR #791, opened 2023-06-23
refactor: Use `$` as suffix for observable variables

MoritzWeber0:
This PR contains some consistency improvements. 

- All variables containing Observables are suffixed with `$`. This convention was introduced by Cycle.js, but is widely used: https://cycle.js.org/basic-examples.html#basic-examples-increment-a-counter-what-is-the-convention 
- For private BehaviourSubjects in services, the `private` annotation is added.
- For public Observables in services, the `public` and `readonly` annotations are added.


amolenaar (approved):
LGTM

PR #793, opened 2023-06-26
Use new-style t4c/client backup functionality

amolenaar:
This allows jobs to work with the new consolidated t4c/client image.
To improve maintainability I changed calls to `_create_job_spec` to
named arguments.
    
I decided to provide the command to the k8s api, instead of hard-coding
it there, to keep the symantics of the methods used --
`create_(cron)job()` -- agnostic.
    
The changes should be compatible with older t4c backup images, since
they just ignore the command (args).

This PR bulds on https://github.com/DSD-DBS/capella-dockerimages/pull/162

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ capella:
 	$(CAPELLA_DOCKERIMAGES) CAPELLA_VERSIONS="$(CAPELLA_VERSIONS)" capella/remote capella/readonly
 
 t4c-client:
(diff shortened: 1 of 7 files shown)


dominik003:
There are at least two places where we have to adjust things:
1. `migration.py`: Adjust the default backup image
2. `Makefile`: Exchange `t4c/client/backup` by `t4c/client/base`

amolenaar:
`migration.py` and `Makefile` have been updates as well.

PR #803, opened 2023-06-27
feat: add inactive alert for jupyter notebooks

amolenaar:
So jupyter notebooks are automatically shut down after being idle for a while.

### DIFF
--- a/backend/capellacollab/sessions/sessions.py
+++ b/backend/capellacollab/sessions/sessions.py
@@ -65,7 +65,7 @@ def _get_last_seen(idletime: int | float) -> str:
     if (idlehours := idletime / 60) > 1:
         return f"{round(idlehours, 2)} hrs ago"
 
-    return f"{idletime} mins ago"
+    return f"{idletime:0f} mins ago"
 
 
 def _determine_session_state(session: DatabaseSession) -> str:
--- a/backend/capellacollab/sessions/sessions.py
+++ b/backend/capellacollab/sessions/sessions.py
@@ -65,7 +65,7 @@ def _get_last_seen(idletime: int | float) -> str:
     if (idlehours := idletime / 60) > 1:
         return f"{round(idlehours, 2)} hrs ago"
 
-    return f"{idletime:0f} mins ago"
+    return f"{idletime:.0f} mins ago"
 
 
 def _determine_session_state(session: DatabaseSession) -> str:
--- a/helm/templates/prometheus/prometheus.configmap.yaml
+++ b/helm/templates/prometheus/prometheus.configmap.yaml
@@ -13,6 +13,8 @@ data:
     groups:
       - name: Inactivity
         rules:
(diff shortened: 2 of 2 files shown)


MoritzWeber0:
image

The last seen column still shows "UNKNOWN" as last state. It would be good to indicate the idletime for Jupyter sessions too.

amolenaar:
> The last seen column still shows "UNKNOWN" as last state. It would be good to indicate the idletime for Jupyter sessions too. Otherwise it looks good and my session was also killed in the local cluster.

I've been looking into this, but I can not see a way to determine how long the idle condition holds true.

I propose we create a new story for this, since it requires some advanced Prometheus magic.

amolenaar:
Never mind:

```
time() - max_over_time(timestamp(absent(sum by (app) (increase(http_request_duration_seconds_count[2m])) < 4.5))[60m:5m])
```

amolenaar:
* the timeout is not triggered when the session has not been opened
* Let's have a look at the `1d:5m` time blocks. Could probably be shorter (`90m:2m`).

helm/templates/prometheus/prometheus.configmap.yaml
```diff
@@ -13,12 +13,14 @@ data:
     groups:
       - name: Inactivity
         rules:
-          - alert: Inactivity session
+          - record: idletime_minutes
+            expr: (time() - max_over_time(timestamp(sum by (app) (increase(http_request_duration_seconds_count[2m])) > 4.5)[1d:5m])) / 60
```
MoritzWeber0, line 17: Instead of having a complex expression, which is Jupyter specific, we should make use of the official Jupyter API instead: https://jupyter-server.readthedocs.io/en/latest/developers/rest-api.html

As discussed, it should be checked if it can be integrated into the `metrics` endpoint of the Jupyter container. Alternatively, we can have our own metrics endpoint.

amolenaar: I checked. The `last_activity` metric is about the activity of the Jupyter kernel. Also the status API requires authentication.

Instead I updated the metric to take into account the start time of the JupyterLab instance. Not it nicely counts idle time from the start of the session.

backend/capellacollab/sessions/sessions.py
```diff
@@ -65,7 +65,7 @@ def _get_last_seen(idletime: int | float) -> str:
     if (idlehours := idletime / 60) > 1:
         return f"{round(idlehours, 2)} hrs ago"
 
-    return f"{idletime} mins ago"
+    return f"{idletime:0f} mins ago"
```
MoritzWeber0, line 68:
```suggestion
    return f"{idletime:.0f} mins ago"
```

The current line does nothing.

MoritzWeber0 (changes requested):

MoritzWeber0 (approved):
Tried it out and works on my side. An impressive query you've built. ChatGPT was able to explain it to me luckily :)

PR #811, opened 2023-06-29
Guacamole version update

MoritzWeber0:
The current Guacamole version seems to be unstable. There are two new minor versions since 1.5.2 and we should install them.


MoritzWeber0:
Unfortunately, sessions do not start anymore. It's running into "Session disconnected" immediately.

Issue #781, opened 2023-06-21
Add CI/CD integration for Github


MoritzWeber0:
It would be nice to extend the current Gitlab specific features for Github instances.
Namely, the model badge & diagram cache should be fetched from Github repositories / model artefacts. 

- [x] Create a new Github workflow (CI) to generate the model badge in the [coffee machine](https://github.com/DSD-DBS/coffee-machine) repository. The Gitlab template can be used as template: https://github.com/DSD-DBS/py-capellambse/blob/master/ci-templates/gitlab/model-badge.yml
- [x] Create a new Github workflow (CI) to generate the diagram cache in the [coffee machine](https://github.com/DSD-DBS/coffee-machine) repository. The Gitlab template can be used as template: https://github.com/DSD-DBS/capella-dockerimages/blob/main/ci-templates/gitlab/diagram-cache.yml
- [x] Create a new Python submodule and add a Github interface.py (analogous to Gitlab) here: https://github.com/DSD-DBS/capella-collab-manager/tree/main/backend/capellacollab/projects/toolmodels/modelsources/git
- [x] Update the model-badge & diagram cache routes to add support for fetching from Github
- [x] Update the documentation (docs/user)
- [x] Add release notes in docs/user
- [x] Make general templates out of the CI created in step 1 & 2, which can be added to this repository.

PR #812, opened 2023-06-29
feat: Add support for github API

Paula-Kli:
fixes parts of https://github.com/DSD-DBS/capella-collab-manager/issues/781

This adds the possibility to retrieve the diagram cache and the model badge from github projects

- [ ] Update changelog when approved


backend/capellacollab/projects/toolmodels/modelsources/git/interface.py
```diff
@@ -0,0 +1,144 @@
...
+    git_model: git_models.DatabaseGitModel,
+    git_instance: settings_git_models.DatabaseGitInstance,
+) -> bytes:
+    return get_git_interface(git_instance).get_artifact_from_job_as_content(
+        project_id,
+        job_id,
+        trusted_path_to_artifact,
+        git_model,
+        git_instance,
+    )
```
MoritzWeber0, line 132: I don't really like that we have to implement each function on the interface and have to derive each time, which interface it is.

Instead, I would suggest the following: 
- Create a superclass / interface. In the interface, you can implement all functions which are independent from the specific instance. Public instance-specific methods should be implemented as abstractmethods.
- Use the factory pattern to determine the right subclass: https://en.wikipedia.org/wiki/Factory_method_pattern
- Each instance has specific sub-class. The class inherits from the subclass / interface and implements the abstract functions + optionally private functions.

backend/tests/projects/toolmodels/test_diagrams.py
MoritzWeber0, line 1: Would be great to see some tests for the Github integration. Right now, all tests are based on the Gitlab interface.

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/interface.py
```diff
@@ -7,17 +7,16 @@
...
 
 
 def get_last_job_run_id_for_git_model(
-    db: orm.Session, job_name: str, git_model: git_models.DatabaseGitModel
+    job_name: str,
+    git_model: git_models.DatabaseGitModel,
+    git_instance: settings_git_models.DatabaseGitInstance,
 ) -> tuple[settings_git_models.DatabaseGitInstance, str, tuple[str, str]]:
-    git_instance = get_git_instance_for_git_model(db, git_model)
     check_git_instance_is_gitlab(git_instance)
```
MoritzWeber0, line 20: Shouldn't be the instance always be gitlab at this point?

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
```diff
@@ -0,0 +1,197 @@
...
+
+def get_file_from_repository(
+    trusted_file_path: str,
+    git_model: git_models.DatabaseGitModel,
+    git_instance: settings_git_models.DatabaseGitInstance,
+) -> bytes:
+    project_id = get_project_id_by_git_url(git_model, git_instance)
+    public_response = requests.get(
+        f"{git_instance.api_url}/repos{project_id}/contents/{parse.quote(trusted_file_path, safe='')}?ref={parse.quote(git_model.revision, safe='')}",
+        timeout=2,
```
MoritzWeber0, line 182: You can read the timeout from the configuration. We have it as configurable option.

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
```diff
@@ -0,0 +1,197 @@
...
+                    "The latest artifact you are requesting expired. Please rerun your pipline and contact your administrator."
+                ),
+            },
+        )
+    artifact_response = requests.get(
+        f"{git_instance.api_url}/repos{project_id}/actions/artifacts/{artifact_id}/zip",
+        headers={
+            "Authorization": f"token {git_model.password}",
+            "X-GitHub-Api-Version": "2022-11-28",
+            "Accept": "application/vnd.github+json",
```
MoritzWeber0, line 150: I would prefer to have this as global variable. In a global variable, we could bump the Github version in one place.

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
```diff
@@ -0,0 +1,197 @@
...
+    with zipfile.ZipFile(io.BytesIO(response.content)) as zip_ref:
+        temp_dir = tempfile.mkdtemp()
+        zip_ref.extractall(temp_dir)
+        with open(
+            temp_dir + "/" + trusted_file_path.split("/")[-1],
+            encoding="utf-8",
+        ) as file:
+            content = file.read()
+        shutil.rmtree(temp_dir)
+        return content
```
MoritzWeber0, line 171: Writing to the disk is slow. Instead, you can do something like this: 
```py
with zipfile.ZipFile(io.BytesIO(response.content)), 'r') as zf:
        # Get the list of files in the zip
        file_list = zf.namelist()

        # Create a dictionary to store the unzipped data
        unzipped_data = {}

        # Go through each file in the zip
        for filename in file_list:
            # Read each file's data into memory
            with zf.open(filename, 'r') as file:
                unzipped_data[filename] = file.read()

        return unzipped_data
```

You can also modify the code snipped to just read the requested file.

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
```diff
@@ -0,0 +1,197 @@
...
+    git_instance: settings_git_models.DatabaseGitInstance,
+) -> bytes:
+    project_id = get_project_id_by_git_url(git_model, git_instance)
+    public_response = requests.get(
+        f"{git_instance.api_url}/repos{project_id}/contents/{parse.quote(trusted_file_path, safe='')}?ref={parse.quote(git_model.revision, safe='')}",
+        timeout=2,
+    )
+    if public_response.ok:
+        response = public_response
+    else:
```
MoritzWeber0, line 186: I don't like the approach that we have to send two requests. If there is a password set, we should send the authenticated request first. The chance is higher to get a successful response.

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
```diff
@@ -0,0 +1,197 @@
...
+    trusted_file_path: str,
+    git_model: git_models.DatabaseGitModel,
+    git_instance: settings_git_models.DatabaseGitInstance,
+) -> bytes:
+    project_id = get_project_id_by_git_url(git_model, git_instance)
+    public_response = requests.get(
+        f"{git_instance.api_url}/repos{project_id}/contents/{parse.quote(trusted_file_path, safe='')}?ref={parse.quote(git_model.revision, safe='')}",
+        timeout=2,
+    )
+    if public_response.ok:
```
MoritzWeber0, line 184: We should more explicit here. We just have to retry in case of access issues (Unauthenticated / Not authorised).

MoritzWeber0, line 184: Can you add an explanation why this is needed as docstring? (It's a specific case and we should remember it when we look at the code again in one year.)

MoritzWeber0 (changes requested):

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
Paula-Kli: good point - I will also change it for the gitlab interface as well then? As there are several times where 2 is hardcoded

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
```diff
@@ -0,0 +1,197 @@
...
+import capellacollab.projects.toolmodels.modelsources.git.models as git_models
+import capellacollab.settings.modelsources.git.models as settings_git_models
+
+
+def get_project_id_by_git_url(
+    git_model: git_models.DatabaseGitModel,
+    git_instance: settings_git_models.DatabaseGitInstance,
+) -> str:
+    # Project ID contains /{owner}/{repo_name}
+    del git_instance  # unused
```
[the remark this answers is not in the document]
Paula-Kli: Shall I then write something that the interface should be the same? otherwise it might be at first sight a bit weird that there is a variable which first thing in the method is deleted?

docs/user/docs/settings/model-sources/git.md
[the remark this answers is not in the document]
Paula-Kli: 👍 I will add a description on how to set up the model complexity badge and diagram cache pipeline later when it actually works - I think that makes more sense

backend/capellacollab/projects/toolmodels/modelsources/git/interface.py
Paula-Kli: how would you name the class then? when referencing it in the routes.py files I named it `interface_class` but I don't think this is really a good option - do you have a good idea?

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
Paula-Kli: I also changed here the access to authenticated if a password is set in contrast to first trying unauthenticated and later try authenticated again

backend/capellacollab/projects/toolmodels/modelsources/git/interface.py
dominik003, line 1: I will use this comment here to describe a more clear structuring of the abstracted git features. 

First of all, we should change the naming from the `GitInterface` to `GitHandler`, `GithubInterface` to `GithubHandler`, and `GitlabInterface` to `GitlabHandler`. The reason for this is that using the term `Interface` in the context of classes might lead to some misunderstandings as an interface usually just describe which functions must be implemented from a class using this interface.

Then I would create a new subdirectory here with the name `handler` with the following four files `__init__.py`, `factory.py`, `injectables.py`, and `models.py`. The content of the `injectables.py` is explained above in the routes, the `factory.py` will be explained here, and the `models.py` in the `interface_class.py`.

backend/capellacollab/projects/toolmodels/modelsources/git/interface.py
```diff
@@ -0,0 +1,63 @@
...
+        case _:
+            raise fastapi.HTTPException(
+                status_code=fastapi.status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail={
+                    "err_code": "GIT_INSTANCE_UNSUPPORTED",
+                    "reason": (
+                        "The used Git instance is neither a Gitlab nor a Github instance.",
+                    ),
+                },
+            )
```
dominik003, line 36: I wouldn't put this into a function but rather use a Factory for that and put it into `factory.py` in the `handler` subdirectory. The factory class could look like this (you may need to slightly adjust it)

```
class GitHandlerFactory:
    @staticmethod
    def create_git_handler(
        db: orm.Session, git_model: git_models.DatabaseGitModel
    ) -> models.GitHandler:
        git_instance = get_git_instance_for_git_model(db, git_model)
        if git_instance.type == settings_git_models.GitType.GITLAB:
            return gitlab_models.GitlabHandler(git_model, git_instance)
        elif git_instance.type == settings_git_models.GitType.GITHUB:
            return github_models.GithubHandler(git_model, git_instance)
        else:
            raise fastapi.HTTPException(
                status_code=fastapi.status.HTTP_500_INTERNAL_SERVER_ERROR,
                detail={
                    "err_code": "GIT_INSTANCE_UNSUPPORTED",
                    "reason": (
                        "The used Git instance is neither a Gitlab nor a Github instance.",
                    ),
                },
            )
```

dominik003 (dismissed):

backend/capellacollab/projects/toolmodels/modelsources/git/github/interface.py
```diff
@@ -0,0 +1,173 @@
...
+            f"{self.git_instance.api_url}/repos{project_id}/contents/{parse.quote(trusted_file_path, safe='')}?ref={parse.quote(self.git_model.revision, safe='')}",
+            timeout=config["requests"]["timeout"],
+        )
+
+        if not response.ok and self.git_model.password:
+            response = requests.get(
+                f"{self.git_instance.api_url}/repos{project_id}/contents/{parse.quote(trusted_file_path, safe='')}?ref={parse.quote(self.git_model.revision, safe='')}",
+                headers=self.get_headers(self.git_model.password),
+                timeout=config["requests"]["timeout"],
+            )
```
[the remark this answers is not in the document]
Paula-Kli: We need that in order to be able to also retrieve the model badge of an open source repository even though we might not have a token to download a file.
If there is an open source repository and I do not authenticate I can download the files. If I authenticate and have not the right permissions I can not download the exact same file. Therefor, we are trying to get the file twice here

docs/user/docs/projects/models/diagrams/view_diagram_cache.md
```diff
@@ -13,6 +13,6 @@
    <!-- prettier-ignore -->
     !!! note
         We'll search for a successful job run in the last 20 pipelines of
-        the specific revision. If there is no successful pipeline, you won't be able
+        the specific revision. If there is no successful pipeline or on Github your artifacts have expired, you won't be able
```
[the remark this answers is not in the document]
Paula-Kli: not entirely sure what you wanted to suggest here - sorry
tried to fix it in a more understandable way

backend/capellacollab/projects/toolmodels/modelsources/git/interface.py
Paula-Kli: What do you mean, that the models.py will be described in the interface_class.py?

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/validation.py
```diff
@@ -27,7 +27,7 @@ async def check_pipeline_health(
 
     try:
         await gitlab_interface.get_last_job_run_id_for_git_model(
```
MoritzWeber0, line 29: The `gitlab_interface` has no member `get_last_job_run_id_for_git_model` anymore. Can you also change the method to run against all Git integrations?

MoritzWeber0 (commented):
Looks pretty good already, thanks for the good work @Paula-Kli :) 

There is just one comment left and I opened a PR with some smaller changes: https://github.com/DSD-DBS/capella-collab-manager/pull/872

MoritzWeber0 (approved):

PR #816, opened 2023-07-03
Fix pipeline job

amolenaar:
Due to an incomplete change, pipeline jobs cannot be created in the latest version of Collab manager.

This PR fixes the regression.


MoritzWeber0 (approved):
Thanks for the fast fix :)

PR #826, opened 2023-07-04
feat: Add caching for model monitoring

MoritzWeber0:
Loading the model monitoring overview takes a while due to many requests to the Gitlab API. This PR adds caching.

image


MoritzWeber0:
I noticed that the UX with caching is pretty bad either. 
Async calls might be the better solution: https://github.com/DSD-DBS/capella-collab-manager/pull/830

PR #830, opened 2023-07-05
feat: Make loading of monitoring overview async

MoritzWeber0:
Loading the monitoring overview can take minutes when having many models.
We're waiting a lot for responses from the Gitlab API. This PR adds support for asynchronous calls in the Gitlab interface.

Tested it against a production database of our largest instance. Loading time of the monitoring view is reduced from 2.5 minutes to 8 seconds.

### DIFF
--- a/backend/capellacollab/health/routes.py
+++ b/backend/capellacollab/health/routes.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 
+import asyncio
 import logging
 
 import fastapi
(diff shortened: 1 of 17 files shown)


backend/capellacollab/health/routes.py
```diff
@@ -46,31 +47,49 @@ def general_status(db: orm.Session = fastapi.Depends(database.get_db)):
...
             pipeline_status=pipelines_validation.check_last_pipeline_run_status(
                 db, model
             ),
-            model_badge_status=modelbadge_validation.check_model_badge_health(
-                db, model
-            ),
-            diagram_cache_status=diagrams_validation.check_diagram_cache_health(
-                db, model
-            ),
+            **results_per_model[model.id],
```
amolenaar, line 90: It may be more readable if the tasks are awaited here, instead of doing a wildcard merge.

```suggestion
            primary_git_repository_status: await tasks_per_model[model.id]["primary_git_repository_status"],
            diagram_cache_status: await tasks_per_model[model.id]["diagram_cache_status"],
            model_badge_status: await tasks_per_model[model.id]["model_badge_status"]         
```

Now it's more clear what fields are exposed from this function.

backend/capellacollab/health/routes.py
```diff
@@ -46,31 +47,49 @@ def general_status(db: orm.Session = fastapi.Depends(database.get_db)):
...
+            ),
+        }
+        for model in toolmodels
+    }
+
+    results_per_model = {}
+    for model_id, models_tasks in tasks_per_model.items():
+        results_per_model[model_id] = {}
+        for task_name, task in models_tasks.items():
+            results_per_model[model_id][task_name] = await task
```
amolenaar, line 80: This bit is not needed if you follow the next suggestion.

amolenaar (approved):
Looks good. What is not clear for me from the code is if one HTTP request fails the rest of the calls are still handled. Or is it returning as soon as a http request fails?

Maybe a few test-cases for the non-happy scenario's?

backend/capellacollab/health/routes.py
MoritzWeber0: Nice cleanup.

MoritzWeber0:
> Looks good. What is not clear for me from the code is if one HTTP request fails the rest of the calls are still handled. Or is it returning as soon as a http request fails?

If one HTTP request fails, the other requests of the specific model are not handled anymore. 

> Maybe a few test-cases for the non-happy scenario's?

Most of the cases are handled in the `test_diagrams.py` and `test_model_badge.py`. I'll add tests for the monitoring too.

PR #832, opened 2023-07-10
Fix: Set fastapi version to `<0.100.0`

dominik003:
In fastapi version 0.100.0 pydantic v2 where introduced. Even though there shouldn't be that many breaking changes, when running `make deploy` locally I get the following error: `AttributeError: module 'pydantic._migration' has no attribute 'GenericModel'`. This PR resolves this for the moment by setting the fastapi version to `<0.100.0`.

Migration to pydantic v2 is tracked in #833

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "alembic==1.10.4",
   "appdirs",
   "cachetools",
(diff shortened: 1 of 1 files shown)


amolenaar (approved):
LGTM

MoritzWeber0 (approved):

PR #845, opened 2023-07-12
Remove more hints for self signed authentication mock

Paula-Kli:


MoritzWeber0 (approved):

MoritzWeber0:
We can actually also remove the `certs` directory.

PR #851, opened 2023-07-14
fix: Reset audit log on project change and display user

MoritzWeber0:
The audit log was not cleared when changing the project. It was displaying the results from the old project. 
Also, the affected user was not shown, only the person who produced the event.


amolenaar (approved):
LGTM

:shipit:

PR #882, opened 2023-07-31
feat: Show users interaction possibilities with diagram cache and generate basic auth token

### LINKED ISSUE #598: Show code snippet and generate techusers for diagram cache
When opening the diagram cache in the frontend, users should get a code snippet how to integrate the diagram cache in `capellambse`. In addition, there should be a way to generate personal access tokens. Personal access tokens can be used to authenticate against the diagram cache endpoint.

Paula-Kli:
closes #598 

this is what it looks like currently:
Bildschirmfoto 2023-08-28 um 16 17 51


frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.css
```diff
@@ -26,6 +26,25 @@
...
+  border-radius: 5px;
+  cursor: pointer;
+  box-shadow: none;
+}
+
+.insert-button {
+  align-self: flex-end;
+  margin-top: auto;
+  color: white;
+}
```
dominik003, line 46: A while ago we added [tailwindcss](https://tailwindcss.com/) to the collab manager. Tailwind has the huge advantage that one can directly define styles in the `class` attribute and in a way shorter form. For example, instead of having `position: absolute` in the style attribute or css, one simply needs to use `class = "absolute"`.
In case we have a class that applies to several elements, using a css class is still a good choice, but in these two cases where both are single buttons, I would transform these rules into tailwind classes.

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -58,7 +70,7 @@ export class ModelDiagramDialogComponent {
     private dialogRef: MatDialogRef<ModelDiagramDialogComponent>,
     private dialog: MatDialog,
     @Inject(MAT_DIALOG_DATA)
-    public data: { modelSlug: string; projectSlug: string }
+    public data: { modelSlug: string; projectSlug: string; path: string }
```
dominik003, line 73: Here I would remove the path from the data, as we can easily use the method you currently use to set the path in here. To do so, I would add `path: string` at the top and in the constructor, we can then set it as follows:
`this.path = this.modelService.backendURLFactory(data.projectSlug, data.modelSlug)` (where we also need to add `private modelService: ModelService` to the constructor. In the html, we can then simply use `this.path` instead of `data.path`.

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -73,6 +85,11 @@ export class ModelDiagramDialogComponent {
       });
   }
 
+  async ngOnInit() {
+    const data = await this.getUser();
+    this.cachedUsername = data.name;
+  }
```
dominik003, line 91: I think, we no longer need this here when adjusting the functionality in here and changing the backend route to use `get_own_user` (i.e., we no longer need the id of the current user)

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -147,6 +164,41 @@ export class ModelDiagramDialogComponent {
...
+  async getUser() {
+    const token = localStorage.getItem('access_token');
+
+    const headers = this.getHeaders(token);
+
+    const response = await fetch(environment.backend_url + '/users/current', {
+      headers,
+    });
+    return response.json();
+  }
```
dominik003, line 177: Same as above, we no longer need this here.

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -147,6 +164,41 @@ export class ModelDiagramDialogComponent {
...
+    const basic_token = await response.text();
+    this.passwordValue = basic_token.replaceAll('"', '');
+  }
+
+  getHeaders(token: string | null) {
+    return {
+      Authorization: `Bearer ${token}`,
+      'Content-Type': 'application/json',
+    };
+  }
```
dominik003, line 201: We have an auth interceptor that automatically adds the authorisation header for each outgoing request (in the `injectAccessToken` method). Therefore, we do not have to create the headers by outselves.

backend/capellacollab/users/tokens/routes.py
```diff
@@ -0,0 +1,37 @@
...
+            auth_injectables.RoleVerification(
+                required_role=users_models.Role.USER
+            )
+        )
+    ]
+)
+
+
+@router.get("/{user_id}/token")
+def get_token_for_user(
```
dominik003, line 28: I would make this a `POST` instead of a `GET` as we actually create something and a `GET` should only be used to retrieve data. Since we don't need anything, the body can simply be empty.

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -147,6 +164,41 @@ export class ModelDiagramDialogComponent {
...
+
+    const response = await fetch(
+      environment.backend_url + `/users/${id}/token`,
+      {
+        headers,
+      }
+    );
+    const basic_token = await response.text();
+    this.passwordValue = basic_token.replaceAll('"', '');
+  }
```
dominik003, line 194: Here I would move the actual http call to the `UserService` which is responsible for any user related actions (e.g., creating a user, getting the current user, deleting a user). There, we can add a `createToken` function (or any other fitting name) that could look like this:
```typescript
createToken(): Observable<string> {
    return this.http.post<string>(
      this.BACKEND_URL_PREFIX + `current/token`,
      {}
    );
  }
```

In here, we then just have to put in the following:
```typescript
insertToken() {
  this.userService
      .createToken()
      .subscribe((token) => (this.passwordValue = token.replaceAll('"', '')));
}
```

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -25,15 +27,20 @@ import {
...
 @Component({
   selector: 'app-model-diagram-dialog',
   templateUrl: './model-diagram-dialog.component.html',
   styleUrls: ['./model-diagram-dialog.component.css'],
 })
-export class ModelDiagramDialogComponent {
+export class ModelDiagramDialogComponent implements OnInit {
   diagramMetadata?: DiagramCacheMetadata;
   diagrams: Diagrams = {};
+  cachedUsername: string | null = null;
```
dominik003, line 40: I would simply name this `username` instead of `cachedUsername`. In addition, I would also use `undefined` as default value (we typically use `undefined` for the case where data were not yet loaded but should be there, e.g., when initialising the value via an api call). So it can look like this: `username?: string` (short for `username: string | undefined = undefined`.

The actual initialisation can then be done inside of the constructor as follows:

```typescript
this.userService
      .getCurrentUser()
      .subscribe((user) => (this.username = user.name));
```

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -25,15 +27,20 @@ import {
...
+  passwordValue: string =
+    'You can insert your project token here using the "Insert token" button. This will invalidate any old token for your user';
```
dominik003, line 42: I would make this here possibly undefined (i.e., `passwordValue?: string`) and then use the following in the html:
`{{ passwordValue !== undefined ? passwordValue : 'You can insert your project token here using the "Insert token" button. This will invalidate any old token for your user' }}`. In doing so, we can remove the `isButtonClicked` variable here as we can simply use the state of the `passwordValue` variable.

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
```diff
@@ -42,6 +49,11 @@ export class ModelDiagramDialogComponent {
 
   search = '';
 
+  @ViewChild('codeBlock', { static: false }) codeBlock!: ElementRef;
```
dominik003, line 52:
```suggestion
  @ViewChild('codeBlock') codeBlock!: ElementRef;
```
`false` should be the default value here, so we can omit it

frontend/src/app/projects/project-detail/model-overview/model-overview.component.ts
```diff
@@ -64,7 +64,14 @@ export class ModelOverviewComponent implements OnInit {
...
       width: '80vw',
-      data: { modelSlug: model.slug, projectSlug: this.projectSlug },
+      data: {
+        modelSlug: model.slug,
+        projectSlug: this.projectSlug,
+        path: this.modelService.backendURLFactory(
+          this.projectSlug ? this.projectSlug : '',
+          model.slug
+        ),
+      },
```
dominik003, line 74:
```suggestion
      data: {
        modelSlug: model.slug,
        projectSlug: this.projectSlug,
      },
```

dominik003 (changes requested):
Here are the missing comments, based on our discussion today.

backend/capellacollab/users/tokens/routes.py
```diff
@@ -0,0 +1,37 @@
...
+    user: users_models.DatabaseUser = fastapi.Depends(
+        user_injectables.get_existing_user
+    ),
+    db: orm.Session = fastapi.Depends(database.get_db),
+):
+    password = credentials.generate_password(32)
+    crud.create_token(db, user.id, password, "new_token")
+    new_token = user.name + ":" + password
+    return base64.b64encode(new_token.encode("ascii"))
```
Paula-Kli, line 37: we should return the bare password as capellambse does not use the `basic token` but rather the username and password and encodes them in the request then

backend/capellacollab/core/authentication/basic_auth.py
```diff
@@ -0,0 +1,51 @@
...
+import fastapi
+from fastapi import security, status
+
+from capellacollab.core import database
+from capellacollab.users import crud as user_crud
+from capellacollab.users.tokens import crud as token_crud
+
+
+class HTTPBasicAuth(security.HTTPBasic):
+    async def __call__(
```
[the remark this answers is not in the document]
Paula-Kli: Should we also add a comment why the type ignore is here? or will everybody easily find this as one can just have a look in the super class?

backend/capellacollab/core/authentication/injectables.py
```diff
@@ -18,25 +19,53 @@
...
+        if type(username) == dict:  # pylint: disable=unidiomatic-typecheck
+            username = helper.get_username(username)
         if not (user := users_crud.get_user_by_name(db, username)):
             if self.verify:
                 raise fastapi.HTTPException(
                     status_code=status.HTTP_404_NOT_FOUND,
                     detail={"reason": f"User {username} was not found"},
                 )
             return False
-
```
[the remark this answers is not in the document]
Paula-Kli: I do not have hard feelings about it

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.css
Paula-Kli: I tried to replace it with tailwind classes but it does not seem to work completely yet which is why I will leave the css here for the moment

frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.ts
Paula-Kli: good catch! Yes we do not need it anymore but not due to the change in the backend but rather due to the change of using the `UserService` to get the username

backend/capellacollab/core/authentication/basic_auth.py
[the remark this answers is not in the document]
dominik003: I would just ignore the type here. As you said, if someone is interested in why it is ignored, the first step would be to check the superclass and see that it is ignored there as well.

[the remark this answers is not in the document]
Paula-Kli: alright 👍

Paula-Kli:
will be refactored and rebuild - first step in #1005

PR #884, opened 2023-07-31
feat: Get model badge from artifacts

### LINKED ISSUE #754: Get the Model Badge from Pipelines Artifacts instead of Repository
As we want to add an option to let the user decide whether the model badge is being pushed to the repository, we should get the model badge in Collab Manager from the Pipelines Artifacts instead of the Repository.
This issue is related to the change requested in [py-capellambse](https://github.com/DSD-DBS/py-capellambse/issues/299)

Paula-Kli:
closes #754

### DIFF
--- a/backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
+++ b/backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
@@ -22,7 +22,7 @@ class GithubHandler(handler.GitHandler):
         return parse.urlparse(self.git_model.path).path
 
     async def get_last_job_run_id_for_git_model(
-        self, job_name: str, project_id: t.Optional[str] = None
+        self, job_name: str, project_id: str | None = None
     ) -> handler.JobIdAttributes:
         if not project_id:
             project_id = await self.get_project_id_by_git_url()
@@ -70,7 +70,7 @@ class GithubHandler(handler.GitHandler):
 
     async def get_file_from_repository(
         self, project_id: str, trusted_file_path: str
-    ) -> tuple[requests.Response, bytes]:
+    ) -> bytes:
         """
         If a repository is public but the permissions are not set correctly, you might be able to download the file without authentication
         but get an error when trying to load it authenticated.
@@ -88,10 +88,13 @@ class GithubHandler(handler.GitHandler):
                 headers=self.__get_headers(self.git_model.password),
             )
 
-        if response.ok:
-            return (response, base64.b64decode(response.json()["content"]))
-        else:
-            return (response, None)
+        if response.status_code == 404:
+            raise git_exceptions.GitRepositoryFileNotFoundError(
+                filename=trusted_file_path
+            )
+        response.raise_for_status()
+
+        return base64.b64decode(response.json()["content"])
 
     def get_last_pipeline_runs(
         self,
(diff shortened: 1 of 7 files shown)


backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -69,16 +69,14 @@ def __get_file_from_repository(
         )
 
     async def get_file_from_repository(
-        self,
-        trusted_file_path: str,
-    ) -> bytes:
+        self, project_id: str, trusted_file_path: str
+    ) -> tuple[requests.Response, bytes]:
```
dominik003, line 73: As discussed either adjust the type hint to match the case that bytes can also be None or change it back to just `bytes` and raise an error if no file found

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
```diff
@@ -39,10 +40,10 @@ async def get_project_id_by_git_url(self) -> str:
                 return (await response.json())["id"]
 
     async def get_last_job_run_id_for_git_model(
-        self,
-        job_name: str,
+        self, job_name: str, project_id: t.Optional[str] = None
```
dominik003, line 43:
```suggestion
        self, job_name: str, project_id: str | None = None
```
And I think the import can then also be removed

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
```diff
@@ -132,19 +133,15 @@ def get_artifact_from_job(
         return response
 
     async def get_file_from_repository(
-        self,
-        trusted_file_path: str,
-    ) -> bytes:
-        project_id = await self.get_project_id_by_git_url()
+        self, project_id: str, trusted_file_path: str
+    ) -> tuple[requests.Response, bytes | None]:
```
dominik003, line 137: In case you adjust the abstract function (i.e., simply use `bytes` as return type and otherwise raise an error) you also need to adjust this function

dominik003 (changes requested):

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -61,7 +61,29 @@ def get_artifact_from_job_as_content(
...
+
+    async def get_file_from_repository_or_artifacts(
+        self, trusted_file_path: str, job_name: str | None
+    ) -> bytes:
+        project_id = await self.get_project_id_by_git_url()
+        response, file = await self.get_file_from_repository(
+            project_id, trusted_file_path
+        )
+        if file:
+            return file
```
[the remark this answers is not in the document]
Paula-Kli: I think that should be fixed now

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -59,9 +58,32 @@ def get_artifact_from_job_as_content(
...
+            if job_name:
+                (
+                    _,
+                    (job_id, _),
+                ) = await self.get_last_job_run_id_for_git_model(
+                    job_name, project_id
+                )
+                return self.get_artifact_from_job_as_content(
+                    project_id, job_id, trusted_file_path
+                )
```
dominik003, line 85: I prefer here to simply put a `pass` into the except block. This decreases the nesting here and usually this increase the readability.
In addition, I would not put `Exception` in here as `requests.HTTPError` and `GitRepositoryFileNotFoundError` should be the only once that can be thrown or?

Paula-Kli: yes I think that should be the only two possibilities

dominik003 (approved):

MoritzWeber0 (approved):

PR #894, opened 2023-08-04
feat: Add project-wide Jupyter notebook "share" for collaboration

MoritzWeber0:
During creation of a Jupyter model in a project, a file-share is created. This feature enables project-wide file-shares. Together with internal projects, it can even serve as global file-share. Per default, the size of the volume is 2GB.

To delete a file-share, just delete the model from the project.

The feature is not complete yet. File-shares are created, but not mounted. The state of the Kubernetes operator is currently not good enough for mounting. We have another PR open to improve the situation. Once the `session hooks` PR is merged, we can build the mounting-feature for Jupyter file-shares: https://github.com/DSD-DBS/capella-collab-manager/pull/912.
Nevertheless, I would merge this PR and implement mounting in a later PR.

Depends on #900, #901, #902, #903, #904 and #905.

### DIFF
--- a/docs/user/docs/sessions/jupyter/collaboration.md
+++ b/docs/user/docs/sessions/jupyter/collaboration.md
@@ -1,0 +1,37 @@
+<!--
+ ~ SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+ ~ SPDX-License-Identifier: Apache-2.0
+ -->
+
+# Collaboration with Jupyter Notebooks
+
+Collaborating on Jupyter notebooks is a common requirement in various
+workflows. This guide describes the approach of collaboration in individual
+sessions.
+
+## Collaboration in Individual Sessions
+
+If you have a notebook in your personal workspace and another user wants to
+connect to your sessions, follow these steps:
+
+### Sharing a Notebook Session
+
+1. **Request a Jupyter Session**: As usual, request a Jupyter session to start.
+2. **Connect to the Session**: Open the requested session.
+3. **Share the Session**: Click on the "Share" button on the top right of the
+   Jupyter session interface.
+4. **Include the Token**: Tick the option "Include the token in the URL".
+5. **Warning! Security Risk**: Be aware that anyone with access to the URL can
+   access your Jupyter session until termination. To revoke access, terminate
+   your session in the Capella Collaboration Manager. After restarting it, a
+   new token will be generated, and old sessions will no longer have access.
+6. **Share the Link**: Distribute the link to all users who should have access.
+   They will be able to open your notebooks and concurrently edit them.
+
+#### Instructional Video
+
+Here's a video that visually guides you through these steps:
+
+<video controls>
+  <source src="../jupyter-collaboration.mp4" type="video/mp4">
+</video>
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,11 +1,8 @@
 # SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
 # SPDX-License-Identifier: CC0-1.0
-
 * text=auto
-
 *.{cmd,[cC][mM][dD]} text eol=crlf
 *.{bat,[bB][aA][tT]} text eol=crlf
-
 *.{sh,py} text eol=lf
-
 .git_archival.txt  export-subst
+*.mp4 filter=lfs diff=lfs merge=lfs -text
--- a/backend/capellacollab/alembic/versions/97c7acd616fc_add_configuration_to_toolmodels.py
+++ b/backend/capellacollab/alembic/versions/97c7acd616fc_add_configuration_to_toolmodels.py
@@ -1,0 +1,30 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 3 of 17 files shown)


docs/user/docs/sessions/jupyter/collaboration.md
```diff
@@ -0,0 +1,62 @@
...
+# Collaboration with Jupyter Notebooks
+
+Collaborating on Jupyter notebooks is a common requirement in various
+workflows. This guide describes two main approaches to collaboration:
+individual sessions and project-level collaboration.
+
+## Collaboration in Individual Sessions
+
+If you have a notebook in your personal workspace and another user wants to
+connect to your sessions, follow these steps:
```
Wuestengecko, line 15: This requires a newer (yet unreleased) version of capella-dockerimages, which is not pulled in by this PR as submodule update. I'm not sure what the release cycle here normally is – should this be mentioned somewhere, or will it just fall into place once everything gets merged together?

docs/user/docs/sessions/jupyter/collaboration.md
```diff
@@ -0,0 +1,62 @@
...
+Here's a video that visually guides you through these steps:
+
+<video controls>
+  <source src="../jupyter-collaboration.mp4" type="video/mp4">
+</video>
+
+## Collaboration on Project Level
+
+If you need a shared workspace for notebooks at the project level, you can
+create a shared space accessible by all project members.
```
Wuestengecko, line 42: IMO this (user) documentation should only land once the feature actually becomes visible to users. The most important thing here being that it won't go into production's docs before the actual feature can go into production as well. If we can make sure that's the case, it should be fine keeping it here, even though the actual (frontend and backend) functionality isn't quite there yet. Although personally I would rather like to see these docs together with the frontend update :)

docs/user/docs/sessions/jupyter/collaboration.md
```diff
@@ -0,0 +1,62 @@
...
+
+1. **Request a Jupyter Session**: As usual, request a Jupyter session to start.
+2. **Connect to the Session**: Open the requested session.
+3. **Share the Session**: Click on the "Share" button on the top right of the
+   Jupyter session interface.
+4. **Include the Token**: Tick the option "Include the token in the URL".
+5. **Warning! Security Risk**: Be aware that anyone with access to the URL can
+   access your Jupyter session until termination. To revoke access, terminate
+   your session in the Capella Collaboration Manager and optionally request a
+   new session.
```
Wuestengecko, line 27: Might clarify this point a bit.

```suggestion
   access your Jupyter session until termination. To revoke access, terminate
   your session in the Capella Collaboration Manager. After restarting it, a
   new token will be generated, and old sessions will no longer have access.
```

I found that, locally, there was a slight delay (maybe a minute or so) where the old Jupyter browser session (opened before terminating the session in collab manager) was still valid before it eventually disconnected for good. This might be worth investigating in more detail. However, I could not find a way to regain access after that disconnection, so I don't think it's a real problem.

Wuestengecko (commented):
I tried breaking it for a while, but it still runs. I guess this is as good a review as I can give currently :D

docs/user/docs/sessions/jupyter/collaboration.md
MoritzWeber0: Rule 1: Submodules are never up-to-date :D 

Joking aside, I'll update the submodule tomorrow and add the commit to this PR. In general, it doesn't affect the release process at all as sessions have their own lifecycle and the version has to be specified in the frontend for the production deployments.

docs/user/docs/sessions/jupyter/collaboration.md
MoritzWeber0: Good point, I'll move this part to a new PR, where I'll add the real feature later on.

docs/user/docs/sessions/jupyter/collaboration.md
MoritzWeber0: Update 1: I've created a new release in the capella-dockerimages repository.

MoritzWeber0: Update 2: Submodule is also updated.

docs/user/docs/sessions/jupyter/collaboration.md
MoritzWeber0: The documentation commit is edited and does only include the collaboration part. 

The file-share part is moved to a new PR, where I'll add the file-share mounting soon: https://github.com/DSD-DBS/capella-collab-manager/pull/918

docs/user/docs/sessions/jupyter/collaboration.md
MoritzWeber0: Jupyter itself is cached in the browser, so there might be a delay indeed. However, the backend and kernel should already be dead and cell execution etc. should result in errors like "kernel died unexpectedly". 
Did you still have access to the kernel after session termination? If yes, this might be a case where Jupyter didn't terminate after SIGTERM. Sessions have then 30 seconds to terminate properly. If Jupyter didn't do that, Kubernetes kills the Pod with SIGKILL.

MoritzWeber0: I opened a new PR for proper handling of the SIGTERM signal: https://github.com/DSD-DBS/capella-dockerimages/pull/186

PR #896, opened 2023-08-04
feat: Get diagram cache from repository

### LINKED ISSUE #895: Get diagram cache from repository
The github action gives the opportunity to retrieve the diagram cache form the repository instead of the artifacts, enabling the diagram cache to be used by open source projects without authentication

Paula-Kli:
Especially for open source projects it is helpful to retrieve the diagram cache from the repository as then no authentication is needed. In addition it speeds up the retrieval of the diagrams because we can seperately download them from the repository

closes #895


backend/tests/projects/toolmodels/conftest.py
```diff
@@ -58,6 +58,11 @@ def fixture_git_type(request: pytest.FixtureRequest) -> git_models.GitType:
     return request.param
 
 
+@pytest.fixture(name="status", params=[200])
+def fixture_status(request: pytest.FixtureRequest) -> int:
+    return request.param
```
MoritzWeber0, line 63: Please use a more descriptive name for it, for example `github_response_status`.

Also, it's only used in `test_diagrams.py`, so I'd also move it into the file.

MoritzWeber0 (changes requested):

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,17 +61,35 @@ def get_artifact_from_job_as_content(
...
+        self,
+        trusted_file_path: str,
+        job_name: str | None,
+        branch_name: str | None = None,
+    ) -> dict:
+        result = await self.get_file_from_repository_or_artifacts(
+            trusted_file_path, job_name, branch_name
+        )
+        result.decode("utf-8")
+        return json.loads(result)
```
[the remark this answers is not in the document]
Paula-Kli: We can retrieve the artifacts from gitlab directly as json (`get_artifact_from_job_as_json`) but to use the other interface and to have less code duplication I just retrieve in both cases the binary and convert it to json if possible basically - do you have another good idea to use the other one? or shall I just delete it as we are not using it anymore

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,26 +62,65 @@ def get_artifact_from_job_as_content(
...
+        self, project_id: str, file_path: str, revision: str | None
+    ) -> t.Any:
+        pass
+
+    async def get_file_from_repository_or_artifacts_as_json(
+        self,
+        trusted_file_path: str,
+        job_name: str | None,
+        revision: str | None = None,
+    ) -> tuple[t.Any, dict]:
```
MoritzWeber0, line 83:
```suggestion
    ) -> tuple[datetime.datetime, dict[str, t.Any]]:
```

backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -74,18 +70,11 @@ async def get_diagram(
...
-        last_successful_job,
-    ) = await handler.get_last_job_run_id_for_git_model(
-        "update_capella_diagram_cache"
-    )
-
     try:
-        diagram = handler.get_artifact_from_job_as_content(
-            project_id,
-            last_successful_job[0],
+        (_, diagram) = await handler.get_file_from_repository_or_artifacts(
```
MoritzWeber0, line 74:
```suggestion
        _, diagram = await handler.get_file_from_repository_or_artifacts(
```

backend/capellacollab/projects/toolmodels/modelbadge/routes.py
```diff
@@ -35,10 +35,14 @@ async def get_model_complexity_badge(
...
+            _,
+            model_badge,
+        ) = await git_handler.get_file_from_repository_or_artifacts(
+            "model-complexity-badge.svg", "generate-model-badge"
+        )
         return fastapi.responses.Response(
-            content=await git_handler.get_file_from_repository_or_artifacts(
-                "model-complexity-badge.svg", "generate-model-badge"
-            ),
+            content=model_badge,
```
MoritzWeber0, line 45:
```suggestion
            content=git_handler.get_file_from_repository_or_artifacts(
                "model-complexity-badge.svg", "generate-model-badge"
            )[1],
```

I'd prefer this solution as it saves two variable definitions.

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -60,16 +60,24 @@ def get_artifact_from_job_as_content(
         ).encode()
 
     def __get_file_from_repository(
-        self, project_id: str, trusted_file_path: str, headers: dict = None
+        self,
+        project_id: str,
+        trusted_file_path: str,
+        revision: str | None = None,
+        headers: dict | None = None,
```
MoritzWeber0, line 67:
```suggestion
        headers: dict[str, str] | None = None,
```

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -60,16 +60,24 @@ def get_artifact_from_job_as_content(
...
     ) -> requests.Response:
+        branch = revision if revision else self.git_model.revision
```
MoritzWeber0, line 69:
```suggestion
        revision = revision or self.git_model.revision
```

Ever better would it be to put `revision or self.git_model.revision` directly in the requests call.

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -60,16 +60,24 @@ def get_artifact_from_job_as_content(
...
         return requests.get(
-            f"{self.git_instance.api_url}/repos{project_id}/contents/{parse.quote(trusted_file_path, safe='')}?ref={parse.quote(self.git_model.revision, safe='')}",
+            f"{self.git_instance.api_url}/repos{project_id}/contents/{parse.quote(trusted_file_path)}?ref={parse.quote(branch, safe='')}",
```
MoritzWeber0, line 71:
```suggestion
            f"{self.git_instance.api_url}/repos/{project_id}/contents/{parse.quote(trusted_file_path)}?ref={parse.quote(branch, safe='')}",
```

Without a separating `/`, the URL looks a bit weird. Instead, I'd change the project_id to be in the format `{owner}/{repo_name}` (remove the leading `/`).
It has to be updated in all references of the Github URL.

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -131,6 +140,25 @@ def get_artifact_from_job(
...
+    ) -> t.Any:
+        if revision:
+            attribute = f"rev={revision}"
+        else:
+            attribute = f"path={file_path}&rev={self.git_model.revision}"
+        response = requests.get(
+            f"{self.git_instance.api_url}/repos{project_id}/commits?{attribute}",
+            headers=self.__get_headers(self.git_model.password)
+            if self.git_model.password
+            else "",
```
MoritzWeber0, line 154:
```suggestion
            else None,
```

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -131,6 +140,25 @@ def get_artifact_from_job(
...
+            timeout=config["requests"]["timeout"],
+        )
+        response.raise_for_status()
+        if len(response.json()) == 0:
+            return ""
```
MoritzWeber0, line 159:
```suggestion
            return None
```

When does it run into this case? If the file is not found?

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -131,6 +140,25 @@ def get_artifact_from_job(
...
 
+    def get_last_commit_id_for_file_path(
+        self, project_id: str, file_path: str, revision: str | None
+    ) -> t.Any:
+        if revision:
+            attribute = f"rev={revision}"
+        else:
+            attribute = f"path={file_path}&rev={self.git_model.revision}"
+        response = requests.get(
+            f"{self.git_instance.api_url}/repos{project_id}/commits?{attribute}",
```
MoritzWeber0, line 151:
```suggestion
        response = requests.get(
            f"{self.git_instance.api_url}/repos{project_id}/commits?path={file_path}&rev={revision or self.git_model.revision}",
            headers=self.__get_headers(self.git_model.password)
```

Why is the `file_path` ignored when revision is set?

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
```diff
@@ -53,6 +54,20 @@ async def get_last_job_run_id_for_git_model(
 
         raise git_exceptions.GitPipelineJobNotFoundError(job_name=job_name)
 
+    def get_last_commit_id_for_file_path(
+        self, project_id: str, file_path: str, revision: str | None
+    ) -> t.Any:
+        revision = revision if revision else self.git_model.revision
+        response = requests.get(
+            f"{self.git_instance.api_url}/projects/{project_id}/repository/commits?ref_name={revision}&path={file_path}",
```
MoritzWeber0, line 62:
```suggestion
        response = requests.get(
            f"{self.git_instance.api_url}/projects/{project_id}/repository/commits?ref_name={revision or self.git_model.revision}&path={file_path}",
```

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
```diff
@@ -53,6 +54,20 @@ async def get_last_job_run_id_for_git_model(
...
+            headers={"PRIVATE-TOKEN": self.git_model.password},
+            timeout=config["requests"]["timeout"],
+        )
+        response.raise_for_status()
+        if len(response.json()) == 0:
+            return ""
```
MoritzWeber0, line 68:
```suggestion
            return None
```

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
```diff
@@ -53,6 +54,20 @@ async def get_last_job_run_id_for_git_model(
 
         raise git_exceptions.GitPipelineJobNotFoundError(job_name=job_name)
 
+    def get_last_commit_id_for_file_path(
+        self, project_id: str, file_path: str, revision: str | None
+    ) -> t.Any:
```
MoritzWeber0, line 59:
```suggestion
    ) -> datetime.datetime | None:
```

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
```diff
@@ -53,6 +54,20 @@ async def get_last_job_run_id_for_git_model(
 
         raise git_exceptions.GitPipelineJobNotFoundError(job_name=job_name)
 
+    def get_last_commit_id_for_file_path(
```
MoritzWeber0, line 57:
```suggestion
    def get_last_updated_for_file_path(
```

You're not returning the commit id. Instead, the `last_updated` timestamp is returned.

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -131,6 +140,25 @@ def get_artifact_from_job(
             artifact_response, trusted_path_to_artifact
         )
 
+    def get_last_commit_id_for_file_path(
```
MoritzWeber0, line 143: The function returns the last_updated timestamp, not a the commit id.

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,26 +62,65 @@ def get_artifact_from_job_as_content(
...
+        project_id: str,
+        trusted_file_path: str,
+        revision: str | None = None,
     ) -> bytes:
         pass
 
+    @abc.abstractmethod
+    def get_last_commit_id_for_file_path(
+        self, project_id: str, file_path: str, revision: str | None
+    ) -> t.Any:
```
MoritzWeber0, line 75:
```suggestion
    ) -> datetime.datetime | None:
```

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,26 +62,65 @@ def get_artifact_from_job_as_content(
...
-        self, project_id: str, trusted_file_path: str
+        self,
+        project_id: str,
+        trusted_file_path: str,
+        revision: str | None = None,
     ) -> bytes:
         pass
 
+    @abc.abstractmethod
+    def get_last_commit_id_for_file_path(
```
MoritzWeber0, line 73:
```suggestion
    def get_last_updated_for_file_path( 
```

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,26 +62,65 @@ def get_artifact_from_job_as_content(
...
+            ) = await self.get_last_job_run_id_for_git_model(
                 job_name, project_id
             )
-            return self.get_artifact_from_job_as_content(
-                project_id, job_id, trusted_file_path
+            return (
+                last_updated,
+                self.get_artifact_from_job_as_content(
+                    project_id, job_id, trusted_file_path
+                ),
```
MoritzWeber0, line 123: Error handling is missing here.

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,26 +62,65 @@ def get_artifact_from_job_as_content(
...
+            trusted_file_path, job_name, revision
+        )
+        return (last_updated, json.loads(result.decode("utf-8")))
+
     async def get_file_from_repository_or_artifacts(
-        self, trusted_file_path: str, job_name: str | None
-    ) -> bytes:
+        self,
+        trusted_file_path: str,
+        job_name: str | None,
```
MoritzWeber0, line 95: `job_name` shouldn't be optional here. The purpose of the function is to get the file from repository or artifacts. If there is no job_name, the user can call the get_file_from_repository function directly.

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,26 +62,65 @@ def get_artifact_from_job_as_content(
...
+                    trusted_file_path,
+                    revision=revision,
+                ),
+                await self.get_file_from_repository(
+                    project_id, trusted_file_path, revision
+                ),
             )
         except (requests.HTTPError, exceptions.GitRepositoryFileNotFoundError):
             pass
         if job_name:
```
MoritzWeber0, line 112: The if can be removed if job_name is not optional anymore.

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -60,26 +62,65 @@ def get_artifact_from_job_as_content(
...
-            (_, (job_id, _),) = await self.get_last_job_run_id_for_git_model(
+            (
+                _,
+                (job_id, last_updated),
+            ) = await self.get_last_job_run_id_for_git_model(
```
MoritzWeber0, line 116: `get_last_job_run_id_for_git_model` gets the project_id as input parameter and returns it again. This doesn't make sense and the returned project_id is also not used.

backend/tests/projects/toolmodels/conftest.py
```diff
@@ -172,6 +172,113 @@ def fixture_mock_git_rest_api_for_artifacts(
...
+                        "author_email": "test-email",
+                        "authored_date": "2050-04-11T10:09:59.000+02:00",
+                        "committer_name": "test-name",
+                        "committer_email": "test-email",
+                        "committed_date": "2050-04-11T10:09:59.000+02:00",
+                    }
+                ],
+            )
+            responses.get(
+                "https://example.com/api/v4/projects/10000/repository/commits?ref_name=diagram-cache/main&path=diagram_cache/_c90e4Hdf2d2UosmJBo0GTw.svg",
```
MoritzWeber0, line 200: Instead of duplicating the same call three times, you can use a regex expression to match all request calls with a specific wildcard: https://github.com/getsentry/responses#dynamic-responses

backend/tests/projects/toolmodels/conftest.py
```diff
@@ -172,6 +172,113 @@ def fixture_mock_git_rest_api_for_artifacts(
...
+                        "author_email": "test-email",
+                        "authored_date": "2050-04-11T10:09:59.000+02:00",
+                        "committer_name": "test-name",
+                        "committer_email": "test-email",
+                        "committed_date": "2050-04-11T10:09:59.000+02:00",
+                    }
+                ],
+            )
+        case git_models.GitType.GITHUB:
+            responses.get(
```
MoritzWeber0, line 236: Same comment as above. Use a compiled regex here instead of duplicating the call.

backend/tests/projects/toolmodels/test_diagrams.py
```diff
@@ -65,13 +71,84 @@ def fixture_mock_git_diagram_cache_index_api(git_type: git_models.GitType):
...
+                    "file_name": "index.json",
+                    "file_path": "diagram_cache/index.json",
+                    "size": 2428,
+                    "encoding": "base64",
+                    "content_sha256": "ec00b56c2c87897dbd6e8d42de61b8ec6e3337cc442a09add86560920ec2872a",
+                    "ref": "main",
+                    "blob_id": "ee4b99b97f008dff549de9a40ddab139ea44435b",
+                    "commit_id": "1273b25fe671e014afb38d594825bb1a9d3206df",
+                    "last_commit_id": "85306aed4cf34c315ca3c31659fde02df9896e98",
+                    "execute_filemode": False,
```
MoritzWeber0, line 100: These are attributes are the same for all calls. Please store them in a variable instead.

MoritzWeber0 (changes requested):

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
Paula-Kli: as a heads up: that is a weird thing in the github api: usually it was totally fine if just the token was None but in that specific case if the token was none it worked as not authenticated while no headers at all worked fine... so e.g. in `__get_file_from_repository` it is ok if the token is None but not here

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
Paula-Kli: what do you mean by that? aren't the errors handled in the lower layers?

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
MoritzWeber0: But you're setting the headers to "", not the token.

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
MoritzWeber0: For the case where the file is not found, we can throw `git_exceptions.GitRepositoryFileNotFoundError`. This will make it much easier to understand. 

The second case I don't understand. Even on orphaned branches there should be at least one commit if the file exists. Could you please explain what you mean?

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
MoritzWeber0: We should always only get last updated state of the referenced files. We have to ignore other file updated in the repository, e.g., changes of the README.

backend/tests/projects/toolmodels/test_diagrams.py
Paula-Kli: does it in general make sense to keep that here? It should mainly show the structure of a response while it is already stripped a bit - should we remove all currently unused attributes?

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
MoritzWeber0: Indeed, you're right. I'll mark the comment as resolved.

MoritzWeber0 (changes requested):
In general, looks really good now. However, we should prefix the revision with `diagram-cache/`. 

The use-case is: read-only sessions and model badge are fetched from the `main` branch, but the diagram cache is served from the `diagram-cache/main` branch. When changing the revision in the linked Git repository, model badge and read-only sessions stop working, so this is not an option.

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -3,6 +3,9 @@
```
MoritzWeber0, line 3: I can't get it to run. The stacktrace is: 

```py
Traceback (most recent call last):
  File "/Users/moritzweber/Projects/github/capella-collab-manager/backend/capellacollab/projects/toolmodels/diagrams/routes.py", line 44, in get_diagram_metadata
    ) = await handler.get_file_from_repository_or_artifacts_as_json(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/moritzweber/Projects/github/capella-collab-manager/backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py", line 84, in get_file_from_repository_or_artifacts_as_json
    ) = await self.get_file_from_repository_or_artifacts(
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/moritzweber/Projects/github/capella-collab-manager/backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py", line 110, in get_file_from_repository_or_artifacts
    job_id, last_updated = await self.get_last_job_run_id_for_git_model(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/moritzweber/Projects/github/capella-collab-manager/backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py", line 30, in get_last_job_run_id_for_git_model
    jobs = self.get_last_pipeline_runs(project_id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/moritzweber/Projects/github/capella-collab-manager/backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py", line 118, in get_last_pipeline_runs
    response.raise_for_status()
  File "/Users/moritzweber/Projects/github/capella-collab-manager/backend/.venv/lib/python3.11/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://api.github.com/repos/DSD-DBS/coffee-machine.git/actions/runs?branch=diagram-cache%2Fmain&per_page=20
```

Paula-Kli:
> In general, looks really good now. However, we should prefix the revision with `diagram-cache/`.
> 
> The use-case is: read-only sessions and model badge are fetched from the `main` branch, but the diagram cache is served from the `diagram-cache/main` branch. When changing the revision in the linked Git repository, model badge and read-only sessions stop working, so this is not an option.

Why do they stop working? They are always fetched from the current revision aren't they?

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
Paula-Kli: what puzzles me is why you are trying to get the actions from the `diagram-cache/main` branch? so it is correct that there will be no action found on that branch as there is no action for that branch.

MoritzWeber0:
> > In general, looks really good now. However, we should prefix the revision with `diagram-cache/`.
> > The use-case is: read-only sessions and model badge are fetched from the `main` branch, but the diagram cache is served from the `diagram-cache/main` branch. When changing the revision in the linked Git repository, model badge and read-only sessions stop working, so this is not an option.
> 
> Why do they stop working? They are always fetched from the current revision aren't they?

Yes, they are. But the model badge doesn't exist on the `diagram-cache/*` branches. Let's take coffee machine as example: 

- If we would use `main` as revision, the diagram cache would be fetched from the branch `main`, however, we want to fetch it from `diagram-cache/main`.
- If we would use `diagram-cache/main`, the `model-complexity-badge.svg` doesn't exist: https://github.com/DSD-DBS/coffee-machine/tree/diagram-cache/main

MoritzWeber0 (approved):

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
```diff
@@ -49,10 +50,23 @@ async def get_last_job_run_id_for_git_model(
...
+        self, project_id: str, file_path: str, revision: str | None
+    ) -> datetime.datetime | None:
+        response = requests.get(
+            f"{self.git_instance.api_url}/projects/{project_id}/repository/commits?ref_name={revision or self.git_model.revision}&path={file_path}",
+            headers={"PRIVATE-TOKEN": self.git_model.password},
+            timeout=config["requests"]["timeout"],
+        )
+        response.raise_for_status()
+        if len(response.json()) == 0:
+            return None
```
MoritzWeber0, line 67: Here is another occurrence of `len(response.json()) == 0` where `git_exceptions.GitRepositoryFileNotFoundError` can be used.

MoritzWeber0 (changes requested):

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/handler.py
Paula-Kli: good catch

PR #900, opened 2023-08-07
fix: Display log out button in mobile view

MoritzWeber0:
In the mobile view, the log out button was only
displayed for administrators, not for users.

This commit adds the Log out button for all users
in the mobile navigation.


amolenaar (approved):
:shipit:

PR #901, opened 2023-08-07
feat: Add option to skip source linking

MoritzWeber0:
Sometimes, users don't have a source ready during model creation. In these cases, they now can skip the source creation.


amolenaar (approved):
:shipit:

PR #902, opened 2023-08-07
refactor: Add custom ExistingDependenciesError

MoritzWeber0:
Some elements can only be deleted or modified if
no objects depend on the object to delete.

If this rule is not met, we throw the new
ExistingDependenciesError.


amolenaar (approved):
Handling the translation from exception to HTTP status code is a lot cleaner like this :+1:

PR #904, opened 2023-08-07
fix: Add backend validation for deps during model deletion

MoritzWeber0:
If a toolmodel had linked t4c or git repositories, the request return a success status even though nothing was deleted.


amolenaar (approved):
Nice one. This provides a more human friendly way that hitting on a database error.

PR #905, opened 2023-08-07
refactor: Add `UserNotFoundError` exception

MoritzWeber0:


backend/capellacollab/users/exceptions.py
```diff
@@ -0,0 +1,34 @@
...
+async def user_not_found_exception_handler(
+    request: fastapi.Request, exc: UserNotFoundError
+) -> fastapi.Response:
+    return await exception_handlers.http_exception_handler(
+        request,
+        fastapi.HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail={
+                "title": "User not found",
+                "reason": f"The user '{exc.username or exc.user_id}' doesn't exist.",
```
amolenaar, line 25: My first reaction is: we should not expose the user id, to avoid malicious intent.

But I see we've been exposing the id in several places anyway.

amolenaar (approved):
Just one remark about exposing user ids. However, since we already do that, I think it's fine for now.

backend/capellacollab/users/exceptions.py
MoritzWeber0: The user ID is only returned if your request contains the ID. 

For example, sending a requests to `/api/v1.0/users/42` returns `The user '42' doesn't exist.`. There is no additional information returned in the request. Actually, we can't return any existing user id here as it's the error case where the user doesn't exist.

PR #914, opened 2023-08-08
refactor: Change method and class import to module imports

MoritzWeber0:
The Google style guide proposes module imports.
In this PR, some old-style imports are changed to the new style.

### DIFF
--- a/backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
+++ b/backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
@@ -11,33 +11,24 @@ from sqlalchemy import orm
 
 from capellacollab.core import database
 from capellacollab.core.authentication import injectables as auth_injectables
+from capellacollab.projects.toolmodels import (
+    injectables as toolmodels_injectables,
+)
+from capellacollab.projects.toolmodels import models as toolmodels_models
 from capellacollab.projects.toolmodels.backups import crud as backups_crud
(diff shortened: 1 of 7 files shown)


backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
```diff
@@ -11,33 +11,29 @@
...
-from capellacollab.projects.toolmodels.modelsources.git.models import (
-    DatabaseGitModel,
-    GitModel,
-    PatchGitModel,
-    PostGitModel,
+from capellacollab.projects.toolmodels import models as toolmodels_models
+from capellacollab.projects.toolmodels.backups import crud as backups_crud
+from capellacollab.projects.toolmodels.modelsources.git import (
+    models as git_models,
 )
```
Paula-Kli, line 21: why are we not importing this as `from . import models` as it is within the same module?

Paula-Kli (commented):
good Idea to make this more consistent 👍

backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
Paula-Kli: ok here it is referenced that no relative imports should be used https://google.github.io/styleguide/pyguide.html#224-decision

MoritzWeber0: Good catch, thanks for finding it! Is fixed.

PR #917, opened 2023-08-08
fix: Add timezone awareness in frontend and backend

MoritzWeber0:
The backend has used different timezones to store datetimes. Sometimes, the local ast timezone was used, sometimes UTC.

In this commit, all datetime objects are now stored in UTC. The backend adds the timezone information to all backend responses.

With the new timezone in the response, Angular automatically maps the datetime to the client timezone.

Thanks to @Wuestengecko for finding this bug.

### DIFF
--- a/backend/capellacollab/projects/toolmodels/backups/runs/interface.py
+++ b/backend/capellacollab/projects/toolmodels/backups/runs/interface.py
@@ -5,7 +5,6 @@ import asyncio
 import datetime
 import logging
 
-import pytz
 from kubernetes import client as k8s_client
 from kubernetes.client import exceptions as k8s_exceptions
 from starlette import concurrency as starlette_concurrency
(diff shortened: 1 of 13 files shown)


backend/capellacollab/projects/toolmodels/backups/runs/interface.py
```diff
@@ -93,35 +93,28 @@ def _schedule_pending_jobs():
...
+    event_timestamp: datetime.datetime = (
+        event.last_timestamp or event.event_time
+    )
     timestamp = (
-        _transform_utc_to_local_timestamp(
-            event.last_timestamp or event.event_time
-        )
-        if event.last_timestamp or event.event_time
-        else datetime.datetime.now()
+        event_timestamp.replace(tzinfo=pytz.UTC)
```
Wuestengecko, line 115: I think this would reinterpret the timestamp as UTC instead of converting it – is this right? In other words, we get a UTC timestamp without attached timezone info? If we get localtime, we need to use `astimezone()` instead of `replace()`.

MoritzWeber0: The `event_timestamp` comes from the Kubernetes API as response. I did just expect that Kubernetes always runs in UTC, but this doesn't to seem the case (it's the case for all of our environment though).

I've checked the response from Kubernetes Python API and the datetime attributes are timezone-aware. 

The datetime comparision also is timezone-aware, so we don't need to replace/convert here at all. 

```py
>>> now = datetime.datetime.now(datetime.UTC)
>>> now.astimezone(datetime.UTC) - now.astimezone(zoneinfo.ZoneInfo("America/Los_Angeles"))
datetime.timedelta(0)
```

Also, `timestamp()` is timezone-aware: 

```py
>>> now.astimezone(datetime.UTC).timestamp()
1691586121.067041
>>> now.astimezone(zoneinfo.ZoneInfo("America/Los_Angeles")).timestamp()
1691586121.067041
```

I'll remove the `.replace()` here.

Issue #678, opened 2023-04-04
Collaborate on Jupyter notebooks


amolenaar:
Currently users can have individual notebook instances. It would be great if we can allow notebooks on _project_ level. That would make it easier for users to share notebooks and scripts.

For this to work, we need to be able to import (Capella) models in the shared notebook environment.

**Open Questions**

* Can and should we differentiate between users in the notebook environment? It would make it easier to see who changed what.
* How do we decide which models to load in the shared notebook environment?
* Are there any security risks involved?

MoritzWeber0:
Instead of creating a shared persistent volume claim for each project, I'd use a more "explicit" approach. As we already have the infrastructure for toolmodels within projects and "Jupyter" can also be selected as tool, I'd prefer this approach:

- To create a shared volume, a model with the tool "Jupyter" in a project has to be created. During creation, the PVC is created.
- If the user has access to the project with `read-only` permission, the PVC is mounted as read-only volume in Jupyter sessions. 
- If the user has access to the project with `write` permission, the PVC is mounted as read/write volume in Jupyter sessions. 
- When the model is deleted, the PVC is also deleted.

Also, it could benefit from workspace automation in the future, e.g. to run regular/nightly backups of the Jupyter notebooks to a linked Git repository (like we have to for TeamForCapella).

PR #918, opened 2023-08-09
feat: Add mounting for project-level Jupyter file-shares

MoritzWeber0:
This is a follow-up of https://github.com/DSD-DBS/capella-collab-manager/pull/894.

In https://github.com/DSD-DBS/capella-collab-manager/pull/894, project-wide Jupyter file-shared were added, but not mounted to sessions. This missing bit is added in this PR.

The shares are mounted to `/shared/<project_slug>/<toolmodel_slug>`. To access the shares directly, a symlink from `/workspace/notebooks/shared` to `/shared` is required in the session container. This was added in https://github.com/DSD-DBS/capella-dockerimages/pull/187.

We do consider the access permission for users in projects:

- All users can see the file-shares from internal projects in a read-only mode.
- Administrators can access all file-shares in a read/write-mode
- Project members of private projects have access according to their permission (read-only or read/write).
- If a user has no access to a project, file-shares of the project are not visible for the user.

To prevent data-loss, we forbid writing files to `/shared` and `/shared/*`. These paths can not be mapped to a file-share.

Resolves https://github.com/DSD-DBS/capella-collab-manager/issues/678

### DIFF
--- a/backend/capellacollab/sessions/hooks/jupyter.py
+++ b/backend/capellacollab/sessions/hooks/jupyter.py
@@ -2,14 +2,22 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import logging
+import pathlib
 import typing as t
 from urllib import parse as urllib_parse
+
+from sqlalchemy import orm
 
 from capellacollab.config import config
 from capellacollab.core import credentials
 from capellacollab.core import models as core_models
+from capellacollab.core.authentication import injectables as auth_injectables
+from capellacollab.projects.toolmodels import crud as toolmodels_crud
+from capellacollab.projects.toolmodels import models as toolmodels_models
+from capellacollab.projects.users import models as projects_users_models
 from capellacollab.sessions import operators
 from capellacollab.sessions.operators import models as operators_models
+from capellacollab.tools import models as tools_models
 from capellacollab.users import models as users_models
 
 from .. import models as sessions_models
@@ -22,6 +30,7 @@ class JupyterConfigEnvironment(t.TypedDict):
     JUPYTER_BASE_URL: str
     JUPYTER_TOKEN: str
     JUPYTER_PORT: str
+    JUPYTER_URI: str
     CSP_ORIGIN_HOST: str
 
 
@@ -34,7 +43,10 @@ class JupyterIntegration(interface.HookRegistration):
 
     def configuration_hook(
         self,
+        db: orm.Session,
         user: users_models.DatabaseUser,
+        tool: tools_models.DatabaseTool,
+        token: dict[str, t.Any],
         **kwargs,
     ) -> tuple[
         JupyterConfigEnvironment,
@@ -43,7 +55,7 @@ class JupyterIntegration(interface.HookRegistration):
     ]:
         jupyter_token = credentials.generate_password(length=64)
 
-        environment = {
+        environment: JupyterConfigEnvironment = {
             "JUPYTER_TOKEN": jupyter_token,
             "JUPYTER_BASE_URL": self._determine_base_url(user.name),
             "JUPYTER_PORT": "8888",
@@ -51,7 +63,9 @@ class JupyterIntegration(interface.HookRegistration):
             "CSP_ORIGIN_HOST": f"{self._general_conf.get('scheme')}://{self._general_conf.get('host')}:{self._general_conf.get('port')}",
         }
 
-        return environment, [], []
+        volumes = self._get_project_share_volume_mounts(db, token, tool)
+
+        return environment, volumes, []
 
     def post_session_creation_hook(
         self,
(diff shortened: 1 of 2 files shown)


backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -77,3 +88,38 @@ def pre_session_termination_hook(
...
+        tool: tools_models.DatabaseTool,
+    ) -> list[operators_models.Volume]:
+        volumes = []
+
+        for model in toolmodels_crud.get_models_by_tool(db, tool.id):
+            if model.configuration and "workspace" in model.configuration:
+                if not auth_injectables.ProjectRoleVerification(
+                    required_role=projects_users_models.ProjectUserRole.USER,
+                    verify=False,
+                )(model.project.slug, token, db):
```
dominik003, line 105: Can we move this to a `is_project_member` function? Even though we plan to change the permissions at some point, it is a really nice descriptive function

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -77,3 +88,38 @@ def pre_session_termination_hook(
...
+                    # No access to the project
+                    continue
+
+                read_only = not auth_injectables.ProjectRoleVerification(
+                    required_role=projects_users_models.ProjectUserRole.USER,
+                    required_permission=projects_users_models.ProjectUserPermission.WRITE,
+                    verify=False,
+                )(model.project.slug, token, db)
```
dominik003, line 113: Same as above. Can you move this to a `has_project_write_access` function?

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -77,3 +88,38 @@ def pre_session_termination_hook(
...
+
+                volumes.append(
+                    operators_models.PersistentVolume(
+                        name=model.configuration["workspace"],
+                        read_only=read_only,
+                        container_path=f"/shared/{model.project.slug}/{model.slug}",
```
dominik003, line 119: You just either adjust the type of `container_path` to also allow `str` or provide a `pathlib.PurePosixPath` here

dominik003 (changes requested):

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -34,7 +40,10 @@ def __init__(self):
 
     def configuration_hook(
         self,
+        db: orm.Session,
         user: users_models.DatabaseUser,
+        tool: tools_models.DatabaseTool,
+        token: dict[str, t.Any],
         **kwargs,
     ) -> tuple[
         JupyterConfigEnvironment,
```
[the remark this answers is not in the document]
MoritzWeber0: > I think the `JupyterConfigEnvironment` is missing the `JUPYTER_URI` field.

Is added.

> In addition the `configuration_hook` shows a mypy error for me (as `JupyterConfigEnvironment` is not considered to be a `dict[str, str]` even though it is. Do you know any good way to handle this?

A simple `environment: JupyterConfigEnvironment` seems to do it.

> Furthermore, the expected arguments of the `HookRegistration` and the `JupyterIntegration` are slightly different, e.g., the operator is missing. Isn't this somehow required or should be done when using inheritance here?

This is possible because we always call the hooks with named arguments, never with positional arguments. The intention is to remove the number of unused arguments on the different integrations. Each integration should just "request" the parameters, which are required.

backend/capellacollab/sessions/hooks/jupyter.py
MoritzWeber0: Good comment, is moved 👍

backend/capellacollab/sessions/hooks/jupyter.py
MoritzWeber0: Done.

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -77,3 +88,38 @@ def pre_session_termination_hook(
...
+    def _get_project_share_volume_mounts(
+        self,
+        db: orm.Session,
+        token: dict[str, t.Any],
+        tool: tools_models.DatabaseTool,
+    ) -> list[operators_models.Volume]:
+        volumes = []
+
+        for model in toolmodels_crud.get_models_by_tool(db, tool.id):
+            if model.configuration and "workspace" in model.configuration:
```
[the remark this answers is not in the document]
MoritzWeber0: What do think of a list comprehension instead? The variable name provides the descriptive name and the nesting depth is reduced: 

```py
accessible_models_with_workspace_configuration = [
    model
    for model in toolmodels_crud.get_models_by_tool(db, tool.id)
    if model.configuration
    and "workspace" in model.configuration
    and self._is_project_member(model, token, db)
]

for model in models_with_workspace_configuration:
```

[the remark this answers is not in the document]
MoritzWeber0: I've changed it to the list comprehension. If you're ok with it, mark the conversation as resolved. Otherwise, just add a comment here.

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -77,3 +91,60 @@ def pre_session_termination_hook(
...
+        accessible_models_with_workspace_configuration = [
+            model
+            for model in toolmodels_crud.get_models_by_tool(db, tool.id)
+            if model.configuration
+            and "workspace" in model.configuration
+            and self._is_project_member(model, token, db)
+        ]
+
+        for model in accessible_models_with_workspace_configuration:
+            assert model.configuration
```
dominik003, line 112: Do we need this here for mypy/pylint? Because as far as I can see `accessible_models_with_workspace_configuration` should only contain models where `model.configuration` exists

dominik003 (approved):
LGTM! Nice changes and improvements!

backend/capellacollab/sessions/hooks/jupyter.py
MoritzWeber0: Yes, mypy doesn't get the list comprehension. It did work with a simple for loop, but list comprehension seems to be to complex for mypy.

PR #919, opened 2023-08-10
refactor: Move conflicting persistent session check to method

MoritzWeber0:
This makes the request_persistent_session function a lot more lightweight.
Additionally, the check whether a session is a Jupyter session depends on
the activated Jupyter integration instead of the tool name now.

Should be merged after https://github.com/DSD-DBS/capella-collab-manager/pull/912 is merged.


amolenaar (approved):
Makes sense to now clutter the main flow.

PR #921, opened 2023-08-10
refactor(sessions): Replace `auth_helper` with DatabaseUser

MoritzWeber0:
The auth_helper is depending on the low-level JWT token. As we also want to use other forms of authentication in the future, we should slowly remove dependencies from the JWT token.

Should be merged after https://github.com/DSD-DBS/capella-collab-manager/pull/912 is merged.


amolenaar (approved):
Nice cleanup.

PR #922, opened 2023-08-10
build: Bump frontend versions

Paula-Kli:


frontend/package.json
MoritzWeber0, line 1: Some dependencies are also listed in the `.pre-commit-config.yaml`. Please update them according to the new versions in the `package.json`.

MoritzWeber0, line 1: I'd not use `fix` as type in the commit message. It's not a bug that we're fixing. Instead, I always use `build` to bump dependencies. Please update the commit message and the PR title.

MoritzWeber0 (changes requested):

MoritzWeber0 (approved):

PR #923, opened 2023-08-10
refactor: Cleanup model sources

MoritzWeber0:
- TeamForCapella repositories are only displayed if the user is administrator and the tool has a TeamForCapella integration.
- The margin was moved from the inner mat-card to the outer a-element. The cursor doesn't trigger in the margin area anymore.
- The never implemented "Create new repository" for Git & T4C were removed.


amolenaar (approved):
Looks good

PR #924, opened 2023-08-11
fix: Delete public route in Jupyter hook instead T4C hooks

MoritzWeber0:
The public route for Juypter was deleted if the T4C integration was enabled and not if the Jupyter extension was enabled.

This bug may lead to some additional action, as some routes may not be deleted properly since the release v1.18.2 yesterday. Please check for dangling Ingress or OpenShift routes in your cluster.


amolenaar (approved):
This makes more sense.

PR #926, opened 2023-08-11
Clean up status codes

### LINKED ISSUE #668: Review and change response codes in code with Gitlab API calls
I would reserve 500 for situations we did not anticipate. It's not a fault in the backend service that the status badge can't be loaded. In the end it's a user error. The backend handles it, it does not fail in an unforeseen way (e.i. 500).

_Originally posted by @amolenaar in https://github.com/DSD-DBS/capella-collab-manager/pull/665#discussion_r1154598380_
            
This applies to all endpoints which use a connection to Git/Gitlab.

amolenaar:
Use 4xx status codes for issues that appear in our system. 
Not all "unhappy" flows are error flows. The application anticipates a lot of unhappy cases and should act accordingly.

Fixes #668.

### DIFF
--- a/backend/capellacollab/projects/toolmodels/diagrams/routes.py
+++ b/backend/capellacollab/projects/toolmodels/diagrams/routes.py
@@ -52,7 +52,7 @@ async def get_diagram_metadata(
     except requests.exceptions.HTTPError:
         logger.info("Failed fetching diagram metadata", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_404_NOT_FOUND,
             detail={
                 "reason": (
                     "The diagram cache is not configured properly.",
@@ -90,7 +90,7 @@ async def get_diagram(
     except requests.exceptions.HTTPError:
         logger.info("Failed fetching diagram", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_404_NOT_FOUND,
             detail={
                 "reason": (
                     "The diagram cache is not configured properly.",
--- a/backend/capellacollab/projects/toolmodels/diagrams/routes.py
+++ b/backend/capellacollab/projects/toolmodels/diagrams/routes.py
@@ -52,7 +52,7 @@ async def get_diagram_metadata(
     except requests.exceptions.HTTPError:
         logger.info("Failed fetching diagram metadata", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail={
                 "reason": (
                     "The diagram cache is not configured properly.",
@@ -90,7 +90,7 @@ async def get_diagram(
     except requests.exceptions.HTTPError:
         logger.info("Failed fetching diagram", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail={
                 "reason": (
                     "The diagram cache is not configured properly.",
--- a/backend/capellacollab/projects/toolmodels/modelbadge/routes.py
+++ b/backend/capellacollab/projects/toolmodels/modelbadge/routes.py
@@ -44,7 +44,7 @@ async def get_model_complexity_badge(
     except (aiohttp.web.HTTPException, requests.exceptions.HTTPError):
         logger.info("Failed fetching model complexity badge", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_404_NOT_FOUND,
             detail={
                 "reason": (
                     "The model complexity badge integration is not configured properly.",
--- a/backend/capellacollab/projects/toolmodels/modelbadge/routes.py
+++ b/backend/capellacollab/projects/toolmodels/modelbadge/routes.py
@@ -44,7 +44,7 @@ async def get_model_complexity_badge(
     except (aiohttp.web.HTTPException, requests.exceptions.HTTPError):
         logger.info("Failed fetching model complexity badge", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_404_NOT_FOUND,
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
             detail={
                 "reason": (
                     "The model complexity badge integration is not configured properly.",
--- a/backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
+++ b/backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
@@ -116,7 +116,7 @@ async def github_artifact_expired_handler(
     return await exception_handlers.http_exception_handler(
         request,
         fastapi.HTTPException(
-            status_code=status.HTTP_409_CONFLICT,
+            status_code=status.HTTP_404_NOT_FOUND,
             detail={
                 "err_code": "ARTIFACT_EXPIRED",
                 "reason": (
--- a/backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
+++ b/backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
@@ -62,7 +62,7 @@ async def git_instance_api_endpoint_not_found_handler(
     return await exception_handlers.http_exception_handler(
         request,
         fastapi.HTTPException(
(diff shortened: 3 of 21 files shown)


backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -52,7 +52,7 @@ async def get_diagram_metadata(
     except requests.exceptions.HTTPError:
         logger.info("Failed fetching diagram metadata", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_404_NOT_FOUND,
```
MoritzWeber0, line 55: I think `HTTP_422_UNPROCESSABLE_ENTITY` would be better here. We can't say if the resource doesn't exist or if it's another error. The case is more or less the wildcard for all errors.

backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -90,7 +90,7 @@ async def get_diagram(
     except requests.exceptions.HTTPError:
         logger.info("Failed fetching diagram", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_404_NOT_FOUND,
```
MoritzWeber0, line 93: Same as above. I think that `HTTP_422_UNPROCESSABLE_ENTITY` is the better option.

backend/capellacollab/projects/toolmodels/modelbadge/routes.py
```diff
@@ -44,7 +44,7 @@ async def get_model_complexity_badge(
     except (aiohttp.web.HTTPException, requests.exceptions.HTTPError):
         logger.info("Failed fetching model complexity badge", exc_info=True)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_404_NOT_FOUND,
```
MoritzWeber0, line 47: Same as above. I think `HTTP_422_UNPROCESSABLE_ENTITY` is better.

backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
```diff
@@ -62,7 +62,7 @@ async def git_instance_api_endpoint_not_found_handler(
     return await exception_handlers.http_exception_handler(
         request,
         fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
```
MoritzWeber0, line 65: I'd say that `HTTP_400_BAD_REQUEST` is better here. The client is sending a request referencing an "invalid" Git instance. Unprocessable entity should also be fine.

backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
```diff
@@ -116,7 +116,7 @@ async def github_artifact_expired_handler(
     return await exception_handlers.http_exception_handler(
         request,
         fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_409_CONFLICT,
```
MoritzWeber0, line 119: For expired artifacts, `410 Gone` should be the better choice. Or alternatively, a 404 because the resource is not / no longer available. I don't see a conflict here.

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/exceptions.py
```diff
@@ -42,7 +42,7 @@ async def gitlab_project_not_found_handler(
     return await exception_handlers.http_exception_handler(
         request,
         fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_422_UNPROCESSABLE_ENTITY,
```
MoritzWeber0, line 45: 404 should be better for project "not found" (the name states it already).

backend/capellacollab/projects/toolmodels/modelsources/git/handler/exceptions.py
```diff
@@ -22,7 +22,7 @@ async def git_instance_unsupported_handler(
     return await exception_handlers.http_exception_handler(
         request,
         fastapi.HTTPException(
-            status_code=fastapi.status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=fastapi.status.HTTP_422_UNPROCESSABLE_ENTITY,
```
MoritzWeber0, line 25: Bad request when the request is sent for an unsupported Git instance, I'd say.

backend/capellacollab/projects/toolmodels/modelsources/git/handler/exceptions.py
```diff
@@ -39,7 +39,7 @@ async def no_matching_git_instance_handler(
     return await exception_handlers.http_exception_handler(
         request,
         fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_409_CONFLICT,
```
MoritzWeber0, line 42: 404 if there is no matching instance. Conflict would mean that there are two matching instances from my understanding.

backend/capellacollab/projects/toolmodels/restrictions/routes.py
```diff
@@ -48,7 +48,7 @@ def update_restrictions(
 ) -> models.DatabaseToolModelRestrictions:
     if body.allow_pure_variants and not model.tool.integrations.pure_variants:
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_409_CONFLICT,
```
MoritzWeber0, line 51: I think bad request is better here as the client does a request referencing a model with an invalid tool configuration.

backend/capellacollab/sessions/files/routes.py
```diff
@@ -33,7 +33,7 @@ def list_files(
     except Exception:
         log.exception("Loading of files for session %s failed", session.id)
         raise fastapi.HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            status_code=status.HTTP_404_NOT_FOUND,
```
MoritzWeber0, line 36: Here, I'd expect another error code. 404 would mean that the file doesn't exist, but this is not a statement that we can make. We just don't know if the file exists. Here, I'd actually say that internal server error is the correct response.

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
```diff
@@ -120,7 +120,7 @@ def delete_t4c_repository(
         interface.delete_repository(instance, repository.name)
     except (fastapi.HTTPException, requests.RequestException) as e:
         log.error("Repository deletion failed partially", exc_info=True)
-        response.status_code = status.HTTP_207_MULTI_STATUS
+        response.status_code = status.HTTP_202_ACCEPTED
```
MoritzWeber0, line 123: The reason for multi-status is that it failed partially. So it's somewhere between 2xx and 4xx.

Accepted means: 
"The HyperText Transfer Protocol (HTTP) 202 Accepted response status code indicates that the request has been accepted for processing, but the processing has not been completed"

In our case, the request is completed. According to Mozilla, MULTI_STATUS should only be used as part of the WebDAV protocol, so `HTTP_422_UNPROCESSABLE_ENTITY` is maybe an option.

MoritzWeber0 (changes requested):

backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
amolenaar: I would stick with unprocessable entity then. The request is semantically valid.

backend/capellacollab/projects/toolmodels/modelsources/git/exceptions.py
amolenaar: Let's stick with 404 then.

backend/capellacollab/projects/toolmodels/modelsources/git/gitlab/exceptions.py
amolenaar: I gave it a 422, since the resource is not found in GitLab. But 404 will do it for sure.

backend/capellacollab/sessions/files/routes.py
amolenaar: Fair point. I think I would go for 422 then, since it may be a user error and is an error outside of the backend itself.

backend/capellacollab/settings/modelsources/t4c/repositories/routes.py
amolenaar: Then 422 is probably the most appropriate.

MoritzWeber0 (approved):

PR #927, opened 2023-08-11
ci: Add Codecov upload after pytest in GH action

MoritzWeber0:
More information about Codecov: https://about.codecov.io/


amolenaar (approved):
Looks good.

Shouldn't you also add [a shield to the readme](https://shields.io/badges/codecov)?

PR #936, opened 2023-08-14
refactor: Add custom `UnsupportedSessionTypeError`

MoritzWeber0:
Not all tools support all session types.


Paula-Kli (approved):

backend/capellacollab/sessions/exceptions.py
```diff
@@ -0,0 +1,40 @@
...
+) -> fastapi.Response:
+    return await exception_handlers.http_exception_handler(
+        request,
+        fastapi.HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail={
+                "err_code": "SESSION_TYPE_UNSUPPORTED",
+                "reason": (
+                    f"The tool {exc.tool.name} doesn't support the session type '{exc.session_type.value}'"
+                ),
```
Paula-Kli, line 30: this is written generic - so are we going to support other errors with other session types as well?

MoritzWeber0: Yes, I wanted to keep it extendable for the future. For example, if we should add a new session type for some reason.

Issue #833, opened 2023-07-10
Migrate to pydantic v2


dominik003:
Pydantic v2 was released last week and is introduced in fastapi version `0.100.0`. We should go through the [pydantic migration guide](https://docs.pydantic.dev/latest/migration/) and make the necessary changes. There is also a tool that helps with the migration, so using it probably already does everything we need.

Issue #953, opened 2023-08-17
Datetime serializer fails for None object


MoritzWeber0:
```
Traceback (most recent call last):
  File "/tmp/backend/capellacollab/core/logging/__init__.py", line 75, in dispatch
    return await call_next(request)
           ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 84, in call_next
    raise app_exc
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 70, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 108, in __call__
    response = await self.dispatch_func(request, call_next)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/backend/capellacollab/core/logging/__init__.py", line 86, in dispatch
    response: fastapi.Response = await call_next(request)
                                 ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 84, in call_next
    raise app_exc
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 70, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 108, in __call__
    response = await self.dispatch_func(request, call_next)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/.venv/lib/python3.11/site-packages/starlette_prometheus/middleware.py", line 57, in dispatch
    raise e from None
  File "/opt/.venv/lib/python3.11/site-packages/starlette_prometheus/middleware.py", line 53, in dispatch
    response = await call_next(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 84, in call_next
    raise app_exc
  File "/opt/.venv/lib/python3.11/site-packages/starlette/middleware/base.py", line 70, in coro
    await self.app(scope, receive_or_disconnect, send_no_error)
[...]

PR #956, opened 2023-08-21
Migrate to pydantic v2

dominik003:
This PR migrates to pydantic v2. As it has to deal with the validators it also includes a small fix where the `datetime_serializer` failed due to a None value passed to it.
However, currently there are still two **warnings** open for the pydantic migrations

1. UserWarning: Field "model_slug" has conflict with protected namespace "model_"
2. UserWarning: Field "model_badge_status" has conflict with protected namespace "model_"

The reason for this is that the pydantic configuration now works via `model_...` fields and the two fields above obviously are conflicting with the protected namespace. There are three solutions to this problem:

1. Apply the suggestions: `model_config = pydantic.ConfigDict(protected_namespaces=())
`. However, I am not quite sure what would then happen if we have an actual conflict (i.e., whether we would still get an error in this case)
2. Just keep the warning as it is (don't think that this is a good idea but still wanted to put it here)
3. Rename the fields (Since it currently only affects to fields the effort of renaming them is not as big. However, especially for the `model_slug` field, I am not sure what name to choose.

@MoritzWeber0 What are your ideas on the open problem here?

Resolve: #833 #953

### DIFF
--- a/backend/capellacollab/core/authentication/provider/azure/routes.py
+++ b/backend/capellacollab/core/authentication/provider/azure/routes.py
@@ -58,7 +58,7 @@ async def api_get_token(
     auth_data = global_session_data[body.state]
     del global_session_data[body.state]
     token = ad_session().acquire_token_by_auth_code_flow(
-        auth_data, body.dict(), scopes=[]
+        auth_data, body.model_dump(), scopes=[]
     )
     access_token = token["id_token"]
 
--- a/backend/capellacollab/core/authentication/provider/models.py
+++ b/backend/capellacollab/core/authentication/provider/models.py
@@ -12,8 +12,8 @@ class JSONWebKey(pydantic.BaseModel):
     n: str
     e: str
     kid: str
-    x5t: str | None
-    x5c: list[str] | None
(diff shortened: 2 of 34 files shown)


MoritzWeber0:
> 1. Apply the suggestions: `model_config = pydantic.ConfigDict(protected_namespaces=()) `. However, I am not quite sure what would then happen if we have an actual conflict (i.e., whether we would still get an error in this case)

Can you try it out? Add a conflict and check if an error occurs. If there is an error, this would be my preferred solution.

> 2. Just keep the warning as it is (don't think that this is a good idea but still wanted to put it here)

Would be ok as temporary solution, but not for the long-term.

> 3. Rename the fields (Since it currently only affects to fields the effort of renaming them is not as big. However, especially for the `model_slug` field, I am not sure what name to choose.

`model` is also often called `toolmodel` to prevent conflicts in other places of the code. This would be my preferred solution if (1) doesn't work.

MoritzWeber0 (approved):

MoritzWeber0 (approved):

PR #963, opened 2023-08-23
fix: Add asterisc to entry point when editiing the model sources

Paula-Kli:
Apparently the entrypoint field is not a required field in the strict sense since it has two error messages attached. Nevertheless, it confused me that it seems not to be required while at the same time throwing an error when being left empty.

### DIFF
--- a/frontend/src/app/projects/models/model-source/git/manage-git-model/manage-git-model.component.ts
+++ b/frontend/src/app/projects/models/model-source/git/manage-git-model/manage-git-model.component.ts
@@ -146,9 +146,10 @@ export class ManageGitModelComponent implements OnInit, OnDestroy {
       .subscribe((model) => {
         this.modelSlug = model.slug;
         if (model.tool.name === 'Capella') {
-          this.form.controls.entrypoint.addValidators(
-            Validators.pattern(/^$|\.aird$/)
-          );
+          this.form.controls.entrypoint.addValidators([
+            Validators.pattern(/^$|\.aird$/),
+            Validators.required,
+          ]);
         }
       });
 


frontend/src/app/projects/models/model-source/git/manage-git-model/manage-git-model.component.html
dominik003, line 1: Good catch there! However, simply adding the * here is not a proper fix. The star is automatically added to the label in case there is a required validator on the field. Therefore, the missing * indicates here that the field has not a required validator on it.
Furthermore, the entrypoint is not required for all models but just for *Capella* models (i.e., *Papyrus* and *Jupyter* models do not need an entrypoint).
So to properly fix this, you can change the following in the `manage-git-model.component.ts`:

Old:
```ts
this.form.controls.entrypoint.addValidators(
  Validators.pattern(/^$|\.aird$/),
);
```

New:
```ts
this.form.controls.entrypoint.addValidators([
  Validators.pattern(/^$|\.aird$/),
  Validators.required,
]);
```

dominik003 (changes requested):
Good catch, but a provided a more proper fix for this problem as the entrypoint is not required for all of our model types.

frontend/src/app/projects/models/model-source/git/manage-git-model/manage-git-model.component.html
Paula-Kli: ah ok I see - thanks for the explanation and the hint!

dominik003 (approved):

MoritzWeber0 (approved):

PR #964, opened 2023-08-23
feat: Redesign NavBar and Breadcrumbs

### LINKED ISSUE #784: Redesign navigation bar
The breadcrumbs can get really long, so they should get a separate line. In addition, they can have a smaller font size and should be a bit more modern with a small effect on hover.

dominik003:
This PR mainly moves the breadcrumbs into a separate row below the main NavBar and makes them smaller. This has the advantage that even long breadcrumbs (e.g., long model names) will be displayed correctly. In addition, the breadcrumbs now also have a proper hover visualisation and some broken redirects were fixed. Lastly, just a few unnecessary breadcrumbs where removed (e.g., `projects / overview` -> `Projects`) and some breadcrumb names were adjusted.

Resolves #784

### DIFF
--- a/frontend/src/app/app-routing.module.ts
+++ b/frontend/src/app/app-routing.module.ts
@@ -51,16 +51,16 @@ const routes: Routes = [
       {
         path: '',
         component: SessionsComponent,
-        data: { breadcrumb: 'sessions' },
+        data: { breadcrumb: 'Sessions' },
       },
       {
         path: 'session',
         component: SessionComponent,
-        data: { breadcrumb: 'session' },
+        data: { breadcrumb: 'Session Viewer' },
       },
       {
         path: 'projects',
-        data: { breadcrumb: 'projects' },
+        data: { breadcrumb: 'Projects' },
         children: [
           {
             path: '',
@@ -69,14 +69,14 @@ const routes: Routes = [
           },
           {
             path: 'create',
-            data: { breadcrumb: 'create' },
+            data: { breadcrumb: 'Create Project' },
             component: CreateProjectComponent,
           },
         ],
       },
       {
         path: 'project',
-        data: { breadcrumb: 'projects', redirect: '/projects' },
+        data: { breadcrumb: 'Projects', redirect: '/projects' },
         children: [
           {
             path: ':project',
@@ -88,16 +88,13 @@ const routes: Routes = [
             children: [
               {
                 path: '',
-                data: {
-                  breadcrumb: 'overview',
-                  redirect: (data: Data) => `/project/${data.project?.slug}`,
-                },
+                data: { breadcrumb: undefined },
                 component: ProjectDetailsComponent,
               },
               {
                 path: 'metadata',
                 data: {
-                  breadcrumb: 'metadata',
+                  breadcrumb: 'Configure Project',
                   redirect: (data: Data) =>
                     `/project/${data.project?.slug}/metadata`,
                 },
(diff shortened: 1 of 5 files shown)


frontend/src/app/app-routing.module.ts
```diff
@@ -88,74 +88,69 @@ const routes: Routes = [
...
-                      redirect: (data: Data) =>
-                        `/project/${data.project?.slug}`,
-                    },
 
                     children: [
                       {
                         path: 'modelsources',
                         data: {
-                          breadcrumb: 'model sources',
+                          breadcrumb: 'Model Sources',
```
MoritzWeber0, line 124: image

Here, the model slug is missing in the title. It's not clear for the user which model the model sources belong to.

frontend/src/app/app-routing.module.ts
```diff
@@ -208,21 +203,27 @@ const routes: Routes = [
...
+                              redirect: (data: Data) =>
+                                `/project/${data.project?.slug}/model/${data.model?.slug}/git-model/create`,
+                            },
                             component: ManageGitModelComponent,
                           },
                           {
                             path: ':git-model',
                             data: {
                               breadcrumb: (data: Data) =>
                                 data.gitModel?.id || '...',
```
MoritzWeber0, line 224: When clicking the `...`, it redirects to a URL containing undefined, which leads to an error.

frontend/src/app/app-routing.module.ts
```diff
@@ -270,22 +274,21 @@ const routes: Routes = [
       },
       {
         path: 'sessions',
```
MoritzWeber0, line 276: You capitalized most of the breadcrumbs, but not all (for example, `sessions` and `overview`) are still lower case.

frontend/src/app/general/breadcrumbs/breadcrumbs.component.html
```diff
@@ -5,6 +5,8 @@
 
 <ul>
```
MoritzWeber0, line 6: I'd add a `/` after `<ul>`, which looks like: 

image

This indicates that we have an "absolute" path from the root of the platform and not a relative path.

frontend/src/app/general/header/header.component.html
```diff
@@ -3,83 +3,91 @@
...
-      >
-        Profile <mat-icon>account_circle</mat-icon>
-      </button>
+          Profile <mat-icon>account_circle</mat-icon>
+        </button>
+      </div>
     </div>
   </div>
+
+  <div class="primaryText ml-5 select-none text-base">
```
MoritzWeber0, line 90:
```suggestion
  <div class="primaryText ml-5 pb-3 select-none text-base">
```

Add a padding to the bottom, otherwise the header ends directly with the breadcrumb, which looks unevenly spaced.

frontend/src/app/general/breadcrumbs/breadcrumbs.component.html
MoritzWeber0, line 6: On the `<ul>` element add `class="break-all"` to also break words (try it out in the mobile view with project slug `asdfsdafsafat ft srafdasgasdfaq3fdasfsafasfasfasd`. It should wrap everywhere and not only at the spaces.

MoritzWeber0 (changes requested):
Love it, great improvement. Only some comments related to styling.

frontend/src/app/app-routing.module.ts
dominik003: I added the model slug as intermediate breadcrumb forwarding back to the project page.

frontend/src/app/app-routing.module.ts
dominik003: Adjusted the sessions breadcrumb (in line 54) but the overview breadcrumb was already `Session Overview`. This here is only the path which can not be adjusted.

frontend/src/app/app-routing.module.ts
dominik003: Good catch. I changed the redirect to "undefined" at all three occurrences of "..." in case the id is not yet given.

MoritzWeber0 (approved):

Issue #632, opened 2023-02-20
Change button text to `continue` when adding user during project initialisation


dominik003:
At the moment, the button in the second step of user creation `Add team members` always says "Skip", even if you have added at least one user. This is a bit misleading, as the user can't be sure if adding the user actually worked. I would suggest changing the text to *Continue* if a user has been added in this step.

PR #966, opened 2023-08-28
feat: Conditional display of skip or continue during user creation in the project creation

dominik003:
This adjusts the second step of the project creation. Up to now, there was always the "skip" button shown even if a user was added. As this is quite misleading this PR changes it such that the button states "continue" if at least one user was added.

Resolves: #632

### DIFF
--- a/frontend/src/app/projects/create-project/create-project.component.html
+++ b/frontend/src/app/projects/create-project/create-project.component.html
@@ -101,7 +101,13 @@
                 (click)="stepper.next()"
                 data-testId="button-skipAddMembers"
               >
-                Skip
+                <ng-container
+                  *ngIf="
+                    projectUserService.nonAdminProjectUsers$ | async as users
+                  "
+                >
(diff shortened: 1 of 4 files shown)


Paula-Kli (approved):
Looks like a useful enhancement! 👍

frontend/src/app/projects/create-project/create-project.component.html
```diff
@@ -101,7 +101,13 @@ <h3>Project visibility</h3>
                 (click)="stepper.next()"
                 data-testId="button-skipAddMembers"
               >
-                Skip
+                <ng-container
+                  *ngIf="
+                    projectUserService.nonAdminProjectUsers$ | async as users
```
MoritzWeber0, line 106: This is a cool syntax :)

MoritzWeber0 (approved):

MoritzWeber0:
Seems to break the frontend tests. Has to be fixed before it can be merged.

PR #967, opened 2023-08-28
build: Bump frontend versions

Paula-Kli:


dominik003 (approved):
Thanks for updating all versions!

Paula-Kli:
> Thanks for updating all versions!

Unfortunately it is not all versions, as they were conflicting - but at least some are updated and I hope to update the rest in the future

MoritzWeber0 (approved):

PR #968, opened 2023-08-28
build: Add threshold to codecoverage

Paula-Kli:
I talked with @MoritzWeber0 about it that sometimes you only want to refactor and the coverage drops and failes so it might make sense to introduce a threshold.
Unfortunately the codecov action does not implement that at the moment which is why a yml for codecov then is needed where the coverage is quite deeply nested (also see [this](https://github.com/codecov/codecov-action/issues/554) codecov action issue)

Maybe we need to talk about the hight of the threshold

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: CC0-1.0
+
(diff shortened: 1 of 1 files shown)


dominik003 (approved):
I think starting with 2% is a good idea. If we see that in most cases simple refactoring PRs exceed the 2% threshold, we can still increase it.

MoritzWeber0 (approved):

PR #973, opened 2023-08-28
feat: Introduce Fullscreen Mode

### LINKED ISSUE #981: Fullscreen mode for session viewer and training mode
The session viewer  enables users to operate directly within the collaboration manager, eliminating the need to open separate tabs. and it also allows for the simultaneous viewing of two sessions side-by-side. To prioritize the open sessions and maximize the available workspace, the header and footer should be made "invisible," and any extraneous spacing should be removed.

dominik003:
This PR introduces a fullscreen mode to enhance the user experience in the Session Viewer. The current layout utilizes significant space for the navbar and footer, which can be restrictive when using the Session Viewer. The fullscreen mode aims to provide more space by hiding the navbar and footer, allowing users to focus solely on their session.
Resolves #981 

Without fullscreen mode:
image

With fullscreen mode:
image

### DIFF
--- a/frontend/src/app/app.component.html
+++ b/frontend/src/app/app.component.html
@@ -8,14 +8,29 @@
     <app-nav-bar-menu></app-nav-bar-menu>
   </mat-drawer>
   <mat-drawer-content class="overall-wrapper">
-    <div class="navBar">
+    <div
+      class="navBar"
+      *ngIf="(fullscreenService.isFullscreen$ | async) === false"
+    >
       <app-header *ngIf="pageLayoutService.showHeader"></app-header>
     </div>
-    <div class="wrapper">
-      <app-notice *ngIf="pageLayoutService.showNotice"></app-notice>
+    <div
+      [ngClass]="{
+        wrapper: (fullscreenService.isFullscreen$ | async) === false
+      }"
+    >
+      <app-notice
+        *ngIf="
+          (fullscreenService.isFullscreen$ | async) === false &&
+          pageLayoutService.showNotice
+        "
+      ></app-notice>
       <router-outlet></router-outlet>
     </div>
-    <div class="footer">
+    <div
+      class="footer"
+      *ngIf="(fullscreenService.isFullscreen$ | async) === false"
+    >
       <app-version id="left"></app-version>
       <div id="middle">
         <app-footer
--- a/frontend/src/app/sessions/session/session.component.css
+++ b/frontend/src/app/sessions/session/session.component.css
@@ -7,10 +7,6 @@
   height: calc(100vh - 2vh - 65px - 110px);
 }
 
-.iframe-container {
-  position: relative;
-}
-
 .iframe-overlay {
   position: absolute;
   top: 0;
@@ -19,9 +15,4 @@
   bottom: 0;
 
   z-index: 40;
-}
-
-iframe {
-  width: 100%;
-  height: 100%;
 }
--- a/frontend/src/app/sessions/session/session.component.html
+++ b/frontend/src/app/sessions/session/session.component.html
@@ -11,6 +11,7 @@
   <div *ngIf="cachedSessions && !cachedSessions.length">
     No sessions found. Please create a session in the 'Sessions' tab first.
   </div>
+
   <div *ngIf="cachedSessions === undefined">
     <ngx-skeleton-loader
       *ngFor="let _ of [0, 1, 2, 3]"
@@ -24,6 +25,7 @@
       }"
     ></ngx-skeleton-loader>
   </div>
+
   <div class="flex flex-col" *ngIf="cachedSessions && cachedSessions.length">
     <mat-checkbox
       class="w-fit"
@@ -40,6 +42,7 @@
       >, created at {{ session.created_at | date: "EE, dd MMM y HH:mm:ss" }}
     </mat-checkbox>
   </div>
+
   <div class="border-2 p-2 mt-2 rounded w-fit">
     Please note: We're not technically preventing that you select two
     Eclipse-based sessions (for example two Capella sessions), <br />but the
@@ -59,10 +62,15 @@
   </div>
 </div>
 
-<div class="w-full height flex gap-2" *ngIf="selectedSessions.length">
+<div class="flex gap-0.5" *ngIf="selectedSessions.length">
   <div
-    class="w-full height active:z-30"
-    [ngClass]="session.focused ? 'z-20' : 'z-10'"
+    class="flex flex-col w-full active:z-30"
+    [ngClass]="{
+      'z-20': session.focused,
+      'z-10': !session.focused,
+      height: (fullscreenService.isFullscreen$ | async) === false,
+      'h-screen': fullscreenService.isFullscreen$ | async
+    }"
     (click)="focusSession(session)"
     *ngFor="let session of selectedSessions"
     cdkDrag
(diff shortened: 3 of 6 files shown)


frontend/src/app/app.component.html
```diff
@@ -8,14 +8,29 @@
     <app-nav-bar-menu></app-nav-bar-menu>
   </mat-drawer>
   <mat-drawer-content class="overall-wrapper">
-    <div class="navBar">
+    <div
+      class="navBar"
+      *ngIf="(fullscreenService.isFullscreen$ | async) === false"
```
amolenaar, line 13: Isn't this better than explicitly checking for `false`?

```suggestion
      *ngIf="!(fullscreenService.isFullscreen$ | async)"
```

frontend/src/app/sessions/session/session.component.css
```diff
@@ -7,6 +7,10 @@
   height: calc(100vh - 2vh - 65px - 110px);
 }
 
+.heigh-fullscreen {
+  height: calc(100vh - 2.6vh);
```
amolenaar, line 11: Where does the `2.6vh` come from?

frontend/src/app/app.component.html
dominik003: Yes in theory I would agree with you (and I did it like this before) but eslint had something against using `!` together with an async pipe. Didn't look up the actual reason behind this but now I was curious and [here](http://codelyzer.com/rules/template-no-negated-async/) is the reason behind that.

frontend/src/app/sessions/session/session.component.css
dominik003: It was more or less testing. Using a value of `2vh` led to a scrolling bar on the side while a value of `3vh` resulted in some "unnecessary" white area below the sessions

frontend/src/app/sessions/session/session.component.html
```diff
@@ -70,7 +73,7 @@
     (cdkDragEnded)="dragStop()"
   >
     <div
-      class="flex gap-2 rounded-t p-2 cursor-grab active:cursor-grabbing items-center justify-between"
+      class="flex gap-2 rounded-t cursor-grab active:cursor-grabbing items-center justify-between"
```
MoritzWeber0, line 76:
```suggestion
      class="flex gap-2 rounded-t p-2 cursor-grab active:cursor-grabbing items-center justify-between"
```

Was this intentional? Without the padding it looks very compressed and less modern.

frontend/src/app/sessions/session/session.component.html
```diff
@@ -99,7 +102,12 @@
       <div [ngClass]="draggingActive ? 'iframe-overlay' : ''"></div>
       <iframe
         [id]="'session-' + session.id"
-        class="w-full height"
+        [ngClass]="{
+          'w-full': true,
+          'heigh-fullscreen': fullscreenService.isFullscreen$ | async,
+          height: (fullscreenService.isFullscreen$ | async) === false
+        }"
```
MoritzWeber0, line 109:
```suggestion
```

1. `w-full` is defined twice (here and in the `class` attribute).
2. Neither `height` nor `heigh-fullscreen` are needed here. The height should be handled by the element which contains the session title bar and the iframe.

frontend/src/app/sessions/session/session.component.html
```diff
@@ -99,7 +102,12 @@
...
+        class="w-full height max-w-full max-h-full"
```
MoritzWeber0, line 110:
```suggestion
        class="w-full"
```

If the `width` / `height` is set to a fixed value, there shouldn't be the need of setting a max value. Height should be handled in the parent container (which contains the session title bar and the iframe). To pass the height down, set `h-full` on the `iframe-container` element.

frontend/src/app/sessions/session/session.component.css
MoritzWeber0: You applied the class to the wrong element. If you apply it to the div element, which contains iframe & header, you can use `h-full`. You did just apply it to the iframe and removed the approximate size of the header. This isn't a good style as the header size has untracked dependencies in other CSS classes like this one.

frontend/src/app/services/fullscreen.service.ts
```diff
@@ -0,0 +1,19 @@
+/*
```
MoritzWeber0, line 1: New services should never go into the global `services` directory. They belong to context modules. The existing ones are just legacy ones, which never were moved. See https://github.com/DSD-DBS/capella-collab-manager/issues/320

frontend/src/app/sessions/session/session.component.css
```diff
@@ -7,6 +7,10 @@
...
+}
+
 .iframe-container {
   position: relative;
 }
```
MoritzWeber0, line 16: As far as I can see, this has no effect. We should be able to remove it from CSS & HTML.

MoritzWeber0 (changes requested):

frontend/src/app/services/fullscreen.service.ts
MoritzWeber0, line 1: The Guacamole sessions are not resized after changing to the full-screen mode. Large "black" borders appear in the session element: 

image

This is similar behaviour like resizing the browser window, please have a look at: 

```py
  @HostListener('window:resize', ['$event'])
  onResize() {
    window.clearTimeout(this.debounceTimer);

    this.debounceTimer = window.setTimeout(() => {
      Array.from(document.getElementsByTagName('iframe')).forEach((iframe) => {
        const session = this.selectedSessions.find(
          (session) => 'session-' + session.id === iframe.id
        );

        if (session?.reloadToResize) {
          this.reloadIFrame(iframe);
        }
      });
    }, 250);
  }

```

frontend/src/app/sessions/session/session.component.html
dominik003: The main objective is to maximize the available workspace in the session viewer, which is why the padding was initially removed. However, I agree that some level of padding is beneficial, so I've adjusted it to p-1.

frontend/src/app/sessions/session/session.component.html
dominik003: I completely reworked the styling and tried to remove as much unnecessary css as possible. I think its best to just have another look over it.

frontend/src/app/sessions/session/session.component.html
dominik003: I completely reworked the styling and tried to remove as much unnecessary css as possible. I think its best to just have another look over it.

frontend/src/app/services/fullscreen.service.ts
dominik003: I wasn't sure where to place it but I'll move it to the session scope

frontend/src/app/sessions/session/session.component.css
dominik003: It had an effect on my side but I changed it to a tailwind class so removed it here.

frontend/src/app/services/fullscreen.service.ts
dominik003: I only tested it with Jupyter sessions and forgot about resizing. Fortunately, including it was pretty simple because I only need to call `onResize()` when the fullscreen value changes.

frontend/src/app/sessions/session/session.component.html
```diff
@@ -59,18 +62,23 @@
...
+    }"
     (click)="focusSession(session)"
     *ngFor="let session of selectedSessions"
     cdkDrag
     (cdkDragStarted)="dragStart()"
     (cdkDragEnded)="dragStop()"
   >
     <div
-      class="flex gap-2 rounded-t p-2 cursor-grab active:cursor-grabbing items-center justify-between"
+      class="flex gap-2 p-1 rounded-tcursor-grab active:cursor-grabbing items-center justify-between"
```
MoritzWeber0:
```suggestion
      class="flex gap-2 p-1 rounded-t cursor-grab active:cursor-grabbing items-center justify-between"
```

MoritzWeber0 (approved):

PR #978, opened 2023-08-30
feat: Add possibility to insert colorful background when retrieving diagram

### LINKED ISSUE #950: Postprocess diagram cache SVGs on demand
In py-capellambse, a new feature was added to include a white background to diagrams: https://github.com/DSD-DBS/py-capellambse/pull/323

The `background` flag is passed to the `update_diagram_cache` method. The setting applies to all generated diagrams. 

That's where the Collaboration Manager API could "jump in": The post processing rules can be tuned via query parameters in our `diagrams` API. For example, a request to `/api/v1/projects/xy/models/xy/diagrams/uuid?white_background=true` could return the diagram with a white background, which is added on demand.

The feature could be extended with more SVG post processing features in the future.

Paula-Kli:
There might be occasions when it is handy to easily get a colorful/ white or transparent background to diagrams.

The remaining question is how to make that option available to users? should there be a separate button? Is a dialog popping up when downloading?

closes #950


Paula-Kli:
needs further discussion

Paula-Kli:
there is no usecase for that feature and route

PR #1007, opened 2023-09-11
Update pre-commit black

Wuestengecko:
Update to the latest version of `black`, and switch to the official pre-commit mirror, which runs about 2x faster.


MoritzWeber0 (approved):
Merci 👍

PR #1015, opened 2023-09-12
fix: Add t4c and pure_variants to ToolIntegration during migration

MoritzWeber0:


MoritzWeber0:
Can't reproduce the initial error, so a fix is not needed.

PR #1030, opened 2023-09-19
feat: Enable mypy in pre-commit

### LINKED ISSUE #245: Add `mypy` pre-commit hook
Enforce the use of `mypy` with pre-commit hooks.

dominik003:
Resolves #245

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -44,6 +44,18 @@ repos:
       - id: isort
         entry: bash -c "cd backend && isort ."
         types: [python]
+  - repo: https://github.com/pre-commit/mirrors-mypy
+    rev: v1.5.1
+    hooks:
+      - id: mypy
+        types_or: [python, spec]
+        files: "^backend/capellacollab"
+        exclude: "^backend/capellacollab/alembic/"
+        args: [--config-file=./backend/pyproject.toml]
+        additional_dependencies:
+          - fastapi
+          - pydantic
+          - sqlalchemy
   - repo: local
     hooks:
       - id: pylint
--- a/backend/capellacollab/core/authentication/provider/azure/routes.py
+++ b/backend/capellacollab/core/authentication/provider/azure/routes.py
@@ -62,7 +62,10 @@ async def api_get_token(
     )
     access_token = token["id_token"]
 
-    username = get_username(JWTBearer().validate_token(access_token))
+    validated_token = JWTBearer().validate_token(access_token)
+    assert validated_token
+
+    username = get_username(validated_token)
 
     if user := users_crud.get_user_by_name(db, username):
         users_crud.update_last_login(db, user)
--- a/backend/capellacollab/core/authentication/provider/oauth/keystore.py
+++ b/backend/capellacollab/core/authentication/provider/oauth/keystore.py
@@ -14,6 +14,8 @@ import requests
 from capellacollab.config import config
 from capellacollab.core.authentication.provider import models
 
+from .. import models as provider_models
+
 log = logging.getLogger(__name__)
 cfg = config["authentication"]["oauth"]
 
@@ -35,7 +37,7 @@ class _KeyStore:
         self.get_jwks_uri = get_jwks_uri
         self.jwks_uri = ""
         self.algorithms = algorithms
-        self.public_keys: dict[t.Any, t.Any] = {}
+        self.public_keys: dict[str, provider_models.JSONWebKey] = {}
         self.key_refresh_interval = key_refresh_interval
         self.public_keys_last_refreshed: float = 0
 
@@ -62,7 +64,7 @@ class _KeyStore:
 
     def key_for_token(
         self, token: str, *, in_retry: int = 0
-    ) -> dict[str, t.Any]:
+    ) -> provider_models.JSONWebKey:
         # Before we do anything, the validation keys may need to be refreshed.
         # If so, refresh them.
         if self.keys_need_refresh():
--- a/backend/capellacollab/core/authentication/provider/oauth/routes.py
+++ b/backend/capellacollab/core/authentication/provider/oauth/routes.py
@@ -30,8 +30,12 @@ async def api_get_token(
     body: TokenRequest, db: orm.Session = fastapi.Depends(database.get_db)
 ):
     token = get_token(body.code)
+    access_token = token["id_token"]
 
-    username = get_username(JWTBearer().validate_token(token["access_token"]))
+    validated_token = JWTBearer().validate_token(access_token)
+    assert validated_token
+
+    username = get_username(validated_token)
 
     if user := users_crud.get_user_by_name(db, username):
         users_crud.update_last_login(db, user)
--- a/backend/capellacollab/core/database/__init__.py
+++ b/backend/capellacollab/core/database/__init__.py
@@ -1,6 +1,7 @@
 # SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
 # SPDX-License-Identifier: Apache-2.0
 
+import typing as t
 
 import pydantic
 import sqlalchemy as sa
@@ -23,7 +24,7 @@ class Base(orm.DeclarativeBase):
 from . import models  # isort:skip # pylint: disable=unused-import
 
 
-def get_db() -> orm.Session:
+def get_db() -> t.Iterator[orm.Session]:
     with SessionLocal() as session:
         yield session
 
--- a/backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
+++ b/backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
@@ -61,7 +61,7 @@ class GithubHandler(handler.GitHandler):
         self,
         project_id: str,
         trusted_file_path: str,
-        revision: str | None = None,
+        revision: str,
         headers: dict[str, str] | None = None,
     ) -> requests.Response:
         return requests.get(
@@ -179,7 +179,9 @@ class GithubHandler(handler.GitHandler):
         ):
             raise git_exceptions.GitPipelineJobFailedError(job_name)
 
-        return None
+        raise git_exceptions.GitPipelineJobUnknownStateError(
+            job_name, matched_jobs[0]["conclusion"]
+        )
 
     def __get_latest_artifact_metadata(self, project_id: str, job_id: str):
         response = requests.get(
--- a/backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
+++ b/backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
@@ -7,6 +7,7 @@ import json
 import typing as t
 
 import requests
+from capellambse import diagram_cache
 
 import capellacollab.projects.toolmodels.modelsources.git.models as git_models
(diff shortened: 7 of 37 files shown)


backend/capellacollab/core/authentication/provider/oauth/keystore.py
```diff
@@ -14,6 +14,8 @@
 from capellacollab.config import config
 from capellacollab.core.authentication.provider import models
 
+from .. import models as provider_models
+
```
Paula-Kli, line 18: I think we talked about that already and I forgot your answer but I thought relative imports make only sense in the same package which is why I am curious why did you decided to use it here :D 
[Google says no relative imports even though we are still using them](https://google.github.io/styleguide/pyguide.html#:~:text=Do%20not%20use%20relative%20names%20in%20imports.%20Even%20if%20the%20module%20is%20in%20the%20same%20package%2C%20use%20the%20full%20package%20name.%20This%20helps%20prevent%20unintentionally%20importing%20a%20package%20twice.)

backend/capellacollab/core/authentication/provider/oauth/routes.py
```diff
@@ -31,7 +31,7 @@ async def api_get_token(
 ):
     token = get_token(body.code)
 
-    username = get_username(JWTBearer().validate_token(token["access_token"]))
+    username = get_username(JWTBearer().validate_token(token["access_token"]))  # type: ignore
```
Paula-Kli, line 34: same here as before

Paula-Kli (changes requested):
Thank you for looking into it and resolving many of the issues!

backend/capellacollab/core/authentication/provider/oauth/keystore.py
dominik003: Yes, we discussed this topic in our planning meeting on the 12.06 and decided there to limit the scope of relative imports to up to one level above the current one. This is also currently written in our development documentation [code-style/imports](https://dsd-dbs.github.io/capella-collab-manager/code-style/#imports).
Personally, I at least would want to keep relative imports at the same level but I think we should then briefly discuss this topic in the next planning meeting.

backend/capellacollab/core/authentication/provider/azure/routes.py
```diff
@@ -62,7 +62,7 @@ async def api_get_token(
     )
     access_token = token["id_token"]
 
-    username = get_username(JWTBearer().validate_token(access_token))
+    username = get_username(JWTBearer().validate_token(access_token))  # type: ignore
```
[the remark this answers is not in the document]
dominik003: In theory, we could also just do something like this:

```python
validated_token = JWTBearer().validate_token(access_token)
assert validated_token

username = get_username(validated_token)
```
As JWTBearer has by default auto error on, this should be correct (i.e., either an error is thrown in the `validate_token` call or we got back the validated token.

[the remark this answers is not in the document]
Paula-Kli: sounds like a good idea to me 👍  I do like it!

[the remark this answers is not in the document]
dominik003: Nice, applied it to both places.

.pre-commit-config.yaml
```diff
@@ -54,6 +54,20 @@ repos:
...
+        language: python
+        types: [python]
+        additional_dependencies:
+          - mypy
+          - fastapi
+          - pydantic
+          - sqlalchemy
+        files: "^backend/capellacollab"
+        exclude: "^backend/capellacollab/alembic/"
+        verbose: true
```
amolenaar, line 70: Why not use the hook provided by pre-commit?

```suggestion
  - repo: https://github.com/pre-commit/mirrors-mypy
    rev: v1.5.1
    hooks:
      - id: mypy
        types_or: [python, spec]
        files: "^backend/capellacollab"
        exclude: "^backend/capellacollab/alembic/"
        args: [--config-file=./backend/pyproject.toml]
        additional_dependencies:
         - fastapi
         - pydantic
         - sqlalchemy
```

I expect then you can drop the additional dependencies.

backend/capellacollab/core/database/__init__.py
```diff
@@ -23,7 +24,7 @@ class Base(orm.DeclarativeBase):
 from . import models  # isort:skip # pylint: disable=unused-import
 
 
-def get_db() -> orm.Session:
+def get_db() -> t.Generator[orm.Session, None, None]:
```
amolenaar, line 27: Could also use `t.Iterator` for generators that you cannot send anything to:

```suggestion
def get_db() -> t.Iterator[orm.Session]:
```

https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -179,7 +179,9 @@ def __get_latest_successful_job(self, jobs: list, job_name: str) -> dict:
         ):
             raise git_exceptions.GitPipelineJobFailedError(job_name)
 
-        return None
+        raise git_exceptions.GitPipelineJobUnknownStateError(
+            job_name, matched_jobs[0]["conclusion"]
+        )
```
amolenaar, line 184: Why not make the return type `requests.Response | None`?

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
```diff
@@ -76,7 +76,7 @@ async def get_file_from_repository_or_artifacts_as_json(
         trusted_file_path: str,
         job_name: str,
         revision: str | None = None,
-    ) -> tuple[datetime.datetime, dict[str, t.Any]]:
+    ) -> tuple[datetime.datetime, list[dict[str, str | bool]]]:
```
amolenaar, line 79: This was an error in the type annotation? (it's not obvious from the code)

backend/capellacollab/sessions/files/routes.py
```diff
@@ -63,11 +63,13 @@ def upload_files(
 
         for file in files:
             file.file.seek(0)
-            file.filename = file.filename.replace(" ", "_")
-            tar.addfile(
-                tar.gettarinfo(arcname=file.filename, fileobj=file.file),
-                fileobj=file.file,
-            )
+
+            if file.filename:
```
amolenaar, line 67: If `file.filename` _should_ me set, you can also add an assert, instead of an if statement.

E.g.

`assert file.filename`

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -74,14 +74,20 @@ def post_session_creation_hook(
...
+        host = (
+            self._jupyter_public_uri.hostname
+            if self._jupyter_public_uri.hostname
+            else ""
+        )
+
         operator.create_public_route(
             session_id=session_id,
-            host=self._jupyter_public_uri.hostname,
+            host=host,
```
amolenaar, line 85:
```suggestion
            host=self._jupyter_public_uri.hostname or "",
```

And remove the `host` variable assignment above,

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -65,7 +65,7 @@ def configuration_hook(
 
         volumes = self._get_project_share_volume_mounts(db, token, tool)
 
-        return environment, volumes, []
+        return environment, volumes, []  # type: ignore
```
amolenaar, line 68: I suggest to be specific with what rule to ignore:

```suggestion
        return environment, volumes, []  # type: ignore[return-value]
```

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -41,7 +41,7 @@ def __init__(self):
         )
         self._general_conf = config["general"]
 
-    def configuration_hook(
+    def configuration_hook(  # type: ignore
```
amolenaar, line 44: I suggest to be specific with what rule to ignore:

```suggestion
    def configuration_hook(  # type: ignore[override]
```

amolenaar (changes requested):
Nice improvement.

I have some remarks. Nothing serious.

It looks like you did a fair bit of boyscouting as part of this PR.

In some places it's more difficult to assess the impact of certain changes due to behavioral changes.

.pre-commit-config.yaml
dominik003: I tried this as the very first thing, but switched to the local version as it didn't work properly. However, it seems it was just missing the extra dependencies to work properly (quite annoying to have them, but apparently they can't take the dependencies from the toml).
So switched to the mirror but kept the `additional_dependencies`.

backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
dominik003: We have to deal with the case that no previous condition was successful somewhere, so I thought here would be the best place (i.e., when possibly returning `None` we would also have to either raise or return `None` in the `get_last_job_run_id_for_git_model` function and so on).

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
dominik003: Yes. Its already some time ago that I did this PR but I think the type annotation here corresponds to json a file that look like this [coffee machine index.json](https://github.com/DSD-DBS/coffee-machine/blob/diagram-cache/main/diagram_cache/index.json).

backend/capellacollab/sessions/files/routes.py
dominik003: Perfect, I prefer the `assert` you suggested, reducing the nesting here.

backend/capellacollab/sessions/hooks/jupyter.py
dominik003: Thats way nice!

dominik003:
> Nice improvement.
> 
> I have some remarks. Nothing serious.
> 
> It looks like you did a fair bit of boyscouting as part of this PR.
> 
> In some places it's more difficult to assess the impact of certain changes due to behavioral changes.

Thanks for the review and all of your comments should be addressed (feel free to close or comment on the remaining open ones).
Its some time ago I did this PR but I think that all changes I included where necessary to handle mypy errors and I tried to properly fix each one.

backend/capellacollab/projects/toolmodels/modelsources/git/handler/handler.py
MoritzWeber0: A typed dict would be even better.

dominik003: Certainly, but this first requires https://github.com/DSD-DBS/py-capellambse/pull/342 to be merged

MoritzWeber0 (approved):
Already has two reviews, therefore I won't review it again.

backend/pyproject.toml
```diff
@@ -109,7 +111,26 @@ allow_untyped_defs = true
 [[tool.mypy.overrides]]
 # Untyped third party libraries
 module = [
-  # ...
+  "capellambse.*",
```
MoritzWeber0, line 114: `capellambse` should have mypy support.

PR #1032, opened 2023-09-20
feat: Improve diagram cache margins on mobile phones

MoritzWeber0:
The diagram overview and diagram viewer makes now full use of the available space on mobile phones.
As the previous way to close the diagram by just pressing next to it it not available anymore (there are no margins),
there is a new sticky close button on mobile phones.

The UI on devices with wider screens (>=1024px) doesn't change.

Thanks to @vik378 for proposing this improvement. 

A screenshot of the new UI (grey area is outside of the device screen):

image


dominik003 (approved):
LGTM! Nice improvement

PR #1039, opened 2023-09-22
Workspace backup CLI tool

amolenaar:
A little tool to back up workspaces.

```bash
python -m capellacollab.cli ws volumes --namespace collab-sessions

python -m capellacollab.cli ws ls  persistent-session-admin --namespace collab-sessions

python -m capellacollab.cli ws backup persistent-session-admin --namespace collab-sessions

python -m capellacollab.cli ws restore persistent-session-admin persistent-session-admin-backup.tar.gz --namespace collab-sessions

```

### DIFF
--- a/backend/capellacollab/cli/ws.py
+++ b/backend/capellacollab/cli/ws.py
@@ -64,6 +64,7 @@ def restore(
     namespace: str = None,
     access_modes: str = "ReadWriteMany",
     storage_class_name: str = "persistent-sessions-csi",
+    user_id: str = None,
 ):
     config.load_kube_config()
     v1 = client.CoreV1Api()
@@ -79,6 +80,13 @@ def restore(
 
         with tarfile.open("rb") as infile:
             stream_tar_to_pod(pod_name, namespace, infile, v1)
+
+        adjust_directory_permissions(
+            pod_name,
+            namespace,
+            v1,
+            user_id,
+        )
 
 
 @contextlib.contextmanager
@@ -174,6 +182,39 @@ def create_persistent_volume(
             print(f"Using existing volume {name}")
             return
         raise
+
+
+def adjust_directory_permissions(
+    pod_name: str,
+    namespace: str,
+    v1: client.CoreV1Api,
+    user_id: str,
+    directory: str = "/workspace",
+):
+    resp = stream.stream(
+        v1.connect_get_namespaced_pod_exec,
+        pod_name,
+        namespace,
+        command=[
+            "chown",
+            "-R",
+            f"{user_id}",
+            directory,
+        ],
+        stderr=True,
+        stdin=False,
+        stdout=True,
+        tty=False,
+        _preload_content=False,
+    )
+
+    # Output the result
+    while resp.is_open():
+        resp.update(timeout=1)
+        if resp.peek_stdout():
+            print(f"STDOUT: {resp.read_stdout()}")
+        if resp.peek_stderr():
+            print(f"STDERR: {resp.read_stderr()}")
 
 
 def get_current_namespace():
--- a/backend/capellacollab/cli/ws.py
+++ b/backend/capellacollab/cli/ws.py
@@ -58,9 +58,19 @@ def backup(volume_name: str, namespace: str = None, out: Path = None):
 
 
 @app.command()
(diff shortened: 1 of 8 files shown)


MoritzWeber0:
Thanks for the useful CLI, it works very well for most workspaces.

There is just one small issue with too long PVCs:
image

Also, it would be good to have a `--all` flag to download all workspaces at once. We solved it by a simple bash script: 
```bash
for workspace in  $(python -m capellacollab.cli ws volumes --namespace t4c-sessions)
do
    echo "Downloading workspace $workspace..."
    python -m capellacollab.cli ws backup $workspace --namespace t4c-sessions
    echo "Finished downloading workspace $workspace..."
done;
```

Also, we need a bit of documentation: 
- A documentation for the CLI itself
- A documentation mentioning the complete backend & restore procedure for the Collaboration Manager data.

backend/capellacollab/cli/ws.py
```diff
@@ -0,0 +1,355 @@
...
+
+@contextlib.contextmanager
+def pod_for_volume(
+    volume_name: str,
+    namespace: str,
+    mount_path: str,
+    v1: client.CoreV1Api,
+    read_only=True,
+):
+    name = f"ws-download-{volume_name}-{uuid.uuid1()}"[:63]
```
dominik003, line 137:
```suggestion
    name = f"ws-download-{uuid.uuid1()}"
```
I am not sure whether you investigated this problem further but when we used the cli, I think we run into some problems when there was a `-` at the last position of the name. So probably just using this here is sufficient.

backend/capellacollab/cli/ws.py
```diff
@@ -0,0 +1,355 @@
...
+            volumes=volumes_,
+            restart_policy="Never",
+        ),
+    )
+
+    v1.create_namespaced_pod(namespace, pod)
+
+    while not is_pod_ready(name, namespace, v1):
+        print("Waiting for pod to come online...")
+        time.sleep(2)
```
dominik003, line 179: I am not sure whether we should include a timeout here. Obviously, the user can just cancel the command but there are certainly cases where either the users forgets about the command (when there is a long runtime) or is not sure whether he can actually cancel it due to a problem. I would probably suggest a timeout of 3-5 minutes.
What do you think about this?

backend/capellacollab/cli/ws.py
amolenaar: I wanted to have a container name that I can relate to the volume it's supposed to work with. Hence the volume name.

backend/capellacollab/cli/ws.py
```diff
@@ -0,0 +1,355 @@
...
+
+from __future__ import annotations
+
+import contextlib
+import select
+import sys
+import time
+import uuid
+from pathlib import Path
+from typing import Annotated
```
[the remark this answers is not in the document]
amolenaar: fixed

backend/capellacollab/cli/ws.py
```diff
@@ -0,0 +1,355 @@
...
+    "-n",
+    help="Kubernetes namespace",
+    default_factory=get_current_namespace,
+)
+
+
+@app.command()
+def volumes(namespace: Annotated[str, NamespaceOption]):
+    """List all Persistent Volume Claims in a kubernetes namespace."""
+    core_api = client.CoreV1Api()
```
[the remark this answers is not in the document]
amolenaar: I wanted to avoid global state. What I actually wanted is an injected dependency, such as we have in FastAPI. However, I did not investigate that yet.

Anyway, I would keep the commands as isolated as possible.

backend/capellacollab/cli/ws.py
amolenaar: Yup, a little timeout would be nice.

backend/capellacollab/cli/ws.py
MoritzWeber0: I'd propose to use the original volume name as label of the deployment. We can still relate where it comes from and the issue mentioned by @dominik003  would be fixed.

amolenaar: Changed to use only the volume name. I kept the name variable, though, for readability.

dominik003 (approved):

PR #1054, opened 2023-09-27
docs: Add documentation on how to use the diff/merge tool

### LINKED ISSUE #979: Add documentation to resolve merge conflicts
We have an internal documentation about how to resolve Git merge conflicts in Capella from @micha91. 

- [x] Anonymize the existing document and add it to the user documentation

Paula-Kli:
closes #979


docs/user/docs/getting-started/working-with-git.md
```diff
@@ -110,13 +110,136 @@ there. Having done that you can click "Push".
 Now your changes are saved at the remote repository where they are not lost and
 your collegues can see them.
 
+## Diff and Merge Models
+
+<!-- prettier-ignore -->
+!!! warning
+   If you have to merge due to a merge conflict do not commit unless all merges highlighted are solved.
```
amolenaar, line 117:
```suggestion
   If you have to merge due to a merge conflict do not commit unless all conflicts highlighted are solved.
```

docs/user/docs/getting-started/working-with-git.md
```diff
@@ -110,13 +110,136 @@ there. Having done that you can click "Push".
...
+click on one of the files and select "Compare with" > "Each other as model".
+
+![Open compare View](working-with-git/start-compare-view.png)
+
+In the image you can see the
+[coffee machine model](https://github.com/DSD-DBS/coffee-machine) imported
+twice with different changes made to each model in order to compare them.
+
+In the View "Define the comparison operation to perform" it is important to
+choose the **correct** order of which model is the and which one is the target.
```
amolenaar, line 132:
```suggestion
choose the **correct** order of which model is the source and which one is the target.
```

docs/user/docs/getting-started/working-with-git.md
```diff
@@ -110,13 +110,136 @@ there. Having done that you can click "Push".
...
+In that case we are taking the changes in the source to apply them to the
+target and are going to change the target model.
```
amolenaar, line 134:
```suggestion
target, and are going to change the target model.
```

docs/user/docs/getting-started/working-with-git.md
```diff
@@ -110,13 +110,136 @@ there. Having done that you can click "Push".
...
+between elements. Thus, click the now appeared tab "Matching".
+
+Usually the UUIDs of the objects in your model should be stable. So ticking the
+Absolute criteria "Technical IDs" and "Property IDs" should be sufficient.
+Especially for if you know that there could be changes in diagrams you might
+also want to tick "Diagrams: Match shapes according to represented elements"
+and "Diagrams: Match remaining shapes according to type" as in Diagrams IDs
+might change just by opening them. If you are unsure whether your IDs are
+stable the option "Capella: Match equivalent technical elements" also worked in
+the past. If nothing else helps choose "Names" and click "Use labels as names"
```
amolenaar, line 160: The sentense "If you ... worked in the past" is a bit weird. Maybe it's something for an info block. It may work for a user, but he has to experiment.

docs/user/docs/getting-started/working-with-git.md
```diff
@@ -110,13 +110,136 @@ there. Having done that you can click "Push".
...
+
+In the View "Define the comparison operation to perform" it is important to
+choose the **correct** order of which model is the and which one is the target.
+In that case we are taking the changes in the source to apply them to the
+target and are going to change the target model.
+
+<!-- prettier-ignore -->
+!!! warning
+
+Make sure to select the correct direction of merging in order to modify the
```
amolenaar, line 139: Should this be indented?

docs/user/docs/getting-started/working-with-git.md
```diff
@@ -110,13 +110,136 @@ there. Having done that you can click "Push".
...
+might change just by opening them. If you are unsure whether your IDs are
+stable the option "Capella: Match equivalent technical elements" also worked in
+the past. If nothing else helps choose "Names" and click "Use labels as names"
+to match but labels easily change.
+
+![Configure Matching Strategie](working-with-git/configure-merge-matching.png)
+
+Afterwards, click "Ok" and "Finish".
+
+Then you get into the merge view. Market with the lock you can see which model
```
amolenaar, line 167:
```suggestion
Then you get into the merge view. Marked with the lock you can see which model
```

docs/user/docs/getting-started/working-with-git.md
```diff
@@ -110,13 +110,136 @@ there. Having done that you can click "Push".
...
+### Merging with the Merge Tool
+
+In the above description we made sure to decide which model is our target and
+is going to be modified while the changes would be taken from the source model.
+In the case of a merge conflict we have changes in both models at the same
+position and it is not possible to decide which change to keep and which to
+delete while possibly both are reasonable changes. Thus we as users have to
+decide that and both changes have the same right to be present in the model. In
+order to not just apply changes you can decide from which sides changes can be
+taken.
```
amolenaar, line 224: This paragraph doesn't read fluently. This is about the case where the tooling cannot automatically merge and a user has to do it by hand, right?

amolenaar (changes requested):
One thing I miss is how to start overm when your merge efforts mess up the model.
Is that something that should be part of this doc?

Is there some official docs on the diff/merge tool we can refer to?

docs/user/docs/getting-started/working-with-git.md
Paula-Kli: Do you think that is an option which we should leave out in that case?

MoritzWeber0: I'd remove the sentence completely.

docs/user/docs/getting-started/working-with-git.md
MoritzWeber0: I agree with Arjan that we should rewrite this paragraph. 

Especially the following two parts:

> In the case of a merge conflict we have changes in both models at the same
position and it is not possible to decide which change to keep and which to
delete while possibly both are reasonable changes.

It's possible to decide which changes to keep and which to discard, but the user has to do it. There is just no automatic way to do it.

> Thus we as users have to
decide that and both changes have the same right to be present in the model. In
order to not just apply changes you can decide from which sides changes can be
taken.

This sentence is a bit confusing in terms of structure and clarity.

amolenaar (approved):

docs/user/docs/getting-started/working-with-git.md
Paula-Kli: I tried to make it a bit more clear

MoritzWeber0 (approved):

PR #1056, opened 2023-09-27
feat: Add network policies

dominik003:
This PR introduces network policies for our sessions and the main namespace. The session network policies should forbid any incoming traffic except coming from the main namespace or the ingress (required for jupyter to work). Here, we could also include a more fine grained limitation of the ingress case by only allowing this from jupyter sessions. The main namespace should forbid any incoming traffic from the outside, except from the ingress (otherwise the collab manager would not be accessible). Furthermore, we have to allow traffic between pods in this namespace.

This PR requires to add the `ingressNamespace` property to the `values.yaml` (only required for non openshift cluster in the current state)

### DIFF
--- a/helm/templates/routing/manager.networkpolicy.yaml
+++ b/helm/templates/routing/manager.networkpolicy.yaml
@@ -1,29 +1,27 @@
 # SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
 # SPDX-License-Identifier: Apache-2.0
 
-kind: NetworkPolicy
-apiVersion: networking.k8s.io/v1
-metadata:
-  name: {{ $.Release.Name }}-allow-from-{{ .Release.Namespace }}-namespace
-  namespace: {{ .Values.backend.k8sSessionNamespace }}
-spec:
-  podSelector: {}
-  ingress:
-    - from:
-        - namespaceSelector:
-            matchLabels:
-              kubernetes.io/metadata.name: {{ .Release.Namespace }}
-  policyTypes:
-    - Ingress
 {{- if .Values.cluster.ingressNamespace }}
----
 kind: NetworkPolicy
 apiVersion: networking.k8s.io/v1
 metadata:
   name: default-deny
-  namespace: {{ .Values.backend.k8sSessionNamespace }}
+  namespace: {{ .Release.Namespace }}
 spec:
   podSelector: {}
+  policyTypes:
+    - Ingress
+---
+kind: NetworkPolicy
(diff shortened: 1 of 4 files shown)


helm/templates/routing/manager.networkpolicy.yaml
```diff
@@ -0,0 +1,42 @@
...
+  ingress:
+    - from:
+        - podSelector: {}
+  policyTypes:
+    - Ingress
+---
+kind: NetworkPolicy
+apiVersion: networking.k8s.io/v1
+metadata:
+  name: allow-only-from-ingress
```
MoritzWeber0, line 31:
```suggestion
  name: allow-from-ingress
```
We allow it from ingress, but we don't forbid anything in the rule. `only` would imply that all other connections are forbidden.

helm/templates/routing/manager.networkpolicy.yaml
```diff
@@ -0,0 +1,42 @@
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if ne .Values.cluster.kind "Openshift" }}
```
MoritzWeber0, line 4:
```suggestion
```

I'd also apply the NetworkPolicies in OpenShift. This would align them with the general NetworkPolicies and we can change them via PRs.

MoritzWeber0: Instead, I'd only add the whole section if `ingressNamespace` is set. Otherwise, it would be breaking change and has to be marked accordingly.

MoritzWeber0 (changes requested):

MoritzWeber0:
Works perfectly fine. 

On main, I can query the metrics endpoint of the Capella session: 
image

After a `helm update`, the NetworkPolicies are applied and a connection is refused: 
image

MoritzWeber0:
> Works very well 👍
> 
> On main, I can query the metrics endpoint of the Capella session: image
> 
> After a `helm update`, the NetworkPolicies are applied and a connection is refused: image

helm/values.yaml
```diff
@@ -182,6 +182,8 @@ mocks:
 cluster:
   kind: Kubernetes # Kubernetes | OpenShift
 
+  ingressNamespace:
```
MoritzWeber0, line 185: Now, add a `--set` command to the Makefile, so that the networkpolicy is applied for the local setup.

MoritzWeber0 (approved):

PR #1059, opened 2023-09-28
refactor: Move frontend configuration to `values.yaml`

MoritzWeber0:
The frontend configuration had to be set during the build-time.
This is quite restrictive as it's not configurable for pre-built images,
e.g. our pre-built images in the Github registry.

Also, it was another configuration location, but we want to have all
configuration entries in one file.


backend/capellacollab/core/metadata.py
```diff
@@ -5,20 +5,34 @@
...
 
 class Metadata(pydantic.BaseModel):
     model_config = pydantic.ConfigDict(from_attributes=True)
 
     version: str
+    privacy_policy_url: str | None
+    imprint_url: str | None
+    provider: str | None
+    authentication_provider: str | None
+    environment: str | None
```
dominik003, line 19: There is a mismatch between the type in the config schema and this. I would either adjust the config schema to also allow an empty property (I think you need to add "null" as additional type to allow this) or remove the `| None` here.

(This also hold for the `cfg: dict[str, str | None]`, i.e., either just use `str` as value type or adjust the schema)

frontend/src/app/general/auth/auth/auth.component.ts
```diff
@@ -22,12 +22,12 @@ export class AuthComponent implements OnInit {
     }
   }
 
-  authProvider = environment.authentication;
   public params = {} as ParamMap;
 
   constructor(
     private authService: AuthService,
     private pageLayoutService: PageLayoutService,
+    public metadataService: MetadataService,
```
dominik003, line 30: Thats just a personal opinion but I prefer to have the services sorted by their visibility in the constructor (i.e., all public services at the top followed by the private ones). This helps directly identify and distinguish public and private services. What do you think about this?

frontend/src/app/general/metadata/version/version.component.html
```diff
@@ -7,21 +7,23 @@
   <a
     href="/docs/release-notes/"
     target="_blank"
-    (click)="versionService.clickedOnVersionNotes()"
+    (click)="metadataService.clickedOnVersionNotes()"
   >
     <div
       class="version"
```
dominik003, line 13: Here, at least this rule, can be adjusted to tailwind. Especially as we not often touch this component it might make sense to apply this small change in this PR.

frontend/src/app/general/metadata/version/version.service.ts
dominik003, line 1: First of all, I would rename this here to `metadata.service.ts`.
In addition, I would suggest moving the component + the service up directly into the `metadata` directory (i.e., no longer `metadata/version/metadata.service...` but directly `metadata/metadata.service...`).
Furthermore, I would suggest to also rename the version component to `metadata.component.x`.

dominik003 (changes requested):

backend/capellacollab/core/metadata.py
MoritzWeber0: The configuration schema doesn't allow `None`, but it allows that the key itself is missing. It is not required in the YAML schema. Therefore, it `None` is used if the configuration option is missing (to provide compatibility downwards).

frontend/src/app/general/auth/auth/auth.component.ts
MoritzWeber0: Done, is changed.

frontend/src/app/general/metadata/version/version.component.html
MoritzWeber0: Done and tested.

frontend/src/app/general/metadata/version/version.service.ts
MoritzWeber0: Done.

dominik003 (approved):

PR #1062, opened 2023-09-29
feat: Add possibility to archive projects

### LINKED ISSUE #397: Provide means to archive a project
There are quite a few projects that aren't developed anymore - but read-only access would still be of interest. For those it would be nice to indicate that the project is "frozen" by maybe graying-out the project card and adding an archived tag somewhere on the project page.

### LINKED ISSUE #792: Create custom confirmation dialog
Currently we use windows in various places to confirm certain actions. However, the style doesn't fit our UI and in general it would be better to use a custom component for these purposes. I think it makes sense to create only one custom confirmation dialog component for this.

dominik003:
Resolves #792, resolves #397


backend/capellacollab/projects/routes.py
```diff
@@ -182,6 +188,14 @@ def delete_project(
     crud.delete_project(db, project)
 
 
+def _delete_all_pipelines(db: orm.Session, project: models.DatabaseProject):
```
MoritzWeber0, line 191:
```suggestion
def _delete_all_pipelines_for_project(db: orm.Session, project: models.DatabaseProject):
```

frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.ts
```diff
@@ -64,12 +69,48 @@ export class ProjectMetadataComponent {
...
+
+  toggleArchive(): void {
+    if (this.project) {
+      const projectSlug = this.project.slug;
+
+      if (!this.project.is_archived) {
+        const dialogRef = this.dialog.open(ConfirmationDialogComponent, {
+          data: {
+            title: 'Archive Project',
+            text: 'Do you really want to archive the Project? This will unlink all pipelines of the existing project models.',
```
MoritzWeber0, line 88:
```suggestion
            text: 'Do you really want to archive the project? All pipelines of the existing project models will be deleted.',
```

MoritzWeber0: They are not just unlinked, they are also deleted.

backend/capellacollab/projects/routes.py
```diff
@@ -182,6 +188,14 @@ def delete_project(
...
+    pipelines: list[backups_models.DatabaseBackup] = []
+    for model in project.models:
+        pipelines.extend(backups_crud.get_pipelines_for_tool_model(db, model))
+    for pipeline in pipelines:
+        backups_crud.delete_pipeline(db, pipeline)
```
MoritzWeber0, line 196: A Database deletion is not sufficient. This would leave behind the cronjob in the cluster, which is not referenced anymore!

backend/capellacollab/projects/routes.py
```diff
@@ -85,15 +87,19 @@ def update_project(
...
-    if crud.get_project_by_slug(db, new_slug) and project.slug != new_slug:
-        raise fastapi.HTTPException(
-            status_code=status.HTTP_409_CONFLICT,
-            details={
-                "reason": "A project with a similar name already exists.",
-                "technical": "Slug already used",
-            },
-        )
+    if patch_project.name:
+        new_slug = slugify.slugify(patch_project.name)
```
MoritzWeber0, line 91: The new slug is never updated in the database.

backend/capellacollab/settings/modelsources/t4c/repositories/crud.py
```diff
@@ -85,6 +85,7 @@ def _get_user_write_t4c_repositories(
         .join(toolmodels_models.DatabaseCapellaModel.version)
         .where(tools_models.DatabaseVersion.name == version_name)
         .join(toolmodels_models.DatabaseCapellaModel.project)
+        .where(projects_models.DatabaseProject.is_archived.is_(False))
```
MoritzWeber0, line 88:
```suggestion
        .where(projects_models.DatabaseProject.is_archived == False)
```
Why don't you use this syntax?

backend/tests/projects/test_projects_routes.py
MoritzWeber0, line 1: The codecov/patch pipeline is red. Please add more tests.

frontend/src/app/general/confirmation-dialog/confirmation-dialog.component.html
```diff
@@ -0,0 +1,33 @@
...
+    appearance="outline"
+    *ngIf="data.requiredInput"
+  >
+    <input matInput [(ngModel)]="inputText" placeholder="Type here..." />
+  </mat-form-field>
+  <div class="flex justify-between">
+    <button mat-flat-button color="primary" (click)="onCancelClick()">
+      Cancel
+    </button>
+    <button
```
MoritzWeber0, line 24: Submit on enter doesn't work.

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.html
```diff
@@ -18,7 +18,7 @@ <h2>Project members</h2>
         <button
           (click)="openAddUserDialog()"
           mat-stroked-button
-          matTooltip="Add a new user"
+          [matTooltip]="'Add a new user'"
```
MoritzWeber0, line 21:
```suggestion
          matTooltip="Add a new user"
```

I don't see a need for the change, it only leads to less performance.

frontend/src/app/projects/project-overview/project-overview.component.html
```diff
@@ -22,7 +22,11 @@
       *ngFor="let project of (projectService.projects$ | async)!"
       [routerLink]="['/project', project.slug]"
     >
-      <mat-card matRipple class="mat-card-overview">
+      <mat-card
+        [ngClass]="{ 'bg-gray-300': project.is_archived }"
```
MoritzWeber0, line 26:
```suggestion
        [ngClass]="{ 'bg-archived': project.is_archived }"
```

Let's add a tailwind variable for it for more consistency.

frontend/src/app/projects/service/project.service.ts
```diff
@@ -119,10 +119,14 @@ export type UserMetadata = {
...
-  name?: string;
-  description?: string;
-  visibility?: ProjectVisibility;
+export type PostProject = {
+  name: string;
+  description: string;
+  visibility: ProjectVisibility;
+};
+
+export type PatchProject = Partial<PostProject> & {
```
MoritzWeber0, line 128: What is the `Partial` used for?

frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.ts
MoritzWeber0, line 1: There is a weird behaviour (probably a missing unsubscribe somewhere):

1. Try to add a non-existing user to a project
2. It will fail with an error message
3. Click on the error to make it disappear
4. Archive the project
5. The error message for the project user addition appears again

MoritzWeber0 (changes requested):

backend/capellacollab/projects/routes.py
dominik003: Isn't it? This here is just the safety check/ validation that there is no other project with the same slug. At the end of the function/route, then the `update_project` function is called which makes the actual update. In theory, we could also set `project.slug = new_slug` in here but IMO that should be in the responsibility of the actual update function.

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.html
dominik003: The change was there as I previously disabled the possibility of adding a user in an archived project. Simply forgot to also change this back.

frontend/src/app/projects/service/project.service.ts
dominik003: When we create a project all three properties (i.e., name, description, and visibility) are required. However, when we want to patch/update a project, we do not always want to provide all of these properties but only those we actually want to update. So `Partial` in this case, simply returns a type where all properties of, in this case `PostProject`, are optional.

backend/capellacollab/projects/routes.py
dominik003: Good catch missed that. Deletion should not work similar to. the usual pipeline deletion

frontend/src/app/general/confirmation-dialog/confirmation-dialog.component.html
dominik003: Wasn't sure if we do want to have this for such a confirmation dialog as explicitly clicking on "confirm" is somehow another safety mechanism but I its now added.

frontend/src/app/projects/project-detail/project-metadata/project-metadata.component.ts
dominik003: This was an already existing problem in the add user to project dialog. As discussed, I changed the dialog to no longer subscribe to `project$` but now requires a project to be provided when opening the dialog.

backend/capellacollab/settings/modelsources/t4c/repositories/crud.py
MoritzWeber0: Ah, this is annoying.

frontend/src/app/projects/service/project.service.ts
MoritzWeber0: Thanks for the explanation :)

MoritzWeber0 (approved):

Issue #745, opened 2023-06-12
Getting started failes on a clean ubuntu server due to missing stuff


vik378:
Looks like our "getting started" part of the readme is a bit incomplete - we miss a few dependencies:

- python 3(>=.11?)
- node and npm (>=16?)

it also could be an improvement in onboarding experience if we point a new starter to the published capella-dockerimages instead of build-from-scratch (saving build time). 

another question - do we still need EASE image?

next observation - after successfull launch the thing serves on 0.0.0.0:8080, however requesting that from anything that is not localhost results in 404 (design or intention?)

running the thing on a headless server leads to the below error (there is no browser to open):
```
/bin/bash: line 7: xdg-open: command not found
make: *** [Makefile:93: open] Error 127
```

Out of the box, session start fails with "an error occured, please try again" - but trying again leads to the same result as nothing changed ;-) 

The backend error log is full of exciting messages:

- Could not create an admin token. Please make sure that your Guacamole database is initialized properly. (File "/tmp/backend/capellacollab/sessions/guacamole.py", line 44, in get_admin_token)
- Exception during fetching of last seen. (requests.exceptions.ReadTimeout: HTTPConnectionPool(host='dev-prometheus-server', port=9118): Read timed out. (read timeout=2),   File "/tmp/backend/capellacollab/sessions/sessions.py", line 41, in get_last_seen)
-  Could not handle idle sessions ( File "/tmp/backend/capellacollab/sessions/idletimeout.py", line 50, in loop)

running `make provision-guacamole` gets it a little further - a session card appears with "failed to create session",
the new error is "Could not parse log"
```
Traceback (most recent call last):
  File "/tmp/backend/capellacollab/sessions/sessions.py", line 76, in _determine_session_state
    logs = get_operator().get_session_logs(session.id)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/backend/capellacollab/sessions/operators/k8s.py", line 413, in get_session_logs
    return self.v1_core.read_namespaced_pod_log(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kubernetes/client/api/core_v1_api.py", line 23747, in read_namespaced_pod_log
    return self.read_namespaced_pod_log_with_http_info(name, namespace, **kwargs)  # noqa: E501
[...]

PR #1065, opened 2023-10-02
feat: Add support for wildcard hosts

MoritzWeber0:
In some environments, e.g. local unmanaged single-node clusters, it might make sense to serve the application on another host.

This feature was requested in #745.

Kubernetes ingress has support for wildcard hosts. I can be enabled by just setting the host to `null` in the ingress definition.

Also, the commit fixes an issue with reverse proxies, where the application didn't serve anymore because the internal host didn't match the one configured in the `values.yml`.

### DIFF
--- a/backend/capellacollab/sessions/operators/k8s.py
+++ b/backend/capellacollab/sessions/operators/k8s.py
@@ -120,12 +120,33 @@ class KubernetesOperator:
             return False
 
     def create_public_route(
-        self, session_id: str, host: str, path: str, port: int
+        self,
+        session_id: str,
+        host: str,
+        path: str,
+        port: int,
+        wildcard_host: bool | None = False,
     ):
+        """Create a public route for the session
+
+        Parameters
+        ==========
+        session_id: str
+            The database ID of the session
+        host: str
+            The host to use for the route
+        path: str
+            The path to use for the route
+        port: int
+            The port to use for the route
+        wildcard_host: bool
+            Whether to use a wildcard host or not (serve on all hosts),
+            not supported for OpenShift
+        """
         if self.openshift:
             self._create_openshift_route(session_id, host, path, port)
         else:
-            self._create_ingress(session_id, host, path, port)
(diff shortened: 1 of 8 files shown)


backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -120,12 +120,33 @@ def validate(self) -> bool:
...
+            The database ID of the session
+        host: str
+            The host to use for the route
+        path: str
+            The path to use for the route
+        port: int
+            The port to use for the route
+        wildcard_host: bool
+            Whether to use a wildcard host or not (serve on all hosts),
+            not support for OpenShift
```
amolenaar, line 144:
```suggestion
            not supported by OpenShift
```

amolenaar (approved):
Is the `host` config parameter used in more places? Otherwise you could, for example, just set host to `"*"` in order to use this wildcard feature.

backend/capellacollab/sessions/operators/k8s.py
MoritzWeber0: OpenShift might support it, but we don't support it if OpenShift is used as cluster type.

amolenaar: Then let's change it to
```suggestion
            not supported for OpenShift
```

MoritzWeber0: Ah right, I've overlooked the typo. Thanks for the hint.

MoritzWeber0:
> Is the `host` config parameter used in more places? Otherwise you could, for example, just set host to `"*"` in order to use this wildcard feature.

It's used in a few places: For the Jupyter and Guacamole redirect and in the nginx configuration. Also "Kubernetes" doesn't understand "*" as value for host directly, we have to set it to null. This could be done with a mapping, but this is more explicit for now. 

I'd leave it for now, but we can address it another PR.

PR #1073, opened 2023-10-05
Fix readonly session branch selection

### LINKED ISSUE #1072: Branch selection in create-readonly-session isn't working
When requesting a Readonly session from a Project, one can choose the `Branch, tag or revision*`. This has no effect to the session. After connecting to it, the user is greeted by the model on the main branch.

amolenaar:
The selected branch was not passed to the backend, and the backend did not add it to the configuration of the session container.

Fixes #1072

### DIFF
--- a/backend/capellacollab/sessions/routes.py
+++ b/backend/capellacollab/sessions/routes.py
@@ -256,15 +256,15 @@ def models_as_json(
             )
         ):
             continue
-        yield git_model_as_json(git_model, entry.deep_clone)
+        yield git_model_as_json(git_model, entry.revision, entry.deep_clone)
 
 
 def git_model_as_json(
-    git_model: git_models.DatabaseGitModel, deep_clone: bool
+    git_model: git_models.DatabaseGitModel, revision: str, deep_clone: bool
 ) -> dict[str, str | int]:
     d: dict[str, str | int] = {
         "url": git_model.path,
-        "revision": git_model.revision,
+        "revision": revision,
         "depth": 0 if deep_clone else 1,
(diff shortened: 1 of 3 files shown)


backend/capellacollab/sessions/routes.py
```diff
@@ -257,15 +257,15 @@ def models_as_json(
...
 
 
 def git_model_as_json(
-    git_model: git_models.DatabaseGitModel, deep_clone: bool
+    git_model: git_models.DatabaseGitModel, revision: str, deep_clone: bool
 ) -> dict[str, str | int]:
     d = {
         "url": git_model.path,
-        "revision": git_model.revision,
+        "revision": revision or git_model.revision,
```
dominik003, line 268: Either the `or git_model.revision` is unnecessary here or the type hint should of `revision` should be `str | None`. However, from what I can see, we only use the `git_model_as_json` function here and `entry.revision` should never be `None`. Therefore, I would suggest to simply remove the `or git_model.revision`.

capella-dockerimages
dominik003, line 1: Is this intended to be in this PR?

dominik003 (approved):
LGTM. Interesting that we just completely ignored the revision we got from the front end.
I've already approved the PR, but it would be great if you could include/address the two comments I made before we merge it.

backend/capellacollab/sessions/routes.py
amolenaar: Indeed, it's probably caught by the model validator anyway.

capella-dockerimages
amolenaar: Not again :/

PR #1074, opened 2023-10-05
Group Angular updates in one PR

amolenaar:
Instead of having a bunch of NPM updates provided by Dependabot, each one not building, group all Angular related updates, so we have one concise update for all Angular packages.

This way we do not have to do the frontend updates ourselves.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,13 @@ updates:
     directory: "/frontend"
     schedule:
       interval: "weekly"
+    groups:
+      angular:
+        patterns:
+        - "@angular*"
+        update-types:
+        - "minor"
+        - "patch"
   # Capella Dockerimages
   - package-ecosystem: "gitsubmodule"
     directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,6 +18,7 @@ updates:
         patterns:
         - "@angular*"
         update-types:
+        - "major"
         - "minor"
         - "patch"
   # Capella Dockerimages


Paula-Kli (approved):
Good idea!


only question: why is dockerimages commited as well?

amolenaar:
> only question: why is dockerimages commited as well?

Human error. I corrected it.

.github/dependabot.yml
```diff
@@ -13,6 +13,13 @@ updates:
     directory: "/frontend"
     schedule:
       interval: "weekly"
+    groups:
+      angular:
+        patterns:
+        - "@angular*"
+        update-types:
+        - "minor"
+        - "patch"
```
MoritzWeber0, line 22:
```suggestion
        update-types:
        - "major"
        - "minor"
        - "patch"
```

I'd say it makes sense to get notified about new major versions too.

MoritzWeber0 (commented):
Love it, thanks for the PR! 🥳

.github/dependabot.yml
amolenaar: I'm not that familiar with Angular's upgrade scheme. If you do not expect everything to break on a major version upgrade, then this change is fine.

MoritzWeber0: I expect breaking changes, but dependabot doesn't auto-update on main. It just opens a PR so that we can looks into it.

MoritzWeber0 (approved):

PR #1085, opened 2023-10-10
build: Add ESLint and prettier tailwind plugins

MoritzWeber0:
It enforces a unique classname order and prevents contradicting classnames. All supported rules are described here: https://www.npmjs.com/package/eslint-plugin-tailwindcss

You just have to review four commits, you can skip the https://github.com/DSD-DBS/capella-collab-manager/pull/1085/commits/6358e342bf0f6fdeb0cbb9647692af0c82f2cb54 commit. In this commit, only the result `pre-commit run -a` was committed.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,6 +72,10 @@ repos:
     hooks:
       - id: prettier
         types_or: [ts, css, html, markdown]
+        files: "^frontend"
+        additional_dependencies:
+          - "prettier@^3.0.3"
+          - "prettier-plugin-tailwindcss@^0.5.5"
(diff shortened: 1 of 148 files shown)


Paula-Kli (commented):
I am curious: do you know why the tailwind class names order is switched in many cases even though you turned that off in the config?

Paula-Kli (approved):

MoritzWeber0:
> I am curious: do you know why the tailwind class names order is switched in many cases even though you turned that off in the config?

It's turned off because prettier takes care of it. There is no need to run it twice (in ESLint and Prettier). That's why I turned it off in ESLint.

PR #1087, opened 2023-10-10
Support Python 3.12

amolenaar:
Add Python 3.12 to the supported Python versions.

Updated Trove classifiers. Just Python should be enough. Python 2 is no longer a thing, right? ;)


Paula-Kli (approved):
works for me 👍

PR #1090, opened 2023-10-12
fix: Remove file ending in diagram route

Paula-Kli:
When using the diagram cache the route should not run into errors  when calling diagrams with a file ending or without.

### DIFF
--- a/backend/capellacollab/projects/toolmodels/diagrams/routes.py
+++ b/backend/capellacollab/projects/toolmodels/diagrams/routes.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import logging
+import pathlib
 from urllib import parse
 
 import fastapi
@@ -67,14 +68,24 @@ async def get_diagram_metadata(
     )
 
 
-@router.get("/{diagram_uuid}", response_class=fastapi.responses.Response)
+@router.get(
+    "/{diagram_uuid_or_filename}", response_class=fastapi.responses.Response
+)
 async def get_diagram(
-    diagram_uuid: str,
+    diagram_uuid_or_filename: str,
     handler: git_handler.GitHandler = fastapi.Depends(
         git_injectables.get_git_handler
     ),
     logger: logging.LoggerAdapter = fastapi.Depends(log.get_request_logger),
 ):
+    fileextension = pathlib.PurePosixPath(diagram_uuid_or_filename).suffix
+    if fileextension and fileextension.lower() != ".svg":
+        raise fastapi.HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail={"reason": f"File extension {fileextension} not supported"},
+        )
+
+    diagram_uuid = pathlib.PurePosixPath(diagram_uuid_or_filename).stem
     try:
         _, diagram = await handler.get_file_from_repository_or_artifacts(
             f"diagram_cache/{parse.quote(diagram_uuid, safe='')}.svg",
--- a/docs/user/docs/tokens.md
+++ b/docs/user/docs/tokens.md
@@ -41,3 +41,22 @@ authenticate with that against the Collaboration Manager API. One example is:
 ```zsh
 curl -u [username]:[token] https://[baseURL]/api/v1/projects
 ```
+
+Another example is working with the diagram cache of py-capellambse.
+The implementation of the capella modelling tool `capellambse` uses Python and lets you read and
+write models. For more information have a look at the
+[documentation](https://dsd-dbs.github.io/py-capellambse/) or the
+[Github repository](https://github.com/DSD-DBS/py-capellambse).
+
+```python
+model = capellambse.model.MelodyModel(
+  path="<path to the model on your machine>",
+  diagram_cache={
+    "path": "https://<your backend url>/api/v1/projects/<your project slug>/models/<your model slug>/diagrams/%s",
+    "username": "<username>",
+    "password": "<your PAT>",
+  }
+)
+```
+
+Having created a model like that you can e.g. with `model.diagrams[0]` get the first diagram.
--- a/backend/tests/projects/toolmodels/test_diagrams.py
+++ b/backend/tests/projects/toolmodels/test_diagrams.py
@@ -386,6 +386,121 @@ def test_get_single_diagram_from_artifacts(
 
 @responses.activate
 @pytest.mark.parametrize(
+    "git_type,git_response_status,git_query_params",
+    [
+        (
+            git_models.GitType.GITLAB,
+            404,
+            [
+                {
+                    "path": "diagram_cache/index.json",
+                    "ref_name": "diagram-cache/main",
+                },
+                {
+                    "path": "diagram_cache/_c90e4Hdf2d2UosmJBo0GTw.svg",
+                    "ref_name": "diagram-cache/main",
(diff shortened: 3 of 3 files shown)


backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -67,14 +67,15 @@ async def get_diagram_metadata(
     )
 
 
-@router.get("/{diagram_uuid}", response_class=fastapi.responses.Response)
+@router.get("/{diagram_file_name}", response_class=fastapi.responses.Response)
```
MoritzWeber0, line 70:
```suggestion
@router.get("/{diagram_uuid_or_filename}", response_class=fastapi.responses.Response)
```

We do still accept and I'd keep it as preferred method. The clients should not care about the fileending.

backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -67,14 +67,15 @@ async def get_diagram_metadata(
...
 async def get_diagram(
-    diagram_uuid: str,
+    diagram_file_name: str,
     handler: git_handler.GitHandler = fastapi.Depends(
         git_injectables.get_git_handler
     ),
     logger: logging.LoggerAdapter = fastapi.Depends(log.get_request_logger),
 ):
+    diagram_uuid = diagram_file_name.rsplit(".", maxsplit=1)[0]
```
MoritzWeber0, line 78:
```suggestion
    diagram_uuid = diagram_file_name.split(".")[0]
```

UUIDs can't contain a `.`. Therefore, we can safely assume that everything after the first "." belongs to the file extension.

backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -67,14 +67,28 @@ async def get_diagram_metadata(
...
-    diagram_uuid: str,
+    diagram_uuid_or_filename: str,
     handler: git_handler.GitHandler = fastapi.Depends(
         git_injectables.get_git_handler
     ),
     logger: logging.LoggerAdapter = fastapi.Depends(log.get_request_logger),
 ):
+    splitted = diagram_uuid_or_filename.split(".")
+    if len(splitted) > 1:
+        diagram_uuid, fileextension = splitted
```
MoritzWeber0, line 82: Unpacking will fail if `len(splitted) > 2`. We have to consider this case. We can use pathlib to parse the file name and to find out the suffix: https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.suffix

backend/capellacollab/projects/toolmodels/diagrams/routes.py
MoritzWeber0, line 1: We should add it to the documentation again.

MoritzWeber0 (changes requested):

backend/capellacollab/projects/toolmodels/diagrams/routes.py
Paula-Kli: I have seen this with the Path suffix - alternatively we could also use rsplit again https://github.com/DSD-DBS/capella-collab-manager/pull/1090#discussion_r1356658664

MoritzWeber0: I'd prefer the PurePosixPath suffix.

backend/capellacollab/projects/toolmodels/diagrams/routes.py
MoritzWeber0: Yes, like we had it before, I'd say.

docs/user/docs/tokens.md
```diff
@@ -41,3 +41,23 @@ authenticate with that against the Collaboration Manager API. One example is:
 ```zsh
 curl -u [username]:[token] https://[baseURL]/api/v1/projects
 ```
+
+Another example is working with the diagram cache of py-capellambse.
+Capellambse is an
```
MoritzWeber0, line 46:
```suggestion
`capellambse` is an
```

docs/user/docs/tokens.md
```diff
@@ -41,3 +41,23 @@ authenticate with that against the Collaboration Manager API. One example is:
...
+implementation of the capella modelling tool using python and lets you read and
```
MoritzWeber0, line 47:
```suggestion
implementation of the capella modelling tool using Python and lets you read and
```

docs/user/docs/tokens.md
```diff
@@ -41,3 +41,23 @@ authenticate with that against the Collaboration Manager API. One example is:
...
+write models. For more information have a look at the
+[documentation](https://dsd-dbs.github.io/py-capellambse/) or the
+[github repository](https://github.com/DSD-DBS/py-capellambse).
```
MoritzWeber0, line 50:
```suggestion
[Github repository](https://github.com/DSD-DBS/py-capellambse).
```

docs/user/docs/tokens.md
```diff
@@ -41,3 +41,23 @@ authenticate with that against the Collaboration Manager API. One example is:
...
+
+```python
+model = capellambse.model.MelodyModel(
+  path="<path to the model on your machine>",
+  diagram_cache={
+    "path": "http://<your backend url>/api/v1/projects/<your project slug>/models/<your model slug>/diagrams/%s",
```
MoritzWeber0, line 56:
```suggestion
    "path": "https://<your backend url>/api/v1/projects/<your project slug>/models/<your model slug>/diagrams/%s",
```

We should not promote the use of `http`. Instead, we should promote the use of encrypted transfer.

backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -67,14 +68,24 @@ async def get_diagram_metadata(
...
+    diagram_uuid_or_filename: str,
     handler: git_handler.GitHandler = fastapi.Depends(
         git_injectables.get_git_handler
     ),
     logger: logging.LoggerAdapter = fastapi.Depends(log.get_request_logger),
 ):
+    fileextension = pathlib.PurePosixPath(diagram_uuid_or_filename).suffix
+    if fileextension and fileextension.lower() != ".svg":
+        raise fastapi.HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
```
MoritzWeber0, line 84: Should be a bad request, as we don't support SVG and the user requested SVG.

Paula-Kli: I think @Wuestengecko said that in case capellembse requests another file format it only understands if exactly a 404 is thrown...

MoritzWeber0: Oh, alright. I forgot to reply to that, sorry.

In my opinion, this is exactly what should happen. The user tries to use a feature that we don't support. In this case, it's a wrong configuration - the diagram cache doesn't work with the configuration. When I as user pass the configuration that I want to use the diagram cache, then I'd also expect that it either works or an error is raised, it should not be ignored silently.

Paula-Kli: alright - so I will change it to 400 Bad request

MoritzWeber0 (approved):

Issue #1100, opened 2023-10-16
Add commit history


MoritzWeber0:
On model level, we should have a commit history. The commit history is fetched from different sources: 

- Git commits
- TeamForCapella commits (not relevant for the first version)

For each commit, we should display: 
- Author (+ link to use profile, see https://github.com/DSD-DBS/capella-collab-manager/issues/1088)
- Title
- Description
- Date of the commit

romeonicholas:
@MoritzWeber0, few questionsRight now the git_instances table has types for GITHUB, GITLAB, AZUREDEVOPS, and GENERAL.

1. Do we need to worry about GENERAL? I'd probably just show a message saying it wasn't a supported type if it wasn't one of the first three, but I'm not sure if it's even being used
2. I instinctively started looking at GitHub's API first, but do you know which host is the most used? Might make sense to release support for one when it's ready and add in the others as issues come up or to refine the whole thing based on feedback

romeonicholas:
<img width="416" alt="Screenshot 2023-12-04 at 10 55 46" src="https://github.com/DSD-DBS/capella-collab-manager/assets/79602065/43366244-3d52-4ccb-9a7c-b52f0a3e3988">

@MoritzWeber0 -- questions!

Linking to author profile: I should be able to link to a commit author's Collab profile if they used the same email for both the Collab Manager and GitLab (and just leave it unlinked with the commit author's name if not), but do we know if that's the norm for our users on production? And if so, do you think it's easier to read commit authors as their username or their actual name? ("nicholasromeo" or "Nicholas Romeo"?)

Number of branches and commits: Should we only show the primary git model branch, every git model branch, or every branch that we can see for any given model at the provided URL? For a project with 1 model and a single branch, a user on a laptop could see maybe 8 commits on screen at once. For 2 models or branches that becomes about 3ish commits each. I'm thinking primary git model revision should be open and show at least a few of the most recent commits on page load, while the other models or branches start out collapsed. Does that seem like a good start, or do you have some other specific use case you wanted?

Descriptions: Shown from page load, or only with a click-to-expand? (Descriptions not included at all in picture above.) I'm not sure what our users' commits generally look like (just titles? Important info in description?)

PR #1105, opened 2023-10-17
feat: Add user profile page

### LINKED ISSUE #1088: Add user profiles
A user should have a "public" profile.

The profile page should contain the username for the beginning. Also it would be helpful to see all common/joint projects with the user. 
A user should be able to open their profile to see the username.

Paula-Kli:
closes #1088


backend/capellacollab/projects/routes.py
```diff
@@ -68,6 +68,40 @@ def get_projects(
...
+    current_users_projects = [
+        association.project
+        for association in user.projects
+        if not association.project.visibility == models.Visibility.INTERNAL
+    ]
+    projects_for_common = [
+        association.project
+        for association in user_for_common_projects.projects
+        if not association.project.visibility == models.Visibility.INTERNAL
+    ]
```
MoritzWeber0, line 94: This can be outsourced to a function as it's the same implementation for user and user_for_common_projects.

frontend/src/app/users/users-profile/users-profile.component.html
```diff
@@ -0,0 +1,27 @@
+<!--
+ ~ SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+ ~ SPDX-License-Identifier: Apache-2.0
+ -->
+
+<h2>Profile of {{ userName }}</h2>
```
MoritzWeber0, line 6: This page looks very sad: 

image

Can you optimize the UI a bit? A few ideas: 
- Add a profile image (just an avatar placeholder)
- Indicate if it's the own profile
- Add the year when the user joined

frontend/src/app/users/users-profile/users-profile.component.html
```diff
@@ -0,0 +1,27 @@
+<!--
```
MoritzWeber0, line 1: The user overview and the global user overviews should have links to the user profiles when clicking on a new button on an individual user.

frontend/src/app/users/users-profile/users-profile.component.ts
```diff
@@ -0,0 +1,41 @@
...
+  constructor(
+    public userService: UserService,
+    public projectService: ProjectService,
+    private route: ActivatedRoute,
+  ) {}
+
+  ngOnInit() {
+    this.route.queryParams.subscribe((params) => {
+      this.userId = params.userId;
+      this.userName = params.userName;
```
MoritzWeber0, line 29: Please use the consistent URI schema for the frontend, more information here: https://dsd-dbs.github.io/capella-collab-manager/development/frontend/routes/

==> It should be `/user/<id>`, no need for query params here

Also, I'd remove `params.userName`. We shouldn't allow the user to provide the username for a specific user id. For example, the user could open the page with the params `?userId=1&userName=test` and it would display the username test (even though the user has a different name in the backend).

Only the ID is required and can be used to fetch the user from the backend.

frontend/src/app/general/header/header.component.html
```diff
@@ -48,6 +48,14 @@
           </a>
         </div>
         <mat-menu #profileMenu="matMenu">
+          <a
```
MoritzWeber0, line 51: The Profile is missing in the mobile view (add it to the hamburger menu)

frontend/src/app/users/users-profile/users-profile.component.ts
MoritzWeber0: Once you fetch the user from the backend, you can also save the user in the user.service like described in the development docs: https://dsd-dbs.github.io/capella-collab-manager/development/frontend/code-style/#retriving-resources

frontend/src/app/users/users-profile/users-profile.component.html
```diff
@@ -0,0 +1,27 @@
...
+      [routerLink]="['/project', project.slug]"
+    >
+      <mat-card-content>
+        <b>{{ project.name }}</b
+        ><br />
+        {{ project.description }}
+      </mat-card-content>
+    </mat-card>
+  </ng-container>
+  <ng-template #elseBlock>You do not have any common projects.</ng-template>
```
MoritzWeber0, line 26: image

A new line is missing here between Projects and you

frontend/src/app/users/users-profile/users-profile.component.html
```diff
@@ -0,0 +1,27 @@
+<!--
+ ~ SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+ ~ SPDX-License-Identifier: Apache-2.0
+ -->
+
+<h2>Profile of {{ userName }}</h2>
+
+<div *ngIf="userId !== currentUser?.id">
+  Common Projects
```
MoritzWeber0, line 9: This is a small title and should be indicated as title, make it bold and a bit larger

frontend/src/app/users/users-profile/users-profile.component.html
```diff
@@ -0,0 +1,27 @@
...
+
+  <ng-container
+    *ngIf="(projectService.projects$ | async)?.length; else elseBlock"
+  >
+    <mat-card
+      *ngFor="let project of projectService.projects$ | async"
+      class="!ml-0 cursor-pointer"
+      [routerLink]="['/project', project.slug]"
+    >
+      <mat-card-content>
```
MoritzWeber0, line 19: Common projects don't work if at least one admin is involved. For example, try to find common projects while logged in as admin. It always shows that there are no common projects.

backend/capellacollab/projects/routes.py
```diff
@@ -68,6 +68,40 @@ def get_projects(
...
+) -> list[models.DatabaseProject]:
+    current_users_projects = [
+        association.project
+        for association in user.projects
+        if not association.project.visibility == models.Visibility.INTERNAL
+    ]
+    projects_for_common = [
+        association.project
+        for association in user_for_common_projects.projects
+        if not association.project.visibility == models.Visibility.INTERNAL
```
MoritzWeber0, line 93: Why do you exclude internal projects? When a user is explicitly added to a project, for example as project manager, it would be nice to see this on the profile page.

frontend/src/app/users/users-profile/users-profile.component.ts
Paula-Kli: fair - I totally see your point 
the problem that I was trying to solve was: that as a non-admin user I am not allowed to fetch users from the backend... which also makes sense in order to not just ask for ids and get users back thus I wanted to use the username as long as we anyways have it 
do you have another solution for that problem?

frontend/src/app/users/users-profile/users-profile.component.html
Paula-Kli: but what would you expect there to happen? I found this case but as admin I do not necessarily have any projects while I can still see all projects... does that mean I am in all projects and have common projects with every user?

backend/capellacollab/projects/routes.py
```diff
@@ -68,6 +68,40 @@ def get_projects(
     return projects
 
 
+@router.get(
+    "/common/{user_id}", response_model=list[models.Project], tags=["Projects"]
```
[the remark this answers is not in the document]
Paula-Kli: does it make sense to go to a users route and get projects retourned?

frontend/src/app/users/users-profile/users-profile.component.html
MoritzWeber0: Consider an admin to have access to all projects.

User 1 has access to project 1: The admin would have the command project 1. 
User 2 has no access to any projects: There are no command projects with admins. 

More or less, the feature can be used for admins to see all projects the user has access to.

backend/capellacollab/projects/routes.py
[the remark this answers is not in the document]
MoritzWeber0: It's not a "users"-route anymore if you add `common-projects`. You can also declare the router in the projects module and include it from the users routes.

frontend/src/app/users/users-profile/users-profile.component.html
Paula-Kli: what do you mean by that? you can now get to the view via the project user overview and the user overview in the settings - are there any other?

MoritzWeber0: No, these are the only two :)

frontend/src/app/users/users-profile/users-profile.component.html
Paula-Kli: Bildschirmfoto 2023-11-14 um 09 42 35
looks like that currently

MoritzWeber0: I'd move the avatar to the left side. 
Also, we should only indicate the year when the user joined, not the full date (and this date is optional, you should consider it, some users don't have the date).

frontend/src/app/users/users-profile/users-profile.component.html
```diff
@@ -0,0 +1,46 @@
...
+              class="!ml-0 cursor-pointer"
+              [routerLink]="['/project', project.slug]"
+            >
+              <mat-card-content>
+                <b>{{ project.name }}</b
+                ><br />
+                {{ project.description }}
+              </mat-card-content>
+            </mat-card>
+          </ng-container>
```
romeonicholas, line 35: I mostly just wanted to try out Angular 17's new control flow (😄) but I had a few suggestions to fill it out, and keep the text from overflowing in any direction on mobile if you're interested: 
mobile_shared_projects

```suggestion
<mat-divider></mat-divider>
        <div class="grid grid-cols-2 gap-4 md:grid-cols-4">
          <ng-container
            *ngIf="(commonProjects$ | async)?.length; else elseBlock"
          >
            <mat-card
              *ngFor="let project of commonProjects$ | async"
              class="!ml-0 cursor-pointer"
              [routerLink]="['/project', project.slug]"
            >
              <mat-card-content>
                <b class="line-clamp-1">{{ project.name }}</b
                ><br />
                @if (project.description) {
                <span class="line-clamp-3 md:line-clamp-5">{{
                  project.description
                }}</span>
                } @else {
                <i class="line-clamp-3 md:line-clamp-5"
                  >No description provided</i
                >
                }
              </mat-card-content>
            </mat-card>
          </ng-container>
```

Paula-Kli: thanks for the comment! I incorporated the feedback!

romeonicholas (approved):
Looks good ✅ 

Is there an epic or anything for upcoming features that the profile page will enable? As far as I could tell, only admins would be able to find anyone's profile page besides their own, because there's no global users view and the URL doesn't mean anything to a user ("[baseurl]/user?userId=[userId]" instead of "[baseurl]/users/[username]"), but I assume this is just a first step into setting up the user profile?

Paula-Kli:
> Looks good ✅
> 
> Is there an epic or anything for upcoming features that the profile page will enable? As far as I could tell, only admins would be able to find anyone's profile page besides their own, because there's no global users view and the URL doesn't mean anything to a user ("[baseurl]/user?userId=[userId]" instead of "[baseurl]/users/[username]"), but I assume this is just a first step into setting up the user profile?

Maybe @MoritzWeber0 can tell a bit more but yes firstly you can check for common projects with other users if you can see them somewhere. In addition there is a place where you can see your own username as that was nowhere placed anymore. In addition, that sets the starting point for more things which can be hostet at the user profile page

MoritzWeber0:
> > Looks good ✅
> > Is there an epic or anything for upcoming features that the profile page will enable? As far as I could tell, only admins would be able to find anyone's profile page besides their own, because there's no global users view and the URL doesn't mean anything to a user ("[baseurl]/user?userId=[userId]" instead of "[baseurl]/users/[username]"), but I assume this is just a first step into setting up the user profile?
> 
> Maybe @MoritzWeber0 can tell a bit more but yes firstly you can check for common projects with other users if you can see them somewhere. In addition there is a place where you can see your own username as that was nowhere placed anymore. In addition, that sets the starting point for more things which can be hostet at the user profile page

Sure! We have a bunch of features coming up, which will also make other users visible. 
One example would be the commit history (https://github.com/DSD-DBS/capella-collab-manager/issues/1100), which all project users can see - with hyperlinks to commit authors.

frontend/src/app/users/users-profile/users-profile.component.ts
MoritzWeber0: Don't think this is resolved... 
The URL is still `user?userId=17`... 
Will mark it as unresolved.

MoritzWeber0:
> Looks good ✅
> 
> Is there an epic or anything for upcoming features that the profile page will enable? As far as I could tell, only admins would be able to find anyone's profile page besides their own, because there's no global users view and the URL doesn't mean anything to a user ("[baseurl]/user?userId=[userId]" instead of "[baseurl]/users/[username]"), but I assume this is just a first step into setting up the user profile?

I try to avoid usernames in the URL. Our current usernames can also contain special chars, which leads to URL encoded usernames in the URL and is not always straight-forward.

frontend/src/app/users/users-profile/users-profile.component.ts
```diff
@@ -0,0 +1,68 @@
...
+              error: () => this.commonProjects.next(undefined),
+            });
+          } else {
+            this.userService
+              .getUserHistory(this.userService.user!)
+              .subscribe((history) => {
+                if (history.created) {
+                  this.joinedString = `You joined the collaboration manager in ${new Date(
+                    history.created,
+                  )
```
MoritzWeber0, line 52: image

I'd move this up. The text should be below the username.

MoritzWeber0, line 52: The year when the user joined is missing in other user profiles:
image

frontend/src/app/settings/core/user-settings/user-settings.component.html
```diff
@@ -22,25 +22,52 @@ <h2>Manage users</h2>
...
     >
       <div mat-subheader>Administrators</div>
-      <mat-list-option
-        *ngFor="let user of getUsersByRole('administrator')"
-        [value]="user"
-      >
-        <mat-icon mat-list-icon>account_circle</mat-icon>
-        <div mat-line>{{ user.name }}</div>
-        <div mat-line>{{ projectUserService.ADVANCED_ROLES[user.role] }}</div>
+      <mat-list-option *ngFor="let user of getUsersByRole('administrator')">
```
MoritzWeber0, line 27: image

Instead of this unintuitive approach (admins have to click on the info-symbol) to upgrade or downgrade users and overflow of the text in the icon, keep it simple: 

- Implement the same design like in the project details here.
- Move the user information to the user profile page, only visible for admins.

MoritzWeber0 (changes requested):

frontend/src/app/settings/core/user-settings/user-settings.component.html
Paula-Kli: what does your second point mean? as a admin doesn't it make sense to see whether it is a user or a administrator? or do you want to have that indirectly displayed via the up and downgrade arrows?

Paula-Kli: in addition I would leave the "i" symbol as otherwise we are losing the user information right?

MoritzWeber0: > in addition I would leave the "i" symbol as otherwise we are losing the user information right?

No, we don't need it. Check the second point: Move the user information to the user profile page, only visible for admins. It should just be part of the user profile page, not the user settings anymore. To see it, the admin would open the profile page.

MoritzWeber0: > what does your second point mean? as a admin doesn't it make sense to see whether it is a user or a administrator? or do you want to have that indirectly displayed via the up and downgrade arrows?

I don't understand the question. The role is indicated below the username: 
image

frontend/src/app/users/users-profile/users-profile.component.ts
Paula-Kli: should be fixed now

frontend/src/app/settings/core/user-settings/user-settings.component.html
Paula-Kli: I see - thanks for the clarification 
I did not get what you meant

MoritzWeber0 (approved):
I've finished my review now and still had a few important comments. Without their implementation I can't merge the PR.

To not further play ping-pong on the reviews and slow down the review process, I went ahead and implemented my comments. Please review my commit and make sure that: 
- You understand the commit description
- You understand all changes code and why the change was needed. In case of questions, reach out to me. 
- Approve the PR or request additional changes.

I hope that we can get it merged next Monday.

Paula-Kli (commented):
Thank you for trying to solve this fast and implementing your suggestions!

This is my approval as a PR author I cannot approve the PR

PR #1106, opened 2023-10-18
build: Let Git ignore dirty `capella-dockerimages` submodule

MoritzWeber0:
Git submodules are often staged and committed accidentally. We are generally not supposed to commit dirty submodules, so we can let Git ignore them. Dirty submodules are no longer shown as modified in git status.


Paula-Kli (approved):
Good idea!

PR #1112, opened 2023-10-23
fix: Properly set rbac values and disable sidecar

dominik003:


MoritzWeber0:
The options are needed to be "OpenShift" compatible.

MoritzWeber0 (approved):

Issue #1004, opened 2023-09-06
Initialize training setup, spawn related sessions


MoritzWeber0:
It has to be possible to start a training: 

1. Spawn the sessions for all defined required tools
2. Initialize the sessions with the linked Git repository (clone the Git repository)
3. Open the session in the "training mode" viewer with pre-defined layout

vik378:
one quick comment re no persistence required for training - you may want to save your progress and come back to it next day so it would be nice if we could figure something for it

MoritzWeber0:
This should be covered by this paragraph:

> There should be a check-box if the pre-provisioned workspace should be persistent or not. [...] If the check-box is selected, the linked Git repositories of the models are loaded in the user's persistent volume. In training mode, non persistent workspaces are not allowed. There will be no check-box for training projects.

To clarify it: When spawning a pre-provisioned workspace for a training project, there will be no option to request a read-only workspace. Instead, the workspace is always persistent to prevent data loss.

PR #1123, opened 2023-10-24
feat: Training mode for projects

amolenaar:
Support different types of projects. Currently _general_ and _training_. 

By default you're creating a normal project. This is what most users tend to do and I think we should not bother with training projects upon project creation directly. You can, however, modify a project and set it to _training_.

Training mode is depicted by a "study" icon before the name. Maybe we want to change the color of the header too.

image

Synchronization and persistent session buttons are not shown for models in a training project.

No additional behavior has been implemented.

Related to #1004.

### DIFF
--- a/frontend/src/app/projects/create-project/create-project.component.ts
+++ b/frontend/src/app/projects/create-project/create-project.component.ts
@@ -54,6 +54,7 @@ export class CreateProjectComponent implements OnInit, OnDestroy {
           name: this.form.value.name!,
           description: this.form.value.description!,
           visibility: this.form.value.visibility! as ProjectVisibility,
+          type: 'general',
         })
         .subscribe((project) => {
           this.toastService.showSuccess(
--- a/frontend/src/app/projects/models/backup-settings/view-logs-dialog/view-logs-dialog.component.html
+++ b/frontend/src/app/projects/models/backup-settings/view-logs-dialog/view-logs-dialog.component.html
@@ -4,7 +4,7 @@
  -->
 
 <div *ngIf="pipelineRunService.pipelineRun$ | async">
-  <div class="flex flex-col justify-between">
+  <div class="flex flex-row justify-between">
     <h1>
       Logs for pipeline run
       {{ (pipelineRunService.pipelineRun$ | async)?.id }}
--- a/frontend/src/app/projects/models/backup-settings/view-logs-dialog/view-logs-dialog.component.html
+++ b/frontend/src/app/projects/models/backup-settings/view-logs-dialog/view-logs-dialog.component.html
@@ -4,7 +4,7 @@
  -->
 
 <div *ngIf="pipelineRunService.pipelineRun$ | async">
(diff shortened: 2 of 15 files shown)


frontend/src/app/projects/models/diagrams/model-diagram-dialog/model-diagram-dialog.component.html
```diff
@@ -3,7 +3,7 @@
  ~ SPDX-License-Identifier: Apache-2.0
  -->
 <div class="dialog">
-  <div class="header">
+  <div class="header flex-col">
```
dominik003, line 6: Here the `header` class does not contain the `display: flex` attribute, so what exactly does `flex-col` does in case of a non-existing `flex`?

Update: Same as above. Dialog has a custom `header` and has no `mat-card-overview` as parent (so you can simply remove the `flex-col` here again)

frontend/src/app/projects/create-project/create-project.component.ts
dominik003, line 1: Is there a reason why we current cannot create a project of type `training` but when editing a project we can change the type to `training`? So I would probably also directly add the possibility to create a training project (can be similar to the `ProjectVisibility`).

frontend/src/app/projects/service/project.service.ts
```diff
@@ -142,3 +153,8 @@ export const ProjectVisibility = {
   private:
     'Private - users need to be explicitly granted access to see the project.',
 };
+
+export const ProjectTypes = {
```
dominik003, line 157: I was first really confused (also because of the `ProjectVisibility`) but wouldn't it be better to name this here `ProjectTypeDescriptions`? Because right now, we would have `ProjectType` and `ProjectTypes` and when only seeing the names I would not know what they are used for and whats the difference between them.
(It is even more confusing in case of the `ProjectVisibility` as there the type and const have the exact same name but thats not part of this PR)

Probably I am missing something here, so feel free to drop your reasoning/opinion here.

dominik003 (commented):
Nice little PR to include the ability to create training projects! 
There are just a few minor comments that shouldn't take long to fix and then we can merge this.

frontend/src/app/projects/models/backup-settings/view-logs-dialog/view-logs-dialog.component.html
```diff
@@ -4,7 +4,7 @@
  -->
 
 <div *ngIf="pipelineRunService.pipelineRun$ | async">
-  <div class="header">
+  <div class="flex flex-col justify-between">
```
dominik003, line 7:
```suggestion
  <div class="flex flex-row justify-between">
```

dominik003 (approved):
Nice changes! I already approved it but would be great if you could address the one last open comment.

MoritzWeber0 (approved):
LGTM, thanks for the implementation and sorry for the long review wait time.

PR #1125, opened 2023-10-24
feat(k8s): Add Pod disruption budgets

### LINKED ISSUE #1089: Add PodDisruptionBudget to prevent session termination during cluster updates
Cluster operators can update their clusters and nodes at any time. The cluster update process usually requires a rolling restart of deployments. Termination of sessions can lead to disruptions and data loss for the user and should be prevented. In addition, we have to ensure that there is at least one container running for the management portal services.

Session deployments
---------------------

More information is available here: https://kubernetes.io/docs/tasks/run-application/configure-pdb/
The interesting section is: 
```
Single-instance Stateful Application:
Concern: do not terminate this application without talking to me.
Possible Solution 1: Do not use a PDB and tolerate occasional downtime.
Possible Solution 2: Set PDB with maxUnavailable=0. Have an understanding (outside of Kubernetes) that the cluster operator needs to consult you before termination. When the cluster operator contacts you, prepare for downtime, and then delete the PDB to indicate readiness for disruption. Recreate afterwards.
```

While solution 1 is not an option for us (session termination is distracting and can lead to data loss), we should implement solution 2 for sessions.

Management portal deployments
---------------------------------

The management portal deployments (backend, frontend, guacamole and guacd) should always have at least one healthy Pod running. Therefore `minAvailable` should be set to 1.

MoritzWeber0:
Kubernetes allows a lot flexibility when it comes to cluster and node updates.
Cluster updates can happen at any time and can lead to massive interruptions (Guacamole not available) and session termination with potential data loss.

Pod discruption budgets define under which circumstances the cluster operator can terminate or restart Pods.

Our new policy is now:

- For each service of the management portal, at least one Pod has to be up&running at any time.
- Session termination by the control plane is not allowed. The cluster operator has to contact the system administrator or wait until the session is terminated.

Resolves #1089

### DIFF
--- a/backend/capellacollab/sessions/operators/k8s.py
+++ b/backend/capellacollab/sessions/operators/k8s.py
@@ -96,6 +96,7 @@ class KubernetesOperator:
         self.v1_apps = client.AppsV1Api(api_client=self.client)
         self.v1_batch = client.BatchV1Api(api_client=self.client)
         self.v1_networking = client.NetworkingV1Api(api_client=self.client)
+        self.v1_policy = client.PolicyV1Api(api_client=self.client)
         self._openshift = None
 
     @property
@@ -200,6 +201,11 @@ class KubernetesOperator:
             limits=limits,
         )
 
+        self._create_disruption_budget(
+            name=_id,
+            deployment_name=_id,
+        )
+
         service = self._create_service(
             name=_id,
             deployment_name=_id,
@@ -224,6 +230,13 @@ class KubernetesOperator:
         if dep_status := self._delete_deployment(name=_id):
             log.info(
                 "Deleted deployment %s with status %s", _id, dep_status.status
+            )
+
+        if disrupt_status := self._delete_disruptionbudget(name=_id):
+            log.info(
+                "Deleted Pod discruption budget %s with status %s",
+                _id,
+                disrupt_status.status,
             )
 
         if loki_enabled and (conf_status := self._delete_config_map(name=_id)):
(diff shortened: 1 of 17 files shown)


MoritzWeber0:
Did my best to increase the test coverage by testing the `start_session` and `kill_session` functions. Patch coverage is still red as the general coverage of the k8s operator is bad. Overall coverage and k8s coverage are better.

amolenaar (approved):
Looks good.

Isn't it a bit overkill to create one PodDisruptionBudget per backend service?

I'm not sure how the PostgreSQL databases behave with this PDB: the database can be opened by one postgres instance at a time, right?

MoritzWeber0:
> Looks good.
> 
> Isn't it a bit overkill to create one PodDisruptionBudget per backend service?
> 
> I'm not sure how the PostgreSQL databases behave with this PDB: the database can be opened by one postgres instance at a time, right?

The PDB doesn't provide any information about how many Pods can run at the same time. It just says: Don't terminate the Pod if there is  other container of the deployment running. Now, this is kind of deadlock: It can't terminate the old Pod and doesn't start a new Pod because of the "Recreate" strategy type.

Now, the cluster operator has to decide what happens, sometimes they get in touch with the system administrators of the application and tell them to restart the database, sometimes they don't respect it at all. 

Why we should still add the PDB: Cluster administrators can define a CronJob to disable / enable the PDB at specific times. This allows the system administrators to specific timeframes, e.g. during night, where updates are allowed.

PR #1131, opened 2023-10-25
feat: Allow direct RDP connections

### LINKED ISSUE #6: Enable direct rdp connection
For some heavy ops (like upload of initial content) it would be nice to have a direct RDP connection.

MoritzWeber0:
Resolves #6

Error case: 
image


MoritzWeber0:
In v3.0.0, we introduced support for different connection methods. Since this PR has a lot of overlap and conflicts with main, it's easier to recreate it. However, the feature currently has a low priority, so this won't happen anytime soon.

PR #1133, opened 2023-10-26
feat: Special launcher for training projects

amolenaar:
This MR introduces a special launcher for training sessions:

image

This launcher will launch all models in the _training_ project.

TODO:

- [x] Mount persistent workspace
- [x] Load models in special folders for the tool/version.
- [x] slug for tools and versions. Slug should not change when tool/version is renamed

See #1004.

### Open questions

* Should the backend error on misconfigured models, or just skip them? (E.g. models without a primary git model, tooling with no readonly image). Outcome: ignore mis-/incompletely configured models

### Out of scope

* Clicking the launch button currently directs you to the sessions view. It should direct you to the session viewer eventually, but we cannot show that one yet for starting sessions.
* The new card is only for training projects. General projects still have the current "read-only" card.

### DIFF
--- a/backend/capellacollab/alembic/versions/2827788b8d2d_add_slugs_for_tools_and_versions.py
+++ b/backend/capellacollab/alembic/versions/2827788b8d2d_add_slugs_for_tools_and_versions.py
@@ -24,7 +24,8 @@ depends_on = None
 def upgrade():
     for table_name in ("tools", "versions"):
         op.add_column(
-            table_name, sa.Column("slug", sa.String(), nullable=True)
+            table_name,
+            sa.Column("slug", sa.String(), unique=True, nullable=True),
         )
 
         connection = op.get_bind()
--- a/backend/capellacollab/alembic/versions/2827788b8d2d_add_slugs_for_tools_and_versions.py
+++ b/backend/capellacollab/alembic/versions/2827788b8d2d_add_slugs_for_tools_and_versions.py
@@ -1,0 +1,40 @@
+# SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
+# SPDX-License-Identifier: Apache-2.0
+
+# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Slugs for tools and versions
+
+Revision ID: 2827788b8d2d
+Revises: f55b41e32223
+Create Date: 2023-11-07 11:15:07.753055
+
+"""
+import sqlalchemy as sa
+from alembic import op
+from slugify import slugify
+
+revision = "2827788b8d2d"
+down_revision = "f55b41e32223"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    for table_name in ("tools", "versions"):
+        op.add_column(
+            table_name, sa.Column("slug", sa.String(), nullable=True)
+        )
+
+        connection = op.get_bind()
+        rows = connection.execute(sa.text(f"SELECT * from {table_name}"))
+
+        for row in rows:
+            connection.execute(
+                sa.text(
+                    f"UPDATE {table_name} SET slug = '{slugify(row.name)}' WHERE id = {row.id};"
+                )
+            )
+
+        op.alter_column(table_name, "slug", nullable=False)
--- a/backend/capellacollab/sessions/exceptions.py
+++ b/backend/capellacollab/sessions/exceptions.py
@@ -6,6 +6,7 @@ import dataclasses
 import fastapi
 from fastapi import exception_handlers, status
 
+from capellacollab.projects.toolmodels import models as toolmodels_models
 from capellacollab.sessions import models as sessions_models
 from capellacollab.tools import models as tools_models
 
@@ -14,6 +15,11 @@ from capellacollab.tools import models as tools_models
 class UnsupportedSessionTypeError(Exception):
     tool: tools_models.DatabaseTool
     session_type: sessions_models.WorkspaceType
+
+
+@dataclasses.dataclass
+class MissingPrimaryGitModelError(Exception):
+    model: toolmodels_models.DatabaseCapellaModel
 
 
 async def unsupported_session_type_handler(
@@ -33,8 +39,29 @@ async def unsupported_session_type_handler(
     )
 
 
+async def missing_primary_git_model_error(
+    request: fastapi.Request, exc: MissingPrimaryGitModelError
+) -> fastapi.Response:
+    return await exception_handlers.http_exception_handler(
+        request,
+        fastapi.HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail={
+                "err_code": "MISSING_GIT_MODEL",
+                "reason": (
+                    f"The model {exc.model.name} doesn't support have a primary git model defined"
+                ),
+            },
+        ),
+    )
+
+
 def register_exceptions(app: fastapi.FastAPI):
     app.add_exception_handler(
         UnsupportedSessionTypeError,
         unsupported_session_type_handler,  # type: ignore[arg-type]
+    )
+    app.add_exception_handler(
+        MissingPrimaryGitModelError,
+        missing_primary_git_model_error,
     )
--- a/backend/capellacollab/sessions/exceptions.py
+++ b/backend/capellacollab/sessions/exceptions.py
@@ -6,7 +6,6 @@ import dataclasses
 import fastapi
 from fastapi import exception_handlers, status
 
-from capellacollab.projects.toolmodels import models as toolmodels_models
 from capellacollab.sessions import models as sessions_models
 from capellacollab.tools import models as tools_models
 
--- a/backend/capellacollab/sessions/exceptions.py
+++ b/backend/capellacollab/sessions/exceptions.py
@@ -17,11 +17,6 @@ class UnsupportedSessionTypeError(Exception):
     session_type: sessions_models.WorkspaceType
 
 
-@dataclasses.dataclass
-class MissingPrimaryGitModelError(Exception):
-    model: toolmodels_models.DatabaseCapellaModel
-
-
 async def unsupported_session_type_handler(
     request: fastapi.Request, exc: UnsupportedSessionTypeError
 ) -> fastapi.Response:
@@ -39,29 +34,8 @@ async def unsupported_session_type_handler(
     )
 
 
-async def missing_primary_git_model_error(
-    request: fastapi.Request, exc: MissingPrimaryGitModelError
-) -> fastapi.Response:
-    return await exception_handlers.http_exception_handler(
-        request,
-        fastapi.HTTPException(
-            status_code=status.HTTP_409_CONFLICT,
-            detail={
-                "err_code": "MISSING_GIT_MODEL",
-                "reason": (
-                    f"The model {exc.model.name} doesn't support have a primary git model defined"
-                ),
-            },
-        ),
-    )
-
-
 def register_exceptions(app: fastapi.FastAPI):
     app.add_exception_handler(
         UnsupportedSessionTypeError,
         unsupported_session_type_handler,  # type: ignore[arg-type]
-    )
-    app.add_exception_handler(
-        MissingPrimaryGitModelError,
-        missing_primary_git_model_error,
     )
--- a/backend/capellacollab/sessions/routes.py
+++ b/backend/capellacollab/sessions/routes.py
@@ -283,7 +283,6 @@ def request_persistent_session(
     ),
     db: orm.Session = fastapi.Depends(database.get_db),
     operator: k8s.KubernetesOperator = fastapi.Depends(operators.get_operator),
-    username: str = fastapi.Depends(auth_injectables.get_username),
 ):
     log.info("Starting persistent session for user %s", user.name)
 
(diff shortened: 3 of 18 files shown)


backend/capellacollab/tools/crud.py
dominik003, line 1: At least from my perspective, we should validate that the slug is unique for all tools and for all versions (otherwise it is not really a slug). This needs to be done for the `create_tool`, `update_tool`, `create_tool_version`, and `patch_tool_version` routes and can be done quite similar to the `create_project` and `update_project` routes.

backend/tests/sessions/test_sessions_routes.py
```diff
@@ -426,3 +436,78 @@ def test_create_persistent_jupyter_session(client, db, user, kubernetes):
...
+                },
+                {
+                    "model_slug": jupyter_model.slug,
+                },
+            ],
+            "persistent_workspace": True,
+        },
+    )
+    sessions = response.json()
+    print(sessions)
```
dominik003, line 503:
```suggestion
```

There is an additional `print` in the `test_create_read_only_session_as_user` where it would be great if you could remove this as well.

frontend/src/app/projects/project-detail/project-details.component.html
```diff
@@ -7,7 +7,14 @@
...
       <app-project-metadata class="flex-1"></app-project-metadata>
-      <app-create-readonly-session class="flex-1"></app-create-readonly-session>
+      <app-create-readonly-session
+        class="flex-1"
+        *ngIf="projectType === 'general'"
+      ></app-create-readonly-session>
+      <app-provision-workspace
+        class="flex-1"
+        *ngIf="projectType === 'training'"
+      ></app-provision-workspace>
```
dominik003, line 17:
```suggestion
      <app-create-readonly-session
        class="flex-1"
        *ngIf="(projectService.project$ | async)?.type === 'general'"
      ></app-create-readonly-session>
      <app-provision-workspace
        class="flex-1"
        *ngIf="(projectService.project$ | async)?.type === 'training'"
      ></app-provision-workspace>
```
 We can directly use the project observable via the `async` pipe here which safes us from using a separate variable in the `.ts` file and also significantly reduces the size of the `.ts` file:

```ts
/*
 * SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
 * SPDX-License-Identifier: Apache-2.0
 */

import { Component } from '@angular/core';
import { UntilDestroy } from '@ngneat/until-destroy';

import { ProjectUserService } from 'src/app/projects/project-detail/project-users/service/project-user.service';
import { ProjectService } from '../service/project.service';

@UntilDestroy()
@Component({
  selector: 'app-project-details',
  templateUrl: './project-details.component.html',
})
export class ProjectDetailsComponent {
  constructor(
    public projectService: ProjectService,
    public projectUserService: ProjectUserService,
  ) {}
}
```

We can also remove the `.css` file here as, at least from what I can see, the `title` id and the `metadata-item` class are not used in the html anymore.

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.css
dominik003, line 1: This file can be removed (with that PR here https://github.com/DSD-DBS/capella-collab-manager/pull/1183 `.css` and `.spec` files should also no longer be generated by default).

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.ts
```diff
@@ -0,0 +1,75 @@
...
+
+@UntilDestroy()
+@Component({
+  selector: 'app-provision-workspace',
+  templateUrl: './provision-workspace.component.html',
+  styleUrls: ['./provision-workspace.component.css'],
+})
+export class ProvisionWorkspaceComponent implements OnInit {
+  projectSlug?: string;
+  projectType?: string;
```
dominik003, line 27: Here I would either again use the `*ngIf="(projectService.project$ | async)?.type === 'training'"` in the template, or just store the whole project in here and use `project.type` and `project.slug`

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.ts
```diff
@@ -0,0 +1,75 @@
...
+              model_slug: m.slug,
+            };
+          }),
+          true,
+        )
+        .subscribe(() => {
+          this.router.navigateByUrl('/');
+        });
+    }
+  }
```
dominik003, line 74: We need some way of disabling the button after we clicked on it and before the actual subscription is resolved (i.e., directly after the if). Thats because right now, we can still click on the button until the subscribe is resolved which will result in many sessions to be spawned.

dominik003 (changes requested):
Thank you very much for this nice PR. There are a few comments from my side, hopefully all more or less understandable and sensible. 
Also, there are a few things in Sonar that should be addressed as well (some of them are somehow included in my suggestions). So it might make sense to go through the comments first and see if there are any outstanding points from Sonar at the end.

backend/capellacollab/alembic/versions/2827788b8d2d_add_slugs_for_tools_and_versions.py
```diff
@@ -0,0 +1,37 @@
...
+revision = "2827788b8d2d"
+down_revision = "5ca7037ef183"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    for table_name in ("tools", "versions"):
+        op.add_column(
+            table_name, sa.Column("slug", sa.String(), nullable=True)
```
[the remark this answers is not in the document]
amolenaar: Doesn't `unique` imply an index already?

backend/capellacollab/sessions/exceptions.py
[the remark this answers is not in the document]
amolenaar: I changed the code to record it as a warning. The exception is no longer needed.

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.ts
amolenaar: Good point. Fixed.

backend/capellacollab/sessions/routes.py
```diff
@@ -542,6 +544,159 @@ def create_database_and_guacamole_session(
...
+                "reason": "No models have been provided in this request for a read-only session.",
+            },
+        )
+
+    models_ = [
+        toolmodels_injectables.get_existing_capella_model(
+            m.toolmodel_slug, project, db
+        )
+        for m in body.models
+    ]
```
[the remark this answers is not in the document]
amolenaar: I moved the validation in a separate function.

backend/capellacollab/sessions/routes.py
```diff
@@ -542,6 +544,159 @@ def create_database_and_guacamole_session(
...
+        )
+        for m in models_
+    ):
+        raise fastapi.HTTPException(
+            status_code=status.HTTP_409_CONFLICT,
+            detail={
+                "err_code": "EXISTING_SESSION",
+                "reason": f"You already have a session for {project.name}. Close the existing session(s) before starting a new one.",
+            },
+        )
```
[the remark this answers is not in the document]
amolenaar: I added these checks to the validation function.

backend/capellacollab/sessions/routes.py
```diff
@@ -542,6 +566,149 @@ def create_database_and_guacamole_session(
     )
 
 
+@project_router.post(
```
MoritzWeber0, line 569: The new endpoint is good, however, we should merge it with the `read-only` endpoint. They have many lines in common and we should reduce code duplication as best as possible.

As described in https://github.com/DSD-DBS/capella-collab-manager/issues/1004, pre-provisioned sessions shall replace read-only sessions. `pre-provisioned` sessions have the same logic like read-only sessions. The only difference is that no persistent volume is passed.

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.html
```diff
@@ -0,0 +1,33 @@
+<!--
+ ~ SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+ ~ SPDX-License-Identifier: Apache-2.0
+ -->
+<div class="wrapper">
+  <h2 id="title">Provision workspace</h2>
+  <mat-card id="readonly-card">
```
dominik003, line 7:
```suggestion
  <h2>Provision workspace</h2>
  <mat-card>
```

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.html
```diff
@@ -0,0 +1,33 @@
...
+    <mat-card-content>
+      <p>Load the models in a training environment.</p>
+
+      <form [formGroup]="provisionForm" (ngSubmit)="requestSessions()">
+        <fieldset *ngIf="projectType !== 'training'">
+          <mat-slide-toggle formControlName="persistentWorkspace">
+            Use a persistent workspace</mat-slide-toggle
+          >
+        </fieldset>
```
dominik003, line 16:
```suggestion
        <div class="mb-4" *ngIf="projectType !== 'training'">
          <mat-slide-toggle formControlName="persistentWorkspace">
            Use a persistent workspace</mat-slide-toggle
          >
        </div>
```
Resolves the sonar cloud issue and, at least from what I can see, does not change any layouting.
Additionally, I moved the `mt-4` from the button fieldset to a `mb-4` here as we only need this additional margin if this slider exists.

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.html
```diff
@@ -0,0 +1,33 @@
...
+            matInput
+            mat-flat-button
+            color="primary"
+            type="submit"
+            [disabled]="!projectSlug || !models || sessionsRequested"
+          >
+            Start this training
+            <mat-icon>keyboard_arrow_right</mat-icon>
+          </button>
+        </fieldset>
```
dominik003, line 29:
```suggestion
        <button
          matInput
          mat-flat-button
          class="w-full"
          color="primary"
          type="submit"
          [disabled]="!projectSlug || !models || sessionsRequested"
        >
          Start this training
          <mat-icon>keyboard_arrow_right</mat-icon>
        </button>
```
Here we can completely remove the `fieldset`, again to fix the sonar cloud issue (we then need to add the `w-full` to handle the difference layouting.

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.html
dominik003, line 1: In addition to my below comments I have two other more general comments/questions:

1. Currently, this component is only shown in case `projectType === 'training'`. By the components name I suspect that we do want to either use this component somewhere else or also for general projects in the future?
2. Currently the button text is fixed to `Start this training`. In case we do want to also use the component for non training projects, we should make this text conditional on whether we have a training or general project.

dominik003 (commented):
Some last comments from my side and also saw one open comment from @MoritzWeber0 
Unfortunately, I am currently getting an error when building the backend image. Rebasing on main should resolve this. So after the rebase I will try out the functionality and when everything works fine, I'll approve it.

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.ts
amolenaar: I'll store the project then, since we need that when requesting a session anyway.

backend/capellacollab/sessions/routes.py
amolenaar: I like to keep them separate endpoints for now. This PR focuses on spawning training sessions. Doing a rewrite of read-only sessions will blow up the scope of this PR.

frontend/src/app/sessions/user-sessions-wrapper/create-session/provision-workspace/provision-workspace.component.html
amolenaar: I do not want to make to many _assumptions_ on what we're going to do with this component. It supports starting training sessions.

backend/capellacollab/sessions/routes.py
MoritzWeber0: I don't like the two endpoints due to the code duplication, which we try to avoid. The duplication increases future code maintainability and the effort to merge those is not too high. The provision endpoint is very similar to the `read-only` endpoint. The provision endpoint shall take a list of models instead of the project name anyway as decided in #1004. 

Given this, there is only one flag to add: `persistent`, which is a boolean that can be set to true or false. With this flag, we get rid of the code duplication and have one merged endpoint.

As this PR implements https://github.com/DSD-DBS/capella-collab-manager/issues/1004, which explicitly mentions that we want to merge both endpoints, I don't see it out of scope.

MoritzWeber0:
We'll wait for the web-based tool support, which has merged the persistent and read-only routes, but also prepared the interface for pre-provisioned workspaces. After https://github.com/DSD-DBS/capella-collab-manager/pull/1355 is merged, we can rebase this PR.

MoritzWeber0:
Due to the number of conflicts and the status of the PR, I will close the PR. It can still be used as inspiration for an implementation in the future.

Issue #1110, opened 2023-10-20
Improve UI in project and model creation workflow


MoritzWeber0:
The UI elements are not consistently used in the project and tool model creation workflow. 

Some examples: 

![image](https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/f722b6b5-914f-4151-be89-aa9885dd0675)
Here, the elements are aligned on the left side. The button is part of the mat-card. We decided to not have a "Cancel" button (navigation via breadcrumbs already provides this feature), so it should be removed.

![image](https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/f0e62986-5372-49e4-b017-f759873fae5e)
This looks weird as it doesn't use the whole available screen space, but the buttons to add users are still on the right side. Also the "skip" should start with an uppercase character and looks weird in the full-width mat-card.

![image](https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/c8fd68a0-dcfb-4d42-8d05-57cc40db2eeb)
The "Create model" button is on the left side, which is not consistent with the other button locations.
The "Skip model creation [...]" button has a different design again - it's placed in it's own mat-card, which is not full-width. The margins are not the same on the left & right side.

![image](https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/86365587-5af3-418c-a9dd-786c3e4a71ab)
The model sources selection is centered in the middle (other components sticked to the left side).

romeonicholas:
I'll share some updates with my justifications since I'm making up design choices as I go, in case there's anything anyone wants me to reconsider. 

**General Information Form Changes**
- Removed the Cancel button
- Left-aligned Create button
- Shortened the radio text. 
- Changed some HTML for accessibility (appropriate fieldsets and legends)

**General Information Questions**

- Can we shorten the radio text to "Public" and "Private" with an info hover (a user will only need to read the explanation once)
- Are we settled on "Internal"? (e.g. is it used that way throughout our projects?) I think "Internal" v "Private" doesn't communicate the difference very well
- I know the buttons aren't on the mat-card in other forms, but I think it actually needs to stay on the card for accessibility considerations, is that OK as long as I make it consistent throughout?

![Updated_ProjectCreation_GeneralInfo](https://github.com/DSD-DBS/capella-collab-manager/assets/79602065/7b377cfa-79fd-4262-8dc3-892285aaf993)

MoritzWeber0:
[...]

PR #1142, opened 2023-10-30
feat: Improve project creation UI

romeonicholas:
AddModelComparison
ConfigureModelComparison
AddTeamComparison
ChooseSourceComparison
GitComparison
CreateNewProjectComparison

Photos show the old views on the left and updated views on the right. These changes also includes the removal of the unused "model initialization" step. Resolves issue #1110

### DIFF
--- a/frontend/src/app/projects/create-project/create-project.component.html
+++ b/frontend/src/app/projects/create-project/create-project.component.html
@@ -3,14 +3,22 @@
  ~ SPDX-License-Identifier: Apache-2.0
  -->
 
-<div class="wrapper">
-  <mat-horizontal-stepper linear #stepper>
-    <mat-step completed="false" label="General information">
-      <div class="flex-center">
-        <mat-card class="section-card">
+<div class="wrapper flex justify-center">
+  <mat-horizontal-stepper
+    class="flex max-w-[360px] rounded-md shadow-md sm:max-w-[710px]"
+    linear
+    #stepper
+  >
+    <mat-step completed="false">
+      <ng-template matStepLabel>
+        <div class="hidden sm:inline">Create new project</div>
+      </ng-template>
+      <h3 class="flex justify-center sm:hidden">Create new project</h3>
+      <div class="flex justify-center">
+        <mat-card class="w-[350px] sm:min-w-[420px]">
           <form [formGroup]="form">
-            <fieldset>
-              <mat-form-field appearance="fill">
+            <div>
+              <mat-form-field appearance="fill" class="form-field-default">
                 <mat-label>Name</mat-label>
                 <input
                   matInput
@@ -26,9 +34,9 @@
                   A project name is required.
                 </mat-error>
               </mat-form-field>
-            </fieldset>
-            <fieldset>
-              <mat-form-field appearance="fill">
+            </div>
+            <div>
+              <mat-form-field appearance="fill" class="form-field-default">
                 <mat-label>Description</mat-label>
                 <textarea
                   matInput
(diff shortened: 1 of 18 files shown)


frontend/src/styles.css
```diff
@@ -176,27 +171,58 @@ fieldset > :not(.field-separator) {
...
+  width: 680px;
+}
+
+.same-line-input {
+  max-width: 185px;
+}
+.nextline-stepper .mat-step-label.mat-step-label {
+  text-align: center;
+  text-overflow: inherit;
+  white-space: normal;
```
MoritzWeber0, line 198: We try to avoid global CSS classes where possible. Instead, use TailwindCSS.

The reasons are: 
- When code in the HTML templates are removed, CSS classes are often forgotten. 
- It's hard to keep an overview over available CSS classes. Everybody will define their own classes, which is against the concept of consistent and aligned classes.
- Changes to a global class can introduce several side effects.
- We don't have to invent classnames that nobody understands.

A page that is worth to read: https://tailwindcss.com/docs/utility-first

TailwindCSS has the nice concept of custom styles: https://tailwindcss.com/docs/adding-custom-styles, which can be used for the definition of a specific width. If it doesn't make sense to use TailwindCSS, prefer to use component scoped CSS classes.

MoritzWeber0 (changes requested):

frontend/src/styles.css
romeonicholas: I'll move the lesser-used classes to the inline utility styles, but the form styling would probably be a mistake to split up--it's used across 9 forms in 9 different files which would ideally be templated/compontentized but would probably require more effort than it's worth. We can also talk about this at the next weekly, but honestly, I'm concerned the current approach is causing more problems than it's solving (that is, without a visual design spec you'll contniue to see more inconsistancies in styling with utility styles than with reusable classes).

romeonicholas: OK--shifted everything inline except for the form class that gets used across so many files. I can move that inline if you prefer, but I think the class is more maintainable given the project creation flow will be updated/added to at some point and in the absense of a component the various forms are derived from it makes it easier to find them all. Or I can try to componentize everything, but that probably deserves a separate issue because it'd be structurally changing multiple parts of the site.

MoritzWeber0: > We can also talk about this at the next weekly, but honestly, I'm concerned the current approach is causing more problems than it's solving (that is, without a visual design spec you'll contniue to see more inconsistancies in styling with utility styles than with reusable classes).

This might be more suited for discussion during the Collaboration Manager planning, but I can offer some insights based on our experience. We initially adopted both global and component-based styles, but this approach wasn't effective for us. The primary issue was the continuous addition of new classes, while the existing ones were seldom reused.

Style consistencies have never been a big priority and will continue to be less relevant than functionality and stability. In order to really have a consistent style, we need someone who has the time to actively enforce a consistent style. Regardless of whether we use inline or global styles. But we don't have the capacity for that, and I'm also worried that it could slow us down in the development process.

frontend/src/app/projects/create-project/create-project.component.html
```diff
@@ -3,14 +3,14 @@
...
 
-<div class="wrapper">
-  <mat-horizontal-stepper linear #stepper>
-    <mat-step completed="false" label="General information">
-      <div class="flex-center">
-        <mat-card class="section-card">
+<div class="wrapper flex justify-center">
+  <mat-horizontal-stepper class="w-[710px]" linear #stepper>
+    <mat-step completed="false" label="Create new project">
+      <div class="flex justify-center">
```
MoritzWeber0, line 9: The background color seems to be slightly different between global background and the div: 
image

Let's have a quick chat about this after the Daily. I'd like to hear your opinion if we should use the same background color.

frontend/src/app/projects/models/create-model/create-model.component.html
```diff
@@ -5,54 +5,55 @@
...
     linear
     #stepper
   >
-    <mat-step completed="false" label="Create model">
-      <app-create-model-base
-        [asStepper]="asStepper"
-        (create)="afterModelCreated()"
-      ></app-create-model-base>
-    </mat-step>
+    <div class="flex justify-center">
```
MoritzWeber0, line 13: The component is not centered when triggered via the project details: 
image

image

frontend/src/app/projects/models/model-source/choose-source.component.html
```diff
@@ -2,48 +2,65 @@
...
-        <app-mat-icon size="50px">link</app-mat-icon>
-      </mat-card>
-    </a>
-  </div>
+<div class="flex justify-center">
+  <mat-card class="mat-card-form" *ngIf="projectService.project$ | async">
+    <div
+      *ngIf="(projectService.project$ | async) && (modelService.model$ | async)"
+    >
+      <div
```
MoritzWeber0, line 12: Git sources should also be available for non-admins and Git sources don't require the T4C integration.

frontend/src/app/projects/models/model-source/choose-source.component.html
```diff
@@ -2,48 +2,65 @@
...
-        <p>Skip this step</p>
-        <app-mat-icon size="50px">skip_next</app-mat-icon>
-      </mat-card>
-    </a>
-  </div>
+      <h3>Other</h3>
+      <mat-divider></mat-divider>
+      <div class="my-[10px]">
+        <a mat-button (click)="modelSourceSelection.emit('skip')">
+          <span> Manually configure model </span>
```
MoritzWeber0, line 60: `Manually configure model` isn't the right term for it. The button can be used to skip adding the model sources. Maybe something like: "Add the Model source later".

frontend/src/app/projects/create-project/create-project.component.html
```diff
@@ -3,14 +3,14 @@
  ~ SPDX-License-Identifier: Apache-2.0
  -->
 
-<div class="wrapper">
-  <mat-horizontal-stepper linear #stepper>
-    <mat-step completed="false" label="General information">
-      <div class="flex-center">
-        <mat-card class="section-card">
+<div class="wrapper flex justify-center">
```
MoritzWeber0, line 6: The mobile view has room for improvement. It's wasn't perfect on main either. To do so, you can add a few `max-w-mobile md:w-full` classes and add `mobile` to the `spacing` section of the `tailwind.config.js`. 
image

MoritzWeber0 (changes requested):

frontend/src/app/projects/create-project/create-project.component.html
romeonicholas: Added shadow and rounded corners to stepper:
ShadedRoundedStepper

frontend/src/app/projects/models/model-source/choose-source.component.html
romeonicholas: Updated:
ModelSourceSkipping

frontend/src/app/projects/create-project/create-project.component.html
romeonicholas: This one I'm kind of at a loss for. I don't think there's any way to continue using the horizontal stepper on mobile, so maybe I can flip it into a vertical stepper at a certain breakpoint? Removing the back buttons made sense on desktop but they're really important for mobile, but since we sometimes skip steps I don't know if that would require special logic, or do we leave them off?

frontend/src/app/projects/models/model-source/choose-source.component.html
MoritzWeber0: Yes :)

frontend/src/app/projects/create-project/create-project.component.html
romeonicholas: Vertical also didn't look great, so I opted for horizontal with hidden labels and conditionally displayed headers so the user knew what step they were on. There's a weird media query for the stepper-lines and stepper headers, but apparenlty you can only target that with CSS, unless you know a workaround.
mobile

MoritzWeber0: I really like the new mobile design, thanks for the implementation and the idea 👍

frontend/src/app/projects/models/create-model/create-model.component.html
MoritzWeber0: This change got lost with the last push, the model creation dialog has no shadow / rounded anymore: 

image


Also, the width of model description and model tool is not full anymore.

frontend/src/styles.css
MoritzWeber0, line 1: This comment is not related to this file, I just need a file to open a conversation.

In the documentation, we have to update the "Create a Model" page and remove Step 4 (Choose Initialization).

frontend/src/app/projects/models/create-model/create-model.component.html
romeonicholas: Waah, replace-all error (the styling was actually still there but set to be sm instead of md), now fixed. Full-width applied to inputs again.

frontend/src/styles.css
romeonicholas: Removed :white_check_mark:

MoritzWeber0 (approved):
Thanks for the great PR. A very nice UI improvement and nice improvements for mobile devices :)

PR #1149, opened 2023-10-31
feat: Add session cost monitoring

MoritzWeber0:
The resource costs are defined as yml configuration.
We use the `monaco` editor in the frontend.


MoritzWeber0:
This PR didn't contain many cost monitoring specific code, it mainly added support for a global YAML configuration. Since this was introduced in https://github.com/DSD-DBS/capella-collab-manager/pull/1168, I'll close this PR. 
It's easier to recreate it based on the improvements on main.

PR #1150, opened 2023-10-31
feat: Add tiling window manager for training mode

### LINKED ISSUE #982: Dynamically rebalance sessions in training mode
In the session viewer, when two sessions are open side-by-side, users should have the ability to dynamically resize or rebalance the screen space allocated to each session. This feature could be implemented by incorporating a slider between the two sessions, allowing users to easily move it left or right to adjust the balance between the sessions. A challenge to address is how to handle scenarios where the slider is moved entirely to one side, effectively allocating all the available space to a single session. A potential solution could be to automatically collapse the session when its size becomes too small (see #985 for collapsing of sessions).

### LINKED ISSUE #985: Focus sessions in training mode
When two sessions are open in the training mode, users should have the option to collapse one session, allowing the other to occupy the entire screen, effectively creating a "focus mode." The collapsed session could remain accessible through a minimal sidebar, which when clicked, would uncollapse and restore the session to its original size.

### LINKED ISSUE #990: Add new tiling window manager layout for training mode
Comparable to the "Session viewer", which has a "floating window" layout, the "Training mode" should have "tiling window" layout. A new Angular component has to be created for it.

The advantage of the tiling window manager is: 
- We can pre-defined a fixed layout, which is optimized for a specific training. 
- The set of sessions usually doesn't change during the training, there is no disturbing relayouting. 
-  The full screen size is used.

dominik003:
Resolves #990, Resolves #982, Resolves #985

Things still needed before merge:
- [x] trigger session resizing for session requirings this (e.g., capella)
- [x] add focus on the current session (required to have the keyboard input switched)
- [x] add possibility to set a session into fullscreen
- [x] removed filter from unfocuse session and execute "focusSession" only when session is not currently focused


frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.html
```diff
@@ -0,0 +1,75 @@
...
+      <div
+        class="flex items-center justify-between gap-2 rounded-t bg-slate-100 p-1"
+      >
+        <span>
+          {{ session.version?.tool?.name }} {{ session.version?.name }},
+          {{ session.type }}
+          <span *ngIf="session.type === 'readonly'"
+            >(project {{ session.project!.name }})</span
+          >
+          {{ session.id }}
```
MoritzWeber0, line 25: I wouldn't show the session ID. The ID is mainly used internally and the combination (tool, version, type, project?) is unique.

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.ts
```diff
@@ -0,0 +1,199 @@
...
+  public resizeState: ResizeState = {};
+
+  private minimumSessionWidth = 0;
+  private existingMargin = 0;
+
+  constructor(public fullscreenService: FullscreenService) {}
+
+  ngOnInit(): void {
+    this.fullscreenService.isFullscreen$.subscribe((isFullscreen) => {
+      this.existingMargin = isFullscreen ? 0 : 27.48;
```
MoritzWeber0, line 42: Where does the margin of `27.48` come from?

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.ts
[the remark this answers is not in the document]
MoritzWeber0: Two Jupyter session work, but Capella session don't.

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.resizeState = {};
+    this.toggleIFrameEvents(true);
+  }
+
+  @HostListener('window:resize')
+  onResize() {
+    this.resetWidths();
+  }
+
+  onLeftClick(session: TilingSession) {
```
MoritzWeber0, line 89: We should use another name for it. "LeftClick" is for me the "left" mouse button and "RightClick" the right mouse button.

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.html
```diff
@@ -0,0 +1,75 @@
...
+          </mat-icon>
+        </div>
+      </div>
+
+      <div class="relative grow">
+        <div
+          class="iframe-overlay"
+          *ngIf="isValidResizeState(this.resizeState)"
+        ></div>
+        <iframe
```
MoritzWeber0, line 52: We should disable the "selection" of the whole iframe (if possible). This happens sometimes accidentally. 

https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/b11ae22c-465b-4f6a-acfc-c9c3416fd3ac

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.html
```diff
@@ -0,0 +1,75 @@
...
+            >(project {{ session.project!.name }})</span
+          >
+          {{ session.id }}
+        </span>
+
+        <div class="flex items-center">
+          <mat-icon
+            *ngIf="session.index !== 0"
+            class="cursor-pointer text-lg text-gray-600 hover:text-gray-800"
+            (click)="onLeftClick(session)"
```
MoritzWeber0, line 32: https://github.com/DSD-DBS/capella-collab-manager/assets/23395732/566e3ce4-a722-4e7d-9337-50b99ccb68b2

In the video you can see that the session "jumps". This is the case because the session keep their sizes. I'm not sure if we should keep the original layout and just switch the sessions in the existing "boxes".

MoritzWeber0 (commented):
Great job, the slider works really good.

frontend/src/app/sessions/session/floating-window-manager/floating-window-manager.component.html
```diff
@@ -0,0 +1,66 @@
+<!--
```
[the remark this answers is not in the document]
dominik003: For some reason I removed the `z-index: 40` from the `iframe-overlay` class and also wrongly applied the class. It should now properly work for both the floating and tiling window manager.

frontend/src/app/helpers/default-value.pipe.ts
```diff
@@ -0,0 +1,15 @@
...
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'defaultValue',
+})
+export class DefaultValuePipe implements PipeTransform {
+  transform(value: string | undefined, defaultValue: string): string {
+    return value ?? defaultValue;
+  }
+}
```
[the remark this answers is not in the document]
dominik003: I'm not quite sure what the problem was when I tried it this way, but it definitely showed me a weird error message that I couldn't get rid of. However, I tried again and it works fine now (so I must have done something wrong before).

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.html
dominik003: Yes, totally agree. This was just some remaining "debug" code as I tested it with two Jupyter sessions and needed a way to distinguish both.

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.ts
dominik003: From inspection to my browser (left and right margins when not being in full screen mode). But I agree that it's not a good idea to just put it as it's there. However, I have tested whether or not having it makes a difference that doesn't. So I just remove the entire `existingMargin`.

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.ts
dominik003: Good point! I renamed it to `onLeftArrowClick` and `onRightArrowClick` which should be more clear in this context.

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.html
dominik003: I added a `class="select-none"` to the `div` containing the floating and tiling window manager, so now selection should no longer be possible.

frontend/src/app/sessions/session/tiling-window-manager/tiling-window-manager.component.ts
```diff
@@ -0,0 +1,199 @@
+/*
```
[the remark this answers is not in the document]
dominik003: Focusing sessions is now included in both the floating and tiling window manager. Therefore, keyboard handling should work as before.

MoritzWeber0 (approved):
LGTM; good improvement. Will move us forward on the journey towards the training mode.

PR #1152, opened 2023-11-01
docs(dev): Add requirements for responsive design

MoritzWeber0:


romeonicholas (commented):
My biggest concern is that the specific information a developer would need is buried or unclear/missing. It would be perfectly fine as is, but from a general technical writing and new-to-this-project dev standpoint I would suggest:
- Removing or shortening the first section - the requirement **just is** that everything is developed for mobile, no need to justify or list things
- Removing or lowering the section on debugging and testing - this information is widely available and not specific to the project
- Using the best practices section to link out to best practices specificaly for Angular and Tailwind around responsive design rather than listing general practices
- Converting the supported devices and responsive guidelines section into more of a information sheet that clearly highlights the expectations for: minimum width, breakpoints, maximum width, minimum font size, any other hard requirements
- Potentially including a simple example present in the site

MoritzWeber0:
Thanks for the detailed feedback.

> * Removing or shortening the first section - the requirement **just is** that everything is developed for mobile, no need to justify or list things

Done.

> * Removing or lowering the section on debugging and testing - this information is widely available and not specific to the project

I think it's good as quick start as some of our interns didn't know of the dev tools yet. It's more of an hint that they exist and if they need more information, Google can help out. I moved it down the list. 

> * Using the best practices section to link out to best practices specificaly for Angular and Tailwind around responsive design rather than listing general practices

There is not much Angular specific about responsive design. At least I don't have anything specific in mind.
I added a section about Tailwind design and changed the Best Practices section to mention the corresponding Tailwind classes instead of general CSS classes.

> * Converting the supported devices and responsive guidelines section into more of a information sheet that clearly highlights the expectations for: minimum width, breakpoints, maximum width, minimum font size, any other hard requirements

I added smallest width & height and the breakpoint. There is no minimum font size as of now. 

> * Potentially including a simple example present in the site

The navigation bar is a good example (The menu is different on mobile and desktop). I'll check if I can make a good example out of it.

romeonicholas:
Updates look good 👍

PR #1155, opened 2023-11-02
fix: Restrict clickable area on mat-card overviews

MoritzWeber0:
The clickable area in the mat-card overviews was slightly larger than the mat-card. Reason is that that the mat-card element has a margin of 10px and is wrapped by an a-element. The margin is now "moved" to the a-element.

### DIFF
--- a/frontend/src/app/projects/project-overview/project-overview.component.html
+++ b/frontend/src/app/projects/project-overview/project-overview.component.html
@@ -4,8 +4,8 @@
  -->
 
 <div class="flex flex-wrap">
-  <a routerLink="/projects/create">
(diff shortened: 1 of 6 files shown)


romeonicholas (changes requested):
Looks good--very straightforward solution and it worked for me locally. I think you may just be missing the base settings page (settings.component.html) which also uses these cards with oversized clickable areas.

MoritzWeber0:
> Looks good--very straightforward solution and it worked for me locally. I think you may just be missing the base settings page (settings.component.html) which also uses these cards with oversized clickable areas.

Good catch! `settings.component.html` is also updated now.

romeonicholas (approved):
👍

Issue #1158, opened 2023-11-02
Add abstraction layer for web/http-based tools


MoritzWeber0:
We do already support Jupyter notebooks, which exposes the application via http. The infrastructure to deploy web-applications as tools is there, but most of the options are hard-coded to work with Jupyter. Instead, the system administrators should be able to configure these options.

When this is done, we can support nearly all web-applications as tools in the Collaboration Manager.

To accomplish this, the tools configuration has to be extended. New keys are: 
- `connection-type`, can be rdp or http. RDP is the default or HTTP if the Jupyter integration is enabled for the tool.
- `resources`: Define individual resources for sessions. (See example below)
- `version-independent`: Defines if a session is version independent. A tool-session with independent versions is a tool-session that can provision models of all available versions of the tool.

The following options are only available for sessions with connection type http:
- `input`: Define input values. For now, only the type environment will be supported.
- `authentication`: Handling of authentication. Only type token should be allowed for the beginning.  Type token injects a token via environment variable in the session and passes it via a param to the URL.
- `ports`: Define the internal port. The port will be mapped and exposed to 443.

The configuration should be validated with pydantic. One example of a configuration is:

```yml
[...]

- image: registry:12345/capella-model-editor:vX.X.X
    id: cme
    name: Capella model editor
    resources:
      cpu:
        requests: [...]
        limit: [...]
      memory:
        requests: [...]
        limit: [...]
    version-independent: true
    input:
      git:
        type: environment # or capellambse for capellambse CLI format
        variable: CCM_GIT_MODELS
      base-url:
        type: environment
        variable: CCM_BASE_URL
      origin-host:
        type: environment
        variable: CSP_ORIGIN_HOST
    authentication: # Only supported for http
      type: token # Inject token via environment variable and pass via URL
      mount:
        type: environment
        variable: CCM_TOKEN
      url-param: 'token'
    connection-type: http
    ports: 
      - 8080 # Internal port the application runs on, will be exposed via https (443)
    primary: True # Fallback if no ID is passed with the request (downwards compatible).
  - image: registry:12345/capella/read-only:$version-vX.X.X-vX.X.X
    id: capella-native
    name: Native Capella
    resources: [...]
    version-independent: false # The editor can start all Capella versions, is not bound to a specific version.
    connection-type: rdp
[...]
```

This shall be no breaking change! Existing defined tools should me migrated to the new configuration.

Requires https://github.com/DSD-DBS/capella-collab-manager/issues/1156

PR #1162, opened 2023-11-02
feat: Increase resources for Jupyter sessions

MoritzWeber0:
In our environments, the resources were too low and the IPython kernel crashed a few times.

This is just a temporary fix until resources are properly configurable: https://github.com/DSD-DBS/capella-collab-manager/issues/1158


amolenaar (approved):
LGTM

It's quite a lot though, 3Gb for Jupyter.

PR #1169, opened 2023-11-06
build: No longer require `-j4` in `make dev`

MoritzWeber0:
Developers can simply run `make dev` instead of `make -j4 dev`. Since the target does only work with parallel execution endabled, the parallel execution is now the default.


amolenaar (approved):
Nice one. Simple and elegant.

PR #1178, opened 2023-11-07
Fix t4c repo validation

### LINKED ISSUE #1170: TeamForCapella repository validation missing in backend & frontend workflow
When using "Create a new TeamForCapella repository" in the model sources or in the model creation workflow, it's possible to use TeamForCapella repository names with whitespaces. This produced some errors in the backend: 

image

Here, I can enter any name: 
image

In the T4C server settings, the validation is applied properly:
image

In addition to the missing frontend validation, we need a backend validation for it ("never trust the frontend"). We can use the pydantic Field attribute "regex" for it. 

I tried to run it locally (without a running T4C server) and there was no warning that the creation failed partially (it was only created in our database, not on the T4C server). A proper creation requires manual intervention by an administrator.  
image
We could add a message similar to "Repository deletion failed partially".

dominik003:
Resolved #1170


backend/capellacollab/settings/modelsources/t4c/repositories/models.py
```diff
@@ -44,7 +44,7 @@ class DatabaseT4CRepository(database.Base):
 
 
 class CreateT4CRepository(pydantic.BaseModel):
-    name: str
+    name: str = pydantic.Field(pattern="/^[-a-zA-Z0-9_]+$/")
```
MoritzWeber0, line 47:
```suggestion
    name: str = pydantic.Field(pattern="^[-a-zA-Z0-9_]+$")
```

I assume that this a copy&paste error from the JavaScript regex pattern. `/` in the Python `re`-module matches the character `/`.

MoritzWeber0 (approved):

PR #1182, opened 2023-11-09
fix: Enable default link behaviors on nav bar and side nav links

### LINKED ISSUE #1164: Cannot middle-click links at the top of the page
When middle-clicking one of the links at the top (Projects, Sessions, Session Viewer, etc.), nothing happens. Middle-clicking should open the link in a new tab, just like every other link.

romeonicholas:
Fixes #1164 

Moving routerLinks out of the containing divs and into the anchors for nav bar and side nav links makes them behave like normal links (middle click to open new tab, right click to open context menu, etc.)

### DIFF
--- a/frontend/src/app/general/header/header.component.html
+++ b/frontend/src/app/general/header/header.component.html
@@ -21,19 +21,25 @@
     </div>
 
     <div class="hidden gap-2 xl:flex">
-      <div
-        *ngFor="let item of navBarService.navBarItems"
-        [routerLink]="item.routerLink"
-      >
-        <a
-          mat-raised-button
-          color="primary"
-          *ngIf="userService.validateUserRole(item.requiredRole)"
-          [attr.href]="item.href"
-          [attr.target]="item.target"
-        >
-          {{ item.name }} <mat-icon *ngIf="item.icon">{{ item.icon }}</mat-icon>
-        </a>
+      <div *ngFor="let item of navBarService.navBarItems">
+        <ng-container *ngIf="userService.validateUserRole(item.requiredRole)">
+          <ng-container *ngIf="item.href; else router">
+            <a
+              mat-raised-button
+              color="primary"
+              [attr.href]="item.href"
+              [attr.target]="item.target"
+            >
+              {{ item.name }}
+              <mat-icon *ngIf="item.icon">{{ item.icon }}</mat-icon>
+            </a>
+          </ng-container>
+          <ng-template #router>
+            <a mat-raised-button color="primary" [routerLink]="item.routerLink">
+              {{ item.name }}
+            </a>
+          </ng-template>
+        </ng-container>
       </div>
     </div>
     <div>
--- a/frontend/src/app/general/nav-bar-menu/nav-bar-menu.component.html
+++ b/frontend/src/app/general/nav-bar-menu/nav-bar-menu.component.html
@@ -5,20 +5,30 @@
 
 <mat-list>
   <div *ngFor="let item of navBarService.navBarItems">
-    <div [routerLink]="item.routerLink">
-      <a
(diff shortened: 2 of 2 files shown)


frontend/src/app/general/header/header.component.html
```diff
@@ -21,16 +21,14 @@
...
-        [routerLink]="item.routerLink"
-      >
+      <div *ngFor="let item of navBarService.navBarItems">
         <a
           mat-raised-button
           color="primary"
           *ngIf="userService.validateUserRole(item.requiredRole)"
           [attr.href]="item.href"
           [attr.target]="item.target"
+          [routerLink]="item.routerLink"
```
MoritzWeber0, line 31: The `href`-URLs (Grafana, Prometheus, Documentation) stop working with this change. This was also the reason why the routerLink was on the div element, when I remember it correctly.

In the worst case, we have to add conditional `a`-tags for routerLink & href.

frontend/src/app/general/nav-bar-menu/nav-bar-menu.component.html
```diff
@@ -5,13 +5,14 @@
 
 <mat-list>
   <div *ngFor="let item of navBarService.navBarItems">
-    <div [routerLink]="item.routerLink">
+    <div>
```
MoritzWeber0, line 8: We should be able to remove this intermedia div.

MoritzWeber0 (changes requested):

frontend/src/app/general/header/header.component.html
romeonicholas: For some reason Angular won't let me conditionally bind href and target attributes when they exist and leave off routerLink (or vice versa), but because you'll always need to group same-page and new-page links to keep users sane I split them into two arrays with their own template logic. This is a little bit of duplication in the HTML but it's more explicit than it was previously (e.g. for each new page link add a new page icon instead of adding item.icon which was just always new page)

frontend/src/app/general/nav-bar-menu/nav-bar-menu.component.html
romeonicholas: Ah right, think I got rid of any empty tags this time

frontend/src/app/general/header/header.component.html
MoritzWeber0: Yes, the additional HTML is fine (I don't think we'll find a way around it anyway). 

I'd prefer one array over two arrays. You can then use a simple condition with ngIf to decide which HTML element to use. This would also allow to have a mixed order (same page, new page, same page, ...) in the future.

romeonicholas: I find this harder to read and maintain, but I already shared my general concerns with the nav, so done ✔️

MoritzWeber0: It will look better with the new control flow, but it's not there yet: 
image

frontend/angular.json
```diff
@@ -8,13 +8,6 @@
       "schematics": {
         "@schematics/angular:application": {
           "strict": true
-        },
-        "@schematics/angular:component": {
-          "skipTests": true,
-          "style": "none"
-        },
-        "@schematics/angular:service": {
-          "skipTests": true
```
MoritzWeber0, line 17: These lines should not be removed, might be a rebase error.

romeonicholas: Blargh, correct, I'm still getting used to the single commit PR, reverted this file

MoritzWeber0: You don't have to use single commit PRs, I'll also merge multi-commit PRs. Do what you think is best for you.

MoritzWeber0 (approved):

PR #1195, opened 2023-11-20
build: Update Angular Material

### LINKED ISSUE #1177: Remove deprecated/legacy angular material components
With Angular Material UI version 15, some fundamental changes were made by switching to MDC-based components. This change led to some erroneous UI elements (misplacement, wrong styling, ...). For this reason, we have switched most of the affected components to their "legacy" variant. In recent months, we have already reduced the dependency on these components or switched to the new components. However, there are still places where we use the legacy components. As Angular 17 will be released in the second week of November, we should consider a full migration from the legacy to the new MDC-based components. This migration will require replacing all remaining legacy components and fixing any UI/styling issues that arise in the process (I suggest doing this step by step to ensure that all components still work as expected).The official [MDC migration guide](https://material.angular.io/guide/mdc-migration) might be helpful for this issue.

Below you will find a list of the locations and components that need to be treated:

Inside the `custom-theme.scss`:
- [ ] `@include mat.all-legacy-component-typographies();`
- [ ] `@include mat.legacy-core();`
- [ ] `@include mat.all-legacy-component-themes($theme);`

Affected Modules inside the `app.module.ts` (Legacy module is always named `MatLegacy<...>`):
- [ ] `MatAutocompleteModule` 
- [ ] `MatButtonModule`
- [ ] `MatCardModule`
- [ ] `MatCheckboxModule`
- [ ] `MatFormFieldModule`
- [ ] `MatInputModule`
- [ ] `MatListModule`
- [ ] `MatMenuModule`
- [ ] `MatPaginatorModule`
- [ ] `MatProgressBarModule`
- [ ] `MatProgressSpinnerModule`
- [ ] `MatSelectModule`
- [ ] `MatSlideToggleModule`
- [ ] `MatTableModule`
- [ ] `MatTabsModule`
- [ ] `MatTooltipModule`

Affected Components (Legacy component is always named `MatLegacy<...>`):
- [ ] `MatPaginator`
- [ ] `MatTableDataSource`
- [ ] `MatSelectionList`
- [ ] `MatSelectionListChange`

Paula-Kli:
closes #1177 

things to still look into:

- [x] Search field in Project User
- [x] find out why matcard background color cannot be set with tailwind (model-overview component)
- [x] align deleteProject Button
- [x] find out tailwind styling in `sessions.component`
- [x] align "Open Github Issue` button
- [x] File Browser button upper right
- [x] Sync button in T4C instance settings
- [x] check long model names whether the badges overflow
- [x] add model button in project overview
- [x] User menu
- [x] Look into different fonts


frontend/src/styles.css
```diff
@@ -244,6 +259,11 @@ Angular Material Card Styles
   max-width: 85vw;
 }
 
+/* Added due to vertical line in mat-form-field - for more information read: https://github.com/tailwindlabs/tailwindcss/discussions/9993*/
```
MoritzWeber0, line 262:
```suggestion
/* Added due to vertical line in mat-form-field - for more information read: https://github.com/tailwindlabs/tailwindcss/discussions/9993 */
```

frontend/src/styles.css
```diff
@@ -185,6 +191,15 @@ fieldset > :not(.field-separator) {
   margin: auto;
 }
 
+/*
+Own collab manager card styles
+*/
+@layer components {
```
MoritzWeber0, line 197: Is there a reason for using the `layer` CSS feature? We don't use it anywhere else, as far as I know. 
Given that there is only one layer, it doesn't seem to have any effect.

frontend/src/app/users/basic-auth-token/basic-auth-token.component.html
```diff
@@ -56,26 +56,30 @@ <h2 class="!mb-2">Personal Access Tokens</h2>
...
-  <ng-container *ngIf="(tokenService.tokens$ | async)?.length; else elseBlock">
-    <mat-card *ngFor="let token of tokenService.tokens$ | async" class="!ml-0">
-      <mat-card-content>
-        Description: {{ token.description }} <br />
-        Expiration date: {{ token.expiration_date | date }} <br />
-        Creation location: {{ token.source }}
-        <div *ngIf="isTokenExpired(token.expiration_date)">
-          <mat-icon color="warn" class="align-middle">warning</mat-icon>
-          <span class="text-red-600">This token has expired!</span>
+  @if ((tokenService.tokens$ | async)?.length) {
```
MoritzWeber0, line 64: Nice!

frontend/src/index.html
```diff
@@ -21,8 +21,7 @@
       rel="stylesheet"
     />
   </head>
-  <!-- eslint-disable-next-line tailwindcss/no-custom-classname -->
-  <body class="mat-typography">
+  <body>
     <app-root></app-root>
   </body>
 </html>
```
MoritzWeber0, line 27: There is still a lot to do. Take the version before the update into account. For example, compare the rendered frontend version of this branch to one of our production versions.

You will see many differences, some of them are small, that we need to change.

I'll only list some examples: 

- [ ] The global font size is larger than before.
- [ ] The text on the button has a lot of space between characters.
- [ ] Bold texts are not bold anymore.
- [ ] The "Persistent workspace session" card is wider than before, it takes the whole available space.
- [ ] In the "Menu" dropdown menu, there is a non-clickable right padding (should be moved to the buttons to make it clickable)
- [ ] "Welcome to the Capella Collaboration Manager" is not centered vertically anymore, also the font size is different.

image

The same does also apply to other components.

MoritzWeber0 (changes requested):

frontend/src/styles.css
```diff
@@ -5,10 +5,16 @@
 
 /* You can add global styles to this file, and also import other style files */
 
+@import "~@angular/material/prebuilt-themes/indigo-pink.css";
```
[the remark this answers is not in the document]
Paula-Kli: no reason that the pink theme is needed but we seem to need a theme
when updating I had this issue when I did not yet have a theme: 
Bildschirmfoto 2024-01-23 um 10 06 10
or e.g. without:
Bildschirmfoto 2024-01-23 um 10 12 09
with the theme:
Bildschirmfoto 2024-01-23 um 10 11 32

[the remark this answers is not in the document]
MoritzWeber0: Right, `indigo` is the primary color of the theme, which should be good. We can keep it.

frontend/src/app/sessions/sessions.component.html
```diff
@@ -3,14 +3,16 @@
...
 
 <app-user-sessions-wrapper>
   <div class="flex flex-wrap justify-around gap-5">
-    <app-create-persistent-session></app-create-persistent-session>
-    <app-active-sessions></app-active-sessions>
+    <app-create-persistent-session
+      class="shrink grow-[2] basis-0"
+    ></app-create-persistent-session>
+    <app-active-sessions class="shrink grow basis-0"></app-active-sessions>
   </div>
```
romeonicholas, line 17: Popping in at Paula's request (hi Paula 👋). I didn't check the Angular defaults to confirm it, but the Persistent Workspace Session card is 450px wide at all window widths on production so I'd just add that directly and let flex handle the spacing but not the size. I also added that width to Active Sessions because I don't know why they'd need to be different widths, though the contents of that card need ot be adjusted to use the full width (probably just a w-full on the card but I can look into it if you like).

```
<app-user-sessions-wrapper>
  <div class="flex flex-wrap justify-around">
    <app-create-persistent-session
      class="w-[450px]"
    ></app-create-persistent-session>
    <app-active-sessions class="w-[450px]"></app-active-sessions>
  </div>
</app-user-sessions-wrapper>
```
SessionWidths

Paula-Kli: thanks a lot!!

Paula-Kli: Bildschirmfoto 2024-01-23 um 12 55 02
what would you adjust?

frontend/src/app/general/header/header.component.html
```diff
@@ -21,85 +21,89 @@
...
+      </a>
+
+      <mat-menu #profileMenu="matMenu" class="flex !w-full items-center pr-3">
+        <a
+          class="profileMenuButton text-xs"
+          mat-menu-item
+          [routerLink]="['user', userService.user?.id]"
+        >
+          Profile <mat-icon>account_circle</mat-icon>
+        </a>
```
romeonicholas, line 67: Moritz is right, moving the right padding from the menu to the links enables the hover effect to cover the entire element. I removed some CSS that didn't seem to have any effect, and I think this overall looks more standard as a dropdown menu (icon followed by left aligned text).

```<mat-menu #profileMenu="matMenu" class="flex items-center">```

header.component.css:
```
.profileMenuButton {
  padding: 0 15px 0 15px;
  text-align: left;
}
```
MenuDropdown

File: frontend/src/app/sessions/sessions.component.html (hunk 6 above)
Comment (romeonicholas) [reply frontend/src/app/sessions/sessions.component.html]: Unless there's some reason for the Active Sessions element to be responsive (it only appears in the sessions homepage, right?), I think you can probably just delete .mat-card from active-sessions.component.css and let it be constrained by the w-[450px] of the entire component in the sessions.component.html so it looks the same as the persistent sessions card. But it's not essential or anything, on production they're currently different sizes, it's really only noticable between like 500-900px screen width, but putting their widths in the same spot might make it more consistent or easier ot update later.

File: frontend/src/app/general/header/header.component.html (hunk 7 above)
Comment (romeonicholas) [reply frontend/src/app/general/header/header.component.html:67]: (Though it might be nicer to get rid of the external CSS and just use `class="px-[15px] text-left text-xs"` on each link)

File: frontend/src/app/sessions/sessions.component.html (hunk 6 above)
Comment (Paula-Kli) [reply frontend/src/app/sessions/sessions.component.html]: alright I see - thank you!

Comment (Paula-Kli) [reply frontend/src/app/sessions/sessions.component.html]: I think that can be an enhancement for the future as the button above `Open internal session viewer` does not scale the same as the underlying card and thus would result in funny looking results

File: frontend/src/app/general/header/header.component.html (hunk 7 above)
Comment (Paula-Kli) [reply frontend/src/app/general/header/header.component.html:67]: thanks a lot!

> (Though it might be nicer to get rid of the external CSS and just use `class="px-[15px] text-left text-xs"` on each link)
mmh but then we will need to copy the code for every menu item don't we?

File: frontend/src/index.html (hunk 4 above)
Comment (Paula-Kli) [reply frontend/src/index.html:27]: Notes: Buttons have a letter spacing of 1.25 according to [the material desing type system](https://m2.material.io/design/typography/the-type-system.html#type-scale)

File: frontend/src/app/general/header/header.component.html (hunk 7 above)
Comment (Paula-Kli) [reply frontend/src/app/general/header/header.component.html]: alright then I will add it - thanks!

**Comment by romeonicholas** [comment]
I think the landing page/Sessions view is probably set--the most distracting things (button fonts, text sizing/weights, icon placement, responsive elements) have been fixed. Some of these don't actually match production (like icons in the dropdown menu now being on the left instead of right) but that's intentional since it's a more standard placement. Moving on to other pages, should be faster now that I've figured out the font and button issues. 
Sessions_mobile

Sessions_dekstop

**Comment by MoritzWeber0** [comment]
> I think the landing page/Sessions view is probably set--the most distracting things (button fonts, text sizing/weights, icon placement, responsive elements) have been fixed. Some of these don't actually match production (like icons in the dropdown menu now being on the left instead of right) but that's intentional since it's a more standard placement. Moving on to other pages, should be faster now that I've figured out the font and button issues. Sessions_mobile
> 
> Sessions_dekstop

Looks good! :)

**Comment by romeonicholas** [comment]
I _think_ this covers everything, but honestly there are probably pages or states I'm missing. There are definitely things I see that could be centralized or refactored, but that may be better saved for an actual UI overhaul or transition to Storybook. Let me know if anything looks broken to you.

**Comment by romeonicholas** [comment]
Oh, and @MoritzWeber0, I wasn't sure what to do about attribution, do I need to rebase this onto Paula's single commit?

**Comment by MoritzWeber0** [comment]
> Oh, and @MoritzWeber0, I wasn't sure what to do about attribution, do I need to rebase this onto Paula's single commit?

No, in this case, two commits are totally fine 👍

File: frontend/src/app/projects/project-detail/model-overview/model-overview.component.html [hunk 8]
```diff
@@ -4,8 +4,9 @@
  -->
 
 <div>
-  <div class="flex items-center">
-    <h2>Models</h2>
+  <div class="mt-1.5 flex items-center">
+    <h2 class="text-xl font-medium">Models</h2>
+    <h2 class="text-xl font-medium">Models</h2>
```
MoritzWeber0, line 9: The `<h2>` is added twice: 

image

frontend/src/app/projects/project-detail/project-users/project-user-settings.component.html
```diff
@@ -6,12 +6,12 @@
 <div *ngIf="!hasRoute('projects/create')" class="mr-2.5">
```
MoritzWeber0, line 6: image

The "delete" icon is not centered vertically in the button. Also, the `<div>` which contains "techuser and "Manager, read & write" is not centered vertically.

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
 import { HTTP_INTERCEPTORS, HttpClientModule } from '@angular/common/http';
```
MoritzWeber0, line 11: I'll some general findings here (each as single conversations so that you can resolve them step by step):

URL: http://localhost:4200/settings/core/pipelines
Issue: Background of some text is white, not matching the light grey background.
Screenshot:
image

MoritzWeber0, line 11: URL: http://localhost:4200/project/default (Click on the move model button on any model and select a target project)
Issue: Title "Move model" is not large / bold
Screenshot:
image

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
...
 import { NgModule } from '@angular/core';
```
MoritzWeber0, line 12: The "active session" card can be optimized a bit: 
image

The status bar ("Session started") is a bit too wide, The title ("Persistent workspace session") should be larger and bold.

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
...
 import { FormsModule, ReactiveFormsModule } from '@angular/forms';
```
MoritzWeber0, line 13: Issue: 
- View diagrams should be larger and bold
- Margin left & right of the Download button is not the same.
- Gap between mat cards is missing.

Screenshot:
image


PS: If you need help to set up the diagram cache, let me know.

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
...
+import { MatAutocompleteModule } from '@angular/material/autocomplete';
```
MoritzWeber0, line 14: Where? Diagram cache preview

Issues:
- Diagram preview title should be larger & bold
- There are scrollbars verticals and horizontal


image

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
...
 import { MatBadgeModule } from '@angular/material/badge';
```
MoritzWeber0, line 15: URL: http://localhost:4200/project/coffee-machine/metadata
Issue: 
- Remove project configuration title, it's already clear from the breadcrumb

Screenshot:
image

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
...
+import { MatButtonModule } from '@angular/material/button';
```
MoritzWeber0, line 16: URL: http://localhost:4200/project/coffee-machine/model/coffee-machine/metadata
Issue:
- Remove Name of model (Coffee machine), it's already displayed in the form field directory below the title
Screenshot:
image

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
...
 import { MatButtonToggleModule } from '@angular/material/button-toggle';
```
MoritzWeber0, line 17: URL: http://localhost:4200/settings/core/users

Issue:
- admin user is smaller than the other users
Screenshot:
image

frontend/src/app/app.module.ts
```diff
@@ -11,35 +11,35 @@ import { CommonModule } from '@angular/common';
...
+import { MatCardModule } from '@angular/material/card';
```
MoritzWeber0, line 18: URL: http://localhost:4200/settings/modelsources/t4c/instance/1

Issue: Gap between mat-cards is missing
image

frontend/src/app/app.module.ts
MoritzWeber0, line 11: Where? Home screen
Issue: Button shall be full width in the mat card (check current prod for reference)

image

frontend/src/styles.css
```diff
@@ -4,13 +4,21 @@
...
+
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
 
+* {
+  font-family: Roboto, "Helvetica Neue", sans-serif;
+}
+
```
MoritzWeber0, line 17: This has some unwanted side-effects. I think we need to revert the change. 
The CSS rule applies the font-family on all elements, also the elements where another font is defined.

I just found two cases where this an issue: 
- The tailwind class `font-mono` in DisplayValueComponent doesn't apply anymore, it's overwritten by this rule.
- The font in the monaco editor is broken, go to `Menu` > `Settings` > `Configuration`

PR #1241, opened 2023-12-18
fix: Use wider input fields in project metadata dialog

MoritzWeber0:
Before: 
image

After: 
image

### DIFF
--- a/frontend/src/app/projects/project-detail/edit-project-metadata/edit-project-metadata.component.html
+++ b/frontend/src/app/projects/project-detail/edit-project-metadata/edit-project-metadata.component.html
@@ -5,10 +5,9 @@
 <div *ngIf="projectUserService.verifyRole('manager')"></div>
 <div class="wrapper">
   <mat-card id="metadata-card">
-    <h2 id="title">Description</h2>
     <form [formGroup]="form">
       <div>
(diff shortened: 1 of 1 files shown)


frontend/src/app/projects/project-detail/edit-project-metadata/edit-project-metadata.component.html
dominik003, line 1: Your changes look good to be but what is going on in line 5 here? For me it seems like everything should be within this currently empty div (i.e., we can simply move the `ngIf` to the div in line 6.

dominik003 (approved):
LGTM! There is only one interesting finding unrelated to your changes but would be good to have a look at it and potentially fix it.

frontend/src/app/projects/project-detail/edit-project-metadata/edit-project-metadata.component.html
MoritzWeber0: It doesn't make much sense, indeed. I've removed the empty div.

Authentication is also not needed here as non-project-manager doesn't see the button to the page. Requests are refused by the backend anyways.

However, we should add a proper project authentication handling in the frontend via AuthGuards. There should be redirect to an access denied page in case of access issues.

PR #1254, opened 2024-01-02
docs: Update licences and instances of DB Netz AG to DB InfraGO AG

### LINKED ISSUE #1225: Replace license headers with new entity
At the beginning of next year, the license headers should be replaced. 

Old license headers: 
```txt
# SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
```

New license headers:
```txt
# SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
```

romeonicholas:
Resolves #1225

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <!--
- ~ SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+ ~ SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
  ~ SPDX-License-Identifier: Apache-2.0
  -->
 
@@ -14,7 +14,7 @@
 
 A web platform for collaboration on MBSE and [Capella](https://www.eclipse.org/capella/) projects.
 
-**Copyright 2021 - 2023 [DB Netz AG](https://fahrweg.dbnetze.com/),
+**Copyright 2021 - 2024 [DB InfraGO AG](https://fahrweg.dbinfrago.com/fahrweg-de),
 licensed under Apache 2.0 License (see full text [here](./LICENSES/Apache-2.0.txt))**
 
 Turn your local MBSE and Capella experience into a browser-based collaboration platform for
--- a/.git_archival.txt
+++ b/.git_archival.txt
@@ -1,4 +1,4 @@
-SPDX-FileCopyrightText: Copyright DB Netz AG and the capella-collab-manager contributors
+SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
 SPDX-License-Identifier: CC0-1.0
 
 node: $Format:%H$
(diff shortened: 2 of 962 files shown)


romeonicholas:
Hey @dominik003, it looks like because I'm touching every single file sonarcloud has an issue with existing concerns. I'm not sure I want to resolve those as Won't Fix without someone taking a look first though, if you're able  🐞

dominik003:
> Hey @dominik003, it looks like because I'm touching every single file sonarcloud has an issue with existing concerns. I'm not sure I want to resolve those as Won't Fix without someone taking a look first though, if you're able 🐞

Sure, I'll look over! 

In addition, `mypy` seems to fail in our pre-commit hook with the following error:
```
Argument 2 to "add_exception_handler" of "Starlette" has incompatible type "Callable[[Request, UserNotFoundError], Coroutine[Any, Any, Response]]"; expected "Callable[[Request, Exception], Response | Awaitable[Response]] | Callable[[WebSocket, Exception], Awaitable[None]]"  [arg-type]
```

So essentially stating that our type:

```
Callable[[Request, UserNotFoundError], Coroutine[Any, Any, Response]]
```

is not one of the following expected one:

```
Callable[[Request, Exception], Response | Awaitable[Response]] | Callable[[WebSocket, Exception], Awaitable[None]]
```

However, as far as I can see, the our `UserNotFoundError` is a sub type of `Exception` and as stated in the [docs](https://docs.python.org/3/library/collections.abc.html#collections.abc.Coroutine), *All Coroutine instances are also instances of Awaitable*. So together with the fact that this just works fine, I would suggest to simply put `# type: ignore[arg-type]` to all of those lines.

The only weird thing is that this error just occurred now, so not quite sure what changed for it to occur.

README.md
```diff
@@ -14,7 +14,7 @@
 
 A web platform for collaboration on MBSE and [Capella](https://www.eclipse.org/capella/) projects.
 
-**Copyright 2021 - 2023 [DB Netz AG](https://fahrweg.dbnetze.com/),
+**Copyright 2021 - 2023 [DB InfraGO AG](https://fahrweg.dbnetze.com/),
```
dominik003, line 17:
```suggestion
**Copyright 2021 - 2023 [DB InfraGO AG](https://fahrweg.dbinfrago.com/fahrweg-de),
```

dominik003 (approved):
Thank you for taking care of the renaming. There is only one link in the README, which we can already change to the new page. After that I can merge it.

README.md
romeonicholas: Ooh good catch, though should I also update the copyright to 2024?

dominik003: Also good catch from you 😄 Feel free to also update the copyright year

romeonicholas: Not sure why I even asked, done 😆 ✅

MoritzWeber0:
👍

PR #1274, opened 2024-01-16
fix: Correctly call `psql` in probes

dominik003:
The `-W` flag in psql is used to provide a password prompt to the user. However, this lead to a problem that the probes were no longer successful due to that. Chaning this to `-w' ensures that no prompt is given and additionally,
we don't need the password here.


Wuestengecko (approved):

MoritzWeber0 (approved):
Good catch. Thanks for the fix! 👍 

Tested it locally and works as expected.

PR #1259, opened 2024-01-08
build: Update frontend versions

Paula-Kli:

PR #1278, opened 2024-01-22
fix: Bump pre-commit version and fix pre-commit issues

dominik003:


.pre-commit-config.yaml
```diff
@@ -130,31 +130,31 @@ repos:
...
-          - "@typescript-eslint/eslint-plugin@^6.12.0"
-          - "@typescript-eslint/parser@^6.12.0"
+          - "eslint-plugin-import@^2.29.1"
+          - "@typescript-eslint/eslint-plugin@^6.19.0"
+          - "@typescript-eslint/parser@^6.19.0"
           - "eslint-plugin-unused-imports@^3.0.0"
           - "eslint-plugin-deprecation@^2.0.0"
-          - "eslint-plugin-tailwindcss@^3.13.0"
-        args: ["--fix"]
+          - "eslint-plugin-tailwindcss@3.13.1"
```
MoritzWeber0, line 151: The versions of the additional dependencies should always match the versions of the package.json. Otherwise, eslint will behave differently between running via pre-commit and running via the CLI directly.

.pre-commit-config.yaml
```diff
@@ -130,31 +130,31 @@ repos:
...
+        args: ["--fix", "--config", "./frontend/.eslintrc.js"]
```
MoritzWeber0, line 152: ESLint detects the configuration automatically by traversing the file path. I don't think that we need to explicitly specify it, especially as the pre-commit rule is not limited to the frontend directory.

MoritzWeber0 (changes requested):

.pre-commit-config.yaml
Paula-Kli: How do we want to continue with that as here https://github.com/DSD-DBS/capella-collab-manager/pull/1259/files those dependencies are also updated accordingly in the package.json

MoritzWeber0: I've merged #1259 and it should no longer be necessary to update those dependencies in this PR.

MoritzWeber0:
There is one small change that I don't understand. You've downgraded the `eslint-plugin-tailwindcss` dependency from version `3.14.0` to version `3.13.1`. I don't think that this is necessary, it does also run with the latest version on my side.

Therefore, I'll replace this PR in favor of #1288.

PR #1321, opened 2024-02-05
refactor: Optimize some error messages

MoritzWeber0:
Make them clearer and easier to understand.

### DIFF
--- a/backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
+++ b/backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
@@ -45,8 +45,9 @@ def verify_path_prefix(db: orm.Session, path: str):
     raise fastapi.HTTPException(
         status_code=status.HTTP_404_NOT_FOUND,
         detail={
-            "err_code": "no_git_instance_with_prefix_found",
-            "reason": "There exist no git instance having the resolved path as prefix. Please check whether you correctly selected a git instance.",
+            "err_code": "NO_GIT_INSTANCE_WITH_PREFIX_FOUND",
+            "reason": "We couldn't find a matching Git instance. Please that your system administrator allows the given URL.",
+            "technical": f"There is no Git instance, which has a prefix of the path '{path}' as URL.",
         },
     )
 
--- a/backend/capellacollab/projects/toolmodels/routes.py
+++ b/backend/capellacollab/projects/toolmodels/routes.py
@@ -97,8 +97,8 @@ def create_new_tool_model(
         raise fastapi.HTTPException(
             status_code=status.HTTP_409_CONFLICT,
             detail={
-                "reason": "A model with a similar name already exists.",
-                "technical": "Slug already used",
+                "reason": f"A model with the name {new_model.name} already exists.",
+                "technical": f"The slug '{slugify.slugify(new_model.name)}' is already used",
             },
         )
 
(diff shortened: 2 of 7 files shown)


backend/capellacollab/projects/toolmodels/routes.py
```diff
@@ -144,7 +144,8 @@ def patch_tool_model(
             raise fastapi.HTTPException(
                 status_code=status.HTTP_409_CONFLICT,
                 detail={
-                    "reason": "A model with a similar name already exists."
+                    "reason": "A model with a similar name already exists.",
```
romeonicholas, line 147: I'd consider being more explicit e.g. "A model with the name '{new_model.name}' already exists."

backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
```diff
@@ -45,8 +45,8 @@ def verify_path_prefix(db: orm.Session, path: str):
     raise fastapi.HTTPException(
         status_code=status.HTTP_404_NOT_FOUND,
         detail={
-            "err_code": "no_git_instance_with_prefix_found",
-            "reason": "There exist no git instance having the resolved path as prefix. Please check whether you correctly selected a git instance.",
+            "err_code": "NO_GIT_INSTANCE_WITH_PREFIX_FOUND",
+            "reason": "The path validation failed. No Git instance exists with the resolved path as prefix.",
```
romeonicholas, line 49: "The path validation failed" may be a bit redundant in this context. Maybe "No Git instance could be found at the resolved path {whatever resolved path is}"? I guess I'm not sure if they need to know something more specific about that.

backend/capellacollab/projects/toolmodels/routes.py
MoritzWeber0: There might be some slug collisions even if the name is different. Also in this case, we'd enter this error case.
However, as the user doesn't really care about it, I used your proposal and clarified it in the technical section.

backend/capellacollab/projects/toolmodels/modelsources/git/routes.py
MoritzWeber0: We do a bit of prefix matching here. For example: 

A Git instance has `https://github.com` as base url. `https://github.com/DSD-DBS` would a valid path (the instance is a prefix of it), but `https://gitlab.com` is not valid.

But the user doesn't really care about it, so I rewrote it a bit: 
```py
detail={
    "err_code": "NO_GIT_INSTANCE_WITH_PREFIX_FOUND",
    "reason": "We couldn't find a matching Git instance. Please that your system administrator allows the given URL.",
    "technical": f"There is no Git instance, which has a prefix of the path '{path}' as URL.",
},
```

romeonicholas: Ah, ok, that makes more sense. Though unfortunately I think you missed a word in "Please that your system administrator allows the given URL."

MoritzWeber0: Replace "Please" with "Make sure".

MoritzWeber0: https://github.com/DSD-DBS/capella-collab-manager/pull/1325

PR #1323, opened 2024-02-06
chore: Apply prettier layout globally

MoritzWeber0:


romeonicholas (approved):
Running this branch locally worked fine, and consistency is always nice. I don't think I see anything that would break in the .yamls and I have to assume prettier would do any necessary escape changes anyway?

MoritzWeber0:
> Running this branch locally worked fine, and consistency is always nice. I don't think I see anything that would break in the .yamls and I have to assume prettier would do any necessary escape changes anyway?

The PR shouldn't contain any breaking changes 👍

PR #1328, opened 2024-02-07
refactor: Use dataclasses for all SQLAlchemy database classes

### LINKED ISSUE #1326: Integrate SQLAlchemy with dataclasses
https://docs.sqlalchemy.org/en/20/orm/dataclasses.html

The main advantage would be autocomplete for the init methods of Database classes.

MoritzWeber0:
With dataclasses, a `__init__` function is created for all database classes. This adds autocomplete and code recommendations and type checkers like mypy can check the passed types.

Since the `__hash__` method was removed during this change, the intersection
method to get common projects doesn't work anymore. I changed it to a proper
database join, which is faster.

In addition, many type annotations were added. Some mypy and pylint errors
and warnings were fixed in the tests.

Resolves #1326

### DIFF
--- a/backend/capellacollab/core/database/__init__.py
+++ b/backend/capellacollab/core/database/__init__.py
@@ -17,7 +17,7 @@ engine = sa.create_engine(
 SessionLocal = orm.sessionmaker(autocommit=False, autoflush=False, bind=engine)
 
 
-class Base(orm.DeclarativeBase):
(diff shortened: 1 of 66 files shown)


romeonicholas (approved):
Latest update fixed the loading issues I saw on the projects overview and user profile pages 👍

PR #1329, opened 2024-02-07
refactor: Enforce mentioning of lifecycle hook interfaces

MoritzWeber0:


romeonicholas (approved):
Looks good. I think there are a few missing implementations throughout the project, I don't know if we should bother adding those immediately or just wait for the files to be eventually touched (I'm thinking of all the linting warnings/errors I saw when changing the license for every file in January 😆)

MoritzWeber0:
> Looks good. I think there are a few missing implementations throughout the project, I don't know if we should bother adding those immediately or just wait for the files to be eventually touched (I'm thinking of all the linting warnings/errors I saw when changing the license for every file in January 😆)

There shouldn't be any other occurrence since I've changed the level from warning to error. Due to the change, `pre-commit` will error if there is any detected occurrence, but the pipeline is green.

PR #1332, opened 2024-02-08
fix: Add loading case for "Common projects"

MoritzWeber0:
Before, the frontend has displayed "No common projects" while the client waited for the response from the server.

Instead, a simple "Loading..." is now displayed.


romeonicholas (approved):
Ah yeah this makes more sense, I could see a user getting really confused if the response is delayed.

User w/ common projects: saw (obviously very brief with a local setup) load followed by project cards
User w/ no common projects: saw updated error
Admin w/ common projects: same as user experience
Admin w/ no common projects: saw "You do not have..." text

(Also I should probably be updating those *ngIfs when I see them, the new version is _so_ much easier to parse)

PR #1335, opened 2024-02-12
feat: Add descriptions, examples, and validations to pydantic fields

### LINKED ISSUE #1327: Improve API documentation with pydantic fields
To improve the API documentation, each pydantic attribute should have a `pydantic.Field` as default value, which has the attributes "description" and "examples". In addition, the validation can be strengthened with min_len, max_len and pattern attributes.

romeonicholas:
Resolves #1327 

Adding descriptions, examples, and some validations to pydantic models. Figuring out the right balance has been a little complicated. 

For 'description': A lot of these are self-evident, and I considered adding tags to indicate something the user enters (e.g. a model name is chosen by a user, but a model slug is derived and not directly modifiable) but that felt like it was getting too deep into implemenation and things that could change.

For 'examples': I've avoided including examples where they might be meaningless (any integer field), and used partial objects in place of full ones when it's reuse of a model created elsewhere (e.g. a partial user example on the event model, because the full user model is covered elsewhere).

For validations: So far I think I've only added them for user-inputted strings, with relatively arbitrary values that we should probably check against in production. We could potentially just add the limits to post and patch models, leaving existing overly-long offenders in place if we want to avoid a migration? 

Still lots of fields left to work on.

### DIFF
--- a/backend/capellacollab/core/metadata.py
+++ b/backend/capellacollab/core/metadata.py
@@ -4,7 +4,7 @@
 import typing as t
 
 import fastapi
-import pydantic
+from pydantic import BaseModel, ConfigDict, Field
 from sqlalchemy import orm
 
 import capellacollab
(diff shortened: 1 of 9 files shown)


romeonicholas:
Note to future self: be more careful not to ignore default assignments and break the tests 😆 
Assigning fields to something like `name: str | None = None` requires shifting the existing default into Field as in: `name: str | None = Field(default=None, whateverElse="..."`

PR #1342, opened 2024-02-13
build: Update frontend versions

romeonicholas:
This is my first frontend version update, let me know if anything looks off 👍 
(Front end looked good when I built it locally and explored most of the views)


MoritzWeber0 (approved):
Nice, thanks for taking care! :)

PR #1343, opened 2024-02-14
fix: Align projects title in monitoring overview

MoritzWeber0:


romeonicholas (approved):
Risky change, but worth it 😝

PR #1344, opened 2024-02-16
feat: Make resources configurable per tool

### LINKED ISSUE #1136: Add resources to configuration of tools in the core settings
Every tool has its own requirements of resources. Currently the resource requirements are hardcoded and there are only 2 options for low and high limit. The ability to configure the `V1ResourceRequirements` needs to be added to the tool configuration.

MoritzWeber0:
Instead of the hard-coded resource configuration, it can now be set per tool.

In addition, this commit adds job labels to cronjobs.

Resolves #1136.

### DIFF
--- a/backend/capellacollab/sessions/operators/k8s.py
+++ b/backend/capellacollab/sessions/operators/k8s.py
@@ -26,6 +26,8 @@ from openshift import dynamic
 from openshift.dynamic import exceptions as dynamic_exceptions
 
 from capellacollab.config import config
+from capellacollab.sessions import models as sessions_models
+from capellacollab.tools import models as tools_models
 
 from . import helper, models
 
@@ -177,17 +179,18 @@ class KubernetesOperator:
         self,
         image: str,
         username: str,
-        session_type: str,
-        tool_name: str,
-        version_name: str,
+        session_type: sessions_models.WorkspaceType,
+        tool: tools_models.DatabaseTool,
+        version: tools_models.DatabaseVersion,
         environment: dict[str, str],
         ports: dict[str, int],
         volumes: list[models.Volume],
         prometheus_path="/metrics",
         prometheus_port=9118,
-        limits="high",
     ) -> Session:
-        log.info("Launching a %s session for user %s", session_type, username)
+        log.info(
+            "Launching a %s session for user %s", session_type.value, username
+        )
 
         _id = self._generate_id()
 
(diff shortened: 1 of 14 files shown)


romeonicholas (commented):
Still mid-review (maybe finished this afternoon) but just a callout that I appreciate the clearer naming and stronger typing in a lot of these changes

Edited to say: ok wow this is a long one, still looking through it 😆

backend/capellacollab/sessions/operators/k8s.py
```diff
@@ -177,27 +179,28 @@ def start_session(
...
         ports: dict[str, int],
         volumes: list[models.Volume],
         prometheus_path="/metrics",
         prometheus_port=9118,
-        limits="high",
     ) -> Session:
-        log.info("Launching a %s session for user %s", session_type, username)
+        log.info(
+            "Launching a %s session for user %s", session_type.value, username
+        )
```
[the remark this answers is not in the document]
MoritzWeber0: Don't use f-strings in log messages: https://google.github.io/styleguide/pyguide.html#3101-logging

In other places it's fine.

MoritzWeber0:
> Edited to say: ok wow this is a long one, still looking through it 😆

There were some outdated changes in the PR, which were removed during the last rebase.  You only have to review the commit "feat: Make resources configurable per tool". 
Unfortunately I also merged https://github.com/DSD-DBS/capella-collab-manager/pull/1356 already, which you can ignore.

backend/capellacollab/tools/models.py
```diff
@@ -40,6 +40,71 @@ class ToolIntegrations(pydantic.BaseModel):
...
+    requests: str = pydantic.Field(
+        default="1.6Gi",
+        description=(
+            "Each session gets at least the specified amount of memory. "
+            "An inaccurate value can lead to higher costs. " + RESOURCES_DOCS
+        ),
+        examples=["100Mi", "1.6Gi", "2Gi"],
+    )
+    limits: str = pydantic.Field(
+        default="6Gi",
```
romeonicholas, line 84: Seeing mixed numbers and measures makes me nervous but if these are the expected values for configurations I guess it would just be extra work and no gain to store them as enumerated strings or as a combination number + string measurement?

romeonicholas (approved):
Looks good (and way easier to read with fewer files 😆)

backend/capellacollab/tools/models.py
MoritzWeber0: Why does that make you nervous? I would actually say that it makes the documentation easier to read. Split entries are harder to read.

romeonicholas: I'm not _actually_ worried about it, but they can't be validated or constrained as easily (e.g. 500Gi instead of Mi, text other than Mi or Gi)

MoritzWeber0: Validation for valid values should be simple (a simple regex which checks for number + valid unit). 
Enforcing an upper bound is harder, we would need to resolve the unit for it (but this would be the same with separate fields for number + unit). Just a single number is less readable, I'd say.

PR #1356, opened 2024-02-19
feat: Make external links available in development mode

MoritzWeber0:
Grafana, Prometheus and Documentation are not available on relative paths in the development mode. Instead, they have to point to corresponding absolute URLs.

Documentation will point to the mkdocs serve.
Grafana & Prometheus to the local cluster deployment.


romeonicholas (approved):
Makes sense. Our staging instance uses the production environment, right? So this only impacts local dev?

MoritzWeber0:
> Makes sense. Our staging instance uses the production environment, right? So this only impacts local dev?

Yes, it's only used for the local development environment (not the local cluster deployment).

PR #1364, opened 2024-02-22
feat: Convert config template to pydantic model

### LINKED ISSUE #1165: Transform `config_template.yaml` into Pydantic object
The `config_template.yaml` should be transformed into a Pydantic object. Validation can be performed with pydantic directly. The `config_template.yaml` can be deleted (it can be still generated with Pydantic).

romeonicholas:
Resolves #1165 by replacing the configuration template with a pydantic object from which default configurations can be written. 

- Attempting to launch the app without a configuration file will now automatically generate a configuration
- The application loads the configuration as a pydantic object, and all existing references to the YAML have been switched over to using the python object attributes instead
- The new model uses snake case naming conventions for attributes, with camel case aliases for writing to YAML (this is done automatically for all attributes, but a few require manual aliasing for some common acronyms)


MoritzWeber0:
> I'm definitely doing something wrong here but I'm not familiar with yaml. @MoritzWeber0 my yaml output is currently looking like this, do you think I'm doing something wrong in my classes or in deriving the yaml from the classes?: (the function generating the yaml is just at the bottom of config_template.py for now)
> 
> ```
> docker: !!python/name:__main__.DockerConfig ''
> k8s: !!python/name:__main__.K8sConfig ''
> general: !!python/name:__main__.GeneralConfig ''
> extensions: !!python/name:__main__.ExtensionsConfig ''
> ...
> ...
> ```

Use `app_config.model_dump()` instead `app_config.__dict__`

MoritzWeber0:
> Perfect, thanks @MoritzWeber0, I switched to BaseModels so I could use model_dump() and now it's _almost_ correct, just need to tweak a few fields I entered incorrectly. Last question would be: how do we present or implement this for the user?
> 
> * Easiest to implement but most work for the end user: just update the documentation to say something like "you can run the following command to generate your config before first launch" as part of the setup process
> * Hardest to implement but easiest for the user: automatically generate a config based off the template if they don't have one (where would that happen? The Makefile?)
> 
> Option A would obviously only take me a few minutes, I'd need a little starting direction for Option B so I knew where to look.

Go for option A: 
- If no config file exists in the list of places, fail during startup. Log an error message how to generate the config file.
- Add a CLI option (we already have an CLI interface for config.diff) to generate the config file based on the default values.

backend/config/config_template.py
```diff
@@ -0,0 +1,172 @@
+# SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
```
MoritzWeber0, line 1: Integrate the `config_template` properly in the config module (backend/capellacollab/config) is the right module.

backend/.gitignore
```diff
@@ -4,6 +4,7 @@
 .idea
 config/*
 !config/config_template.yaml
+!config/config_template.py
```
MoritzWeber0, line 7: Remove those two lines

backend/config/config_template.py
```diff
@@ -0,0 +1,172 @@
...
+    endpoints: AuthOathEndpointsConfig | None = None
+    audience: str = "default"
+    scopes: list[str] = dataclasses.field(default=lambda: ["openid"])
+    client: AuthOauthClientConfig = AuthOauthClientConfig
+    redirectURI: str = "http://localhost:4200/oauth2/callback"
+
+
+@dataclasses.dataclass
+class AuthenticationConfig:
+    provider: str = "oauth"  # oauth | azure
```
MoritzWeber0, line 109: Support for the Azure provider seems to be missing, check the config_schema for all available option. The validation behaviour shouldn't change.

backend/config/config_template.py
MoritzWeber0, line 1: Make sure to update the config loader to validate the config with the pydantic model and replace the config variable with the pydantic object. Make sure that all references to the config are updated accordingly.

romeonicholas: I wasn't sure if you mean actually putting the config_template in backend/capellacollab/config, but that gets messy when trying to generate the config. I think ideally we'd tell users to run `python -m capellacollab.config.config_template`, but the config module needs a config to already exist for  __init__.py so it will just fail. The workarounds I thought of were to tell users to run the script directly instead of as a module (`python backend/capellacollab/config/config_template.py` seems less elegant), checking what file was calling the module and skipping the config load if it is config_template (seems hacky), or instead of failing to launch the app when no config exists to just pass a default AppConfig without saving it to file and leaving config tweaking to advanced users entirely. Any thoughts?

MoritzWeber0: `python -m capellacollab.config.config_template` looks a bit weird as `config` is mentioned twice and the intention isn't clear. What about `python -m capellacollab.config.generate`? Models always belong to the `models.py` in the specific module. 

> config module needs a config to already exist for init.py so it will just fail when someone with a fresh repo copy tries it

You can move the config loading to the `__main__.py`

> instead of failing to launch the app when no config exists to just pass a default AppConfig without saving it to file and leaving config tweaking to advanced users entirely.

Thought about that too, but then the user doesn't even know that a configuration exists and how to modify it (if it's a user action to generate it, it's clear) and if the config is placed at a wrong location, it would start with the defaults and the user might not notice it directly, is also hard to debug.

romeonicholas: Excellent points, I'll follow this and then I think(?) I'll be set to submit for review, though I saw some weird behavior in diff.py I might need to check first (it wasn't catching new or deleted keys for some reason).

romeonicholas:
@MoritzWeber0 I can't figure out why deployments/test-backend is timing out, and I'm not sure where I should be looking--the generated config seems to match the old one, and even when I include the fallback template_config this still doesn't get past waiting for the containers to be ready. Do you have any ideas?

MoritzWeber0:
> @MoritzWeber0 I can't figure out why deployments/test-backend is timing out, and I'm not sure where I should be looking--the generated config seems to match the old one, and even when I include the fallback template_config this still doesn't get past waiting for the containers to be ready. Do you have any ideas?

You can check the logs of the container (check the "Print logs of the backend container" step), there you can see the stacktrace: 

```py
ERROR:    Traceback (most recent call last):
  File "/opt/.venv/lib/python3.11/site-packages/starlette/routing.py", line 734, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  File "/usr/lib/python3.11/contextlib.py", line 204, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/opt/.venv/lib/python3.11/site-packages/fastapi_pagination/api.py", line 380, in lifespan
    async with _original_lifespan_context(app) as maybe_state:
  File "/opt/.venv/lib/python3.11/site-packages/starlette/routing.py", line 610, in __aenter__
    await self._router.startup()
  File "/opt/.venv/lib/python3.11/site-packages/starlette/routing.py", line 711, in startup
    await handler()
  File "/tmp/backend/capellacollab/__main__.py", line 68, in startup
    operators.get_operator()
  File "/tmp/backend/capellacollab/sessions/operators/__init__.py", line 12, in get_operator
    return KubernetesOperator()
           ^^^^^^^^^^^^^^^^^^^^
  File "/tmp/backend/capellacollab/sessions/operators/k8s.py", line 81, in __init__
    self.load_config()
  File "/tmp/backend/capellacollab/sessions/operators/k8s.py", line 93, in load_config
    kubernetes.config.load_config(context=cfg.context)
  File "/opt/.venv/lib/python3.11/site-packages/kubernetes/config/__init__.py", line 49, in load_config
    load_incluster_config(**kwargs)
TypeError: load_incluster_config() got an unexpected keyword argument 'context'
```

MoritzWeber0:
In the cluster deployment, the helm/config/backend.yaml configuration file is used.

backend/capellacollab/core/authentication/provider/azure/routes.py
```diff
@@ -22,15 +23,21 @@
...
 
 @lru_cache
 def ad_session():
     return ConfidentialClientApplication(
-        cfg["client"]["id"],
-        client_credential=cfg["client"]["secret"],
-        authority=cfg["authorizationEndpoint"],
+        cfg.client.id if cfg else None,
+        client_credential=cfg.client.secret if cfg else None,
+        authority=cfg.authorizationEndpoint if cfg else None,
```
romeonicholas, line 40: The Azure | OAuth config union raises attribute errors in the few places we were already expecting Azure or OAuth configuration. This is obviously an inelegant solution, would it be better to throw an error before the ConfidentialClientApplication return if we're using the wrong config here?

romeonicholas:
```
def ad_session():
     if cfg is None:
         raise ValueError("Configuration for Azure is not set.")
```
Changed it to use a basic exception, I don't know how thorough you want it to be since I'm not sure how someone would get to the azure route without the proper config anyway

MoritzWeber0:
I applied some changes and have force-pushed the changes. Please reset your local state to the new state on remote.
You can find the changes here: 

Mainly, those are: 
- The log message that a configuration was generated is red to make it more visible. 
- The log message prints the full absolute path now
- The config_schema and loading of the config schema is removed
- `write_config` uses pathlib instead of os.path and was simplified
- config fallback locations were removed (loading of `config_template.yaml`)
- scopes can be None (check the config_schema, type can be array or null)
- Jupyter config is removed, was removed with v3.0.0 of the Collaboration Manager.

Next steps / TODO for you: 
- Verify again that the new configuration matches the old `config_schema.yaml`. For example: 
   If type is defined as `string` / `null`, make sure that the pydantic object accepts `string | None`
 as type! For example, None wasn't considered for the scopes attribute, but null is valid according to the config_schema.
- Set extra to forbid for the AppConfig and all other defined configuration pydantic models: https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra. The application should fail if there is an unexpected configuration variable.
- Use snake case for all attributes and define an alias in pydantic. In Python, we want to use snake case, but in the YAML camelCase is more common. https://docs.pydantic.dev/latest/concepts/alias/
- Add descriptions and example values if applicable to the specific entries.

MoritzWeber0 (commented):
Thanks @romeonicholas. This is one of the best PRs of this year and will significantly improve my life. Finally, we have typing of the configuration. 

I love the precision of changes and the accurate descriptions and examples. I did some changes and added conversations for it, but everything is already resolved. It's just for transparency what I've changed. 

Tested on staging and works like a dream.

There is just one issue with the tests, you used `unittest`, but our used framework is `pytest`. Therefore, I started to migrate the tests to pytest, but there is a bit of work to do.

backend/tests/config/test_configuration.py
```diff
@@ -0,0 +1,294 @@
+# SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
+# SPDX-License-Identifier: Apache-2.0
+
+
+from unittest import mock
```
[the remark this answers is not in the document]
MoritzWeber0: Done, removed the diff feature from the Collaboration Manager since it wasn't known enough.

MoritzWeber0 (approved):

PR #1368, opened 2024-02-23
build: Remove `environment.prod.ts`

MoritzWeber0:
The `environment.ts` doesn't contain any user-defined variables anymore. They were all moved to the backend configuration.

Therefore, we can remove the option to customize the `environment.prod.ts`


dominik003 (approved):
LGTM!

PR #1386, opened 2024-02-28
fix: Reenable uvicorn error logger

MoritzWeber0:
The uvicorn error logger was disabled in #440 when our own logger was introduced. Unfortunately, our own logger doesn't work in all stages of the application.

For example, add a startup script which raises an exception to the on_startup list. The application startup will fail, but there is no error message.

This makes debugging hard and errors are overseen. The commit reenables the error log.


romeonicholas (approved):
Ran fine locally, I'm not sure what could go wrong if this was already in use before

Issue #693, opened 2023-04-19
Add T4C license consumption to `/metrics` endpoint.


MoritzWeber0:
Add the current TeamForCapella license consumption to the metrics endpoint. This can be used to gain information about the average license usage.

PR #1387, opened 2024-02-28
feat: Expose metrics about used TeamForCapella license usage

MoritzWeber0:
A new gauge metric `t4c_licenses` was added to the `/metrics` endpoint, which contains the number of currently used licenses per TeamForCapella instance. The instance name is used as label for the metric.

Note: If you have multiple TeamForCapella server instances, which point to the same license server, this is not detected. Therefore, make sure to filter the `t4c_licenses` for one instance first.

Resolves https://github.com/DSD-DBS/capella-collab-manager/issues/693.


dominik003 (approved):
LGTM! I was going to suggest using an (abstract) `collector` class, since the two collectors are very similar (and there is the possibility to outsource the retrieval of the instances), but apparently there is already such an abstract class in the Prometheus client, but `starlett-prometheus` uses a rather old version of it where it is not yet added.

PR #1388, opened 2024-02-28
refactor: Remove usage of bare-except

MoritzWeber0:
More info why bare-except shouldn't be used:
https://pylint.pycqa.org/en/latest/user_guide/messages/warning/bare-except.html


romeonicholas (approved):
Interesting, makes sense

PR #1390, opened 2024-02-28
refactor: Remove OpenShift specific code, use Ingress instead

MoritzWeber0:
OpenShift can translate Ingress routes to OpenShift routes if the `route.openshift.io/termination` annotation is used.

Therefore, we can get rid of the OpenShift specific code and use Ingress instead.

More information is available in the OpenShift documentation: https://docs.openshift.com/container-platform/4.11/networking/routes/route-configuration.html


backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -81,11 +81,11 @@ def post_session_creation_hook(
         **kwargs,
     ):
         assert self._jupyter_public_uri.hostname
-        operator.create_public_route(
-            session_id=session_id,
+        operator.create_ingress(
+            id=session_id,
```
dominik003, line 85: I would either use `id` or `name` in both the create and delete ingress function and not two different names for the same purpose.

dominik003 (approved):
Nice reduced code! Just two small comments and after you've had a look at them, this can be merged.

backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -81,11 +81,11 @@ def post_session_creation_hook(
...
             host=self._jupyter_public_uri.hostname or "",
             path=self._determine_base_url(user.name),
-            port=8888,
+            port_number=8888,
```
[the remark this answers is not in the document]
MoritzWeber0: I haven't renamed it, I just switched to another interface (another function is called). But I will rename it to port.

backend/capellacollab/sessions/hooks/jupyter.py
MoritzWeber0: Makes sense, `id` as variable name is bad practice anyway.

PR #1394, opened 2024-02-29
feat: Make Prometheus persistent

MoritzWeber0:
Prometheus will now store the idletime of sessions and newly introduced TeamForCapella license usage. This allows us to look back to time frames before the last restart.


dominik003 (approved):
LGTM! I just did a rough estimate of how much space is required for a one year period, using the following approach [how-to-calculate-the-storage-space-required-for-prometheus-server](https://tanmay-bhat.github.io/posts/how-to-calculate-the-storage-space-required-for-prometheus-server/) which is at about 0.7GB at the moment. However, as the current retention period is set to ~12h, we should check whether this assumptions holds after we deployed this, and if not, increase the storage space.

PR #1410, opened 2024-03-05
feat: Remove connection method specific variables

dominik003:
This change is possible because we no longer need connection method-specific variables to perform our backups or export. You can find more information on this in the corresponding Docker images PR: https://github.com/DSD-DBS/capella-dockerimages/pull/254.

Requires https://github.com/DSD-DBS/capella-dockerimages/pull/254 to be deployed.


MoritzWeber0 (approved):
Haven't tested it (will do in the staging environment), but the change looks good.

PR #1414, opened 2024-03-11
feat: Disable automatic service account mounting

MoritzWeber0:
Service accounts aren't needed for most of the services since they don't talk to the Kubernetes API.

For increased security, this disables the automatic mounting of secrets.


dominik003 (approved):
LGTM! I have just tried it in a local deployment and everything works as expected..

PR #1422, opened 2024-03-12
fix: Allow downgrading from administrator to user

dominik003:
This PR allows downgrading from administrators to normal users again. To do this, we mainly did some major clean-up work on this component to remove any code that was used when there were no separate icons for each user, but you had to select a user and then buttons appeared. Also, the `*ngFor` statements used here were transformed into the new Angular control flow and additionally the CSS, which mainly contained unused rules, was transformed into Tailwind CSS.


MoritzWeber0 (approved):
Thanks for the change, looks good.

I added some additional changes: 
- Use `w-card` instead of `w-96`
- Added margins in Confirmation & Input dialog
- Removed color primary from cancel button in Confirmation dialog
- Optimized the messages for up- and downgrade of users

PR #1428, opened 2024-03-15
feat: Remember last requested sessions locally

MoritzWeber0:
The last three requested sessions are cached locally and can be used to quick-request those sessions again.
The session history is lazy-loaded, so it doesn't slow down loading of the page.

### DIFF
--- a/frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-session-history/create-session-history.component.ts
+++ b/frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-session-history/create-session-history.component.ts
@@ -1,0 +1,151 @@
+/*
+ * SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import { CommonModule } from '@angular/common';
+import { Component, OnDestroy, OnInit } from '@angular/core';
+import { MatIconModule } from '@angular/material/icon';
+import { MatProgressSpinnerModule } from '@angular/material/progress-spinner';
+import { Subscription, filter, take } from 'rxjs';
+import { SessionService } from 'src/app/sessions/service/session.service';
+import {
+  SessionHistoryService,
+  SessionRequestHistory,
+} from 'src/app/sessions/user-sessions-wrapper/create-sessions/create-session-history/session-history.service';
+import {
+  ConnectionMethod,
+  Tool,
+  ToolService,
+  ToolVersion,
+} from 'src/app/settings/core/tools-settings/tool.service';
+
+@Component({
+  selector: 'app-create-session-history',
+  standalone: true,
+  imports: [CommonModule, MatIconModule, MatProgressSpinnerModule],
+  templateUrl: './create-session-history.component.html',
+  styles: ``,
+})
+export class CreateSessionHistoryComponent implements OnInit, OnDestroy {
+  resolvedHistory: ResolvedSessionRequestHistory[] = [];
+  sessionsToBeLoaded = this.sessionHistoryService.MAX_SESSIONS_IN_HISTORY;
+  sessionsLoaded = 0;
+
+  activeSubscriptions: Subscription[] = [];
+
+  get sortedResolvedHistory() {
+    return this.resolvedHistory.sort(
+      (a, b) => b.lastTimeRequested.getTime() - a.lastTimeRequested.getTime(),
+    );
+  }
+
+  constructor(
+    private toolService: ToolService,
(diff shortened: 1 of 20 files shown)


frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-session-history/create-session-history.component.ts
```diff
@@ -0,0 +1,132 @@
...
+    private sessionHistoryService: SessionHistoryService,
+    private sessionService: SessionService,
+  ) {}
+
+  ngOnInit() {
+    this.loadLastSessions();
+    this.sessionHistoryService.loadSessionRequestHistory();
+  }
+
+  loadLastSessions() {
```
dominik003, line 51: As discussed, here are some ideas on how we could simplify this function. These are all my potential ideas of how I would probably simplify this, so feel free to apply the suggestions you find valuable, but ignore the ones you don't like.

- Introduce a new backend API endpoint where we can get a specific version for a given `toolID` and `versionID`, or if no such version exists, we get an error back. This is especially useful here because we already have the `toolID', `versionID' and `connectionMethodID' in local storage.
- It may be useful to move the code for finding a specific connection method (via its ID) for a given tool to a helper function here. This increases readability here, as the descriptive name of the function tells someone directly what is going on, and understanding the actual implementation is more difficult and also of no real interest in this function (i.e., if interested someone can just look into the helper function).
- Instead of subscribing to all loaded tools and then searching for the one with the correct ID, it may be useful to create a helper function in `tool.service.ts` that gets a toolID and returns the corresponding loaded tool if it exists. The function name should indicate that no API call is made, only the currently loaded tools are considered. Also, it may be easier/necessary to have this function return an observable, which shouldn't be an issue here.
- It may be useful to move the unsubscribe from all active subscriptions to a separate function. Again, the function name tells the reader directly what is going on, and if they are interested, they can simply look at the specific function. In addition, it may be a good idea to have a `ngOnDestroy` function here to ensure that there are no active subscriptions left that can also use this function.
- It may be useful to move the pushing of new entries to a separate function. This function could be called something like `pushValidResolvedSessionRequestHistory`, which first checks if all three values are defined, and if so, pushes them to the list.

dominik003 (changes requested):
Nice improvement, works on my site and looks pretty good too! There are a few comments where it would be great to have some changes, especially considering the complex function in `create-session-history.componet.ts`. Also, please consider adding stories for the new and adapted component.

frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-session-history/create-session-history.component.ts
MoritzWeber0: > Introduce a new backend API endpoint where we can get a specific version for a given toolID and versionID, or if no such version exists, we get an error back. This is especially useful here because we already have the toolID', versionID' and `connectionMethodID' in local storage.

Done. Also added tests for the GET versions and GET natures routes.

MoritzWeber0: > Instead of subscribing to all loaded tools and then searching for the one with the correct ID, it may be useful to create a helper function in tool.service.ts that gets a toolID and returns the corresponding loaded tool if it exists. The function name should indicate that no API call is made, only the currently loaded tools are considered. Also, it may be easier/necessary to have this function return an observable, which shouldn't be an issue here.

Instead of passing the list of tools, I use the value of the BehaviourSubject instead. I don't return an Observable. If it's not loaded, I return undefined. The name contains the name "Cached".

MoritzWeber0: > It may be useful to move the unsubscribe from all active subscriptions to a separate function. Again, the function name tells the reader directly what is going on, and if they are interested, they can simply look at the specific function. In addition, it may be a good idea to have a ngOnDestroy function here to ensure that there are no active subscriptions left that can also use this function.

The name isn't more descriptive than the content of the function, but I added the ngOnDestroy function and a outsourced the code to a function since it's used twice now.

MoritzWeber0: > It may be useful to move the pushing of new entries to a separate function. This function could be called something like pushValidResolvedSessionRequestHistory, which first checks if all three values are defined, and if so, pushes them to the list.

Since I've applied the other comments, the checks are now made at different places. It only checks for the connection method in the function. tool, version and session are passed.

PR #1437, opened 2024-03-19
fix: Remove deprecated interface in authentication

romeonicholas:
Fix for the following deprecation warning:

capellacollab/core/authentication/__init__.py:14
capellacollab/core/authentication/__init__.py:14
  /Users/nicholasromeo/code/capella-collab-manager/backend/capellacollab/core/authentication/__init__.py:14: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
    for i in metadata.entry_points()[

### DIFF
--- a/backend/capellacollab/core/authentication/__init__.py
+++ b/backend/capellacollab/core/authentication/__init__.py
@@ -11,9 +11,9 @@ def get_authentication_entrypoint():
     try:
         ep = next(
             i
-            for i in metadata.entry_points()[
-                "capellacollab.authentication.providers"
-            ]
+            for i in metadata.entry_points().select(
+                group="capellacollab.authentication.providers"
+            )
             if i.name == config["authentication"]["provider"]
         )
         return ep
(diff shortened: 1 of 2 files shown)


backend/capellacollab/core/authentication/__init__.py
dominik003, line 1: Nice cleanup of the deprecation warning. It looks like we also have a sample code snippet on how to include extensions in our documentation at `docs/docs/development/backend/extensions.md`. It would be great if you could also adapt line 81 there to use the `select` as well. Then we can merge them.

dominik003 (commented):
LGTM! I just found a second use of the deprecated variant in our documentation that should be adjusted before we merge this.

backend/capellacollab/core/authentication/__init__.py
romeonicholas: Oh, good catch, I'll do that

MoritzWeber0 (approved):

PR #1440, opened 2024-03-19
docs: Add troubleshooting section for Capella

MoritzWeber0:
For the beginning, it contains two HOW-TOs:
- Debug Capella crashes
- Issues with the description editor


docs/docs/user/tools/capella/troubleshooting.md
```diff
@@ -0,0 +1,45 @@
...
+    [Github repository](https://github.com/eclipse/capella/issues).
+
+    Administrators can see the logs of all sessions to identify the issues
+    remotely. In addition, the session owner can also see the events in the UI.
+    In your session, please follow these steps:
+
+    1. Open the Capella search
+    1. Search for `Event logs` and click on the following entry:
+    1. You should now see the list of events and you can try to find an entry with
+       a matching timestamp.
```
dominik003, line 22: When I started a new capella session and searched for `Event log` in the capella search, no results were displayed. So is it only visible if there was at least one crash/event or should I find it in any case?

Also, I seem to be missing an image in the second item as it says *"[...] and click on the following entry: "* without further information after the colon.

dominik003 (approved):
I only have two points on the first entry about troubleshooting, and after addressing them, this one can be merged.

docs/docs/user/tools/capella/troubleshooting.md
MoritzWeber0: image

I can find Event log, can you search again?

MoritzWeber0: Second comment is valid and fixed.

Issue #1442, opened 2024-03-19
Add support for session pre-authentication


MoritzWeber0:
We shouldn't rely on the tool to implement authentication. Instead, the Collaboration Manager should make sure that all requests to sessions are authenticated.

PR #1447, opened 2024-03-19
feat: Add support for session pre-authentication

MoritzWeber0:
We make sure that requests to all http-based sessions are authenticated.

If the tool handles authentication as required in earlier versions, it will still work, but there is another unnecessary layer of authentication, which can be removed if wanted. 

Resolves https://github.com/DSD-DBS/capella-collab-manager/issues/1442


dominik003 (approved):
LGTM.

PR #1449, opened 2024-03-20
test: Add tests for HTTP session hook and GET session routes

MoritzWeber0:


backend/tests/sessions/hooks/test_http.py
```diff
@@ -0,0 +1,49 @@
...
+    connection_method = tools_models.HTTPConnectionMethod(
+        redirect_url="http://localhost:8000/{TEST2}"
+    )
+    result = http.HTTPIntegration().session_connection_hook(
+        db_session=session,
+        connection_method=connection_method,
+        logger=logging.getLogger(),
+    )
+
+    assert result["warnings"]
```
romeonicholas, line 49: I see that right now the only way "warnings" gets added here is through an exception during redirect_url formatting, but for general test readability or just because "warnings" might get expanded somehow later I'd probably be more specific in the assertion if possible.

MoritzWeber0: I've added an error code to the message, which I use to do a more accurate validation now 👍

PR #1456, opened 2024-03-22
feat: Skip loading of Jupyter file-shares if they don't exist

MoritzWeber0:
If the file-shares don't exist as PVCs in the Kubernetes cluster, continue and start the session without the file-shares mounted.


backend/capellacollab/sessions/hooks/jupyter.py
```diff
@@ -49,6 +57,22 @@ def _get_project_share_volume_mounts(
...
+            )
+
+            if not operator.persistent_volume_exists(volume_name):
+                warnings.append(
+                    core_models.Message(
+                        err_code="JUPYTER_FILE_SHARE_VOLUME_NOT_FOUND",
+                        reason=(
+                            f"The Jupyter file-share volume for the model '{model.name}' in the project '{model.project.name}' couldn't be located. "
+                            "Please contact your system administrator or recreate the model (this will erase all data in the file-share)."
+                        ),
```
romeonicholas, line 71: This doesn't actually require any change, but seeing the core_models Message model after redoing all those exceptions made me think we could probably do a quick clean up of existing Messages too (right now all the fields are optional and pretty inconsistently applied).

MoritzWeber0: Sounds good, can you take care of it?

Issue #945, opened 2023-08-15
Refactor `ProjectRoleVerification` to use injectables


MoritzWeber0:
The `RoleVerification` and `ProjectRoleVerification` depend on the token currently. The `ProjectRoleVerification` depends on the `project_slug` in addition. They can be replaced with the `get_own_user` and `get_existing_project` injectables.

MoritzWeber0:
Done with https://github.com/DSD-DBS/capella-collab-manager/pull/2087

PR #1460, opened 2024-03-26
feat: Introduce McCabe complexity check

### LINKED ISSUE #1380: Introduce McCabe complexity check
More information is available in the internal PR: https://github.com/DSD-DBS/cookiecutter-dbs-python/pull/72

dominik003:
This PR enables the McCabe [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) check for pylint. This forces us not to make extensive use of nesting within functions, but to outsource parts to separate functions. For now we use the default threshold value of *10*.

Resolves #1380

### DIFF
--- a/backend/capellacollab/core/authentication/injectables.py
+++ b/backend/capellacollab/core/authentication/injectables.py
@@ -150,36 +150,16 @@ class ProjectRoleVerification:
         username: str = fastapi.Depends(get_username),
         db: orm.Session = fastapi.Depends(database.get_db),
     ) -> bool:
-        if not (user := users_crud.get_user_by_name(db, username)):
-            if self.verify:
-                raise fastapi.HTTPException(
-                    status_code=status.HTTP_404_NOT_FOUND,
-                    detail={"reason": f"User {username} was not found"},
-                )
+        if not (user := self._get_user_and_check(username, db)):
             return False
 
         if user.role == users_models.Role.ADMIN:
             return True
 
-        if not (
-            project := projects_crud.get_project_by_slug(db, project_slug)
-        ):
-            if self.verify:
-                raise fastapi.HTTPException(
-                    status_code=status.HTTP_404_NOT_FOUND,
-                    detail={
-                        "reason": f"The project {project_slug} was not found"
-                    },
-                )
+        if not (project := self._get_project_and_check(project_slug, db)):
             return False
 
-        if (
-            project.visibility == projects_models.Visibility.INTERNAL
-            and self.required_role
-            == projects_users_models.ProjectUserRole.USER
-            and self.required_permission
-            in (None, projects_users_models.ProjectUserPermission.READ)
-        ):
+        if self._is_internal_project_accessible(project):
             return True
 
         project_user = projects_users_crud.get_project_user_association(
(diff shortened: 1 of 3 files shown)


backend/capellacollab/core/authentication/injectables.py
```diff
@@ -150,36 +150,16 @@ def __call__(
         username: str = fastapi.Depends(get_username),
         db: orm.Session = fastapi.Depends(database.get_db),
     ) -> bool:
-        if not (user := users_crud.get_user_by_name(db, username)):
-            if self.verify:
-                raise fastapi.HTTPException(
-                    status_code=status.HTTP_404_NOT_FOUND,
-                    detail={"reason": f"User {username} was not found"},
-                )
+        if not (user := self._get_user_and_check(username, db)):
```
MoritzWeber0, line 153: This seems to be a bit overkill for me and doesn't reduce the complexity in the code as intended by introducing McCabe. 

Instead, I'd propose do change it properly and update to our new exception style. Instead of `fastapi.HTTPException`, raise custom exceptions. Those exceptions all inherit from a general exception class. 

We can then remove the `verify` option. All places where `verify=False` was used have to be refactored to catch the error with try & except.

MoritzWeber0 (changes requested):

backend/capellacollab/core/authentication/injectables.py
MoritzWeber0: See also: https://github.com/DSD-DBS/capella-collab-manager/issues/945

MoritzWeber0: Dominik and I have agreed on only changing the fastapi.HTTPExceptions to BaseExceptions.

backend/capellacollab/core/authentication/injectables.py
```diff
@@ -197,20 +177,62 @@ def __call__(
...
-            ):
-                if self.verify:
-                    raise fastapi.HTTPException(
-                        status_code=403,
-                        detail={
-                            "reason": f"You need to have '{self.required_permission.value}'-access in the project!",
-                        },
-                    )
-                return False
+            return True
```
MoritzWeber0, line 222: !Important!

This line is dangerous and overrides the check as to whether the minimum required permission is met. 
Probably you meant

```py
if not self.required_permission:
    return True
```

backend/capellacollab/core/authentication/injectables.py
MoritzWeber0, line 1: Please add tests for the authentication module (patch coverage is <= 50%), but the authentication module is extremely important.

In the test, routes should be defined with the corresponding minimum levels of permissions / roles and checked against real requests.

MoritzWeber0 (changes requested):

backend/capellacollab/core/authentication/injectables.py
MoritzWeber0: I'll do the definition of exceptions as part of https://github.com/DSD-DBS/capella-collab-manager/pull/1569.

backend/capellacollab/core/authentication/injectables.py
MoritzWeber0: Tests are added.

PR #1467, opened 2024-03-27
feat: Standardize Message model and update messages to match

romeonicholas:
The existing Message pydantic model used for warnings and errors was pretty loose and inconsistently used, this PR makes all fields required and updates all existing uses of the model.

I removed the _techincal_ field because it was only used once or twice in places where exceptions were already being logged along with the warnings. err_code may also be unecessary, it's currently only being used in tests.


romeonicholas:
codecov/patch is failing, but the new code doesn't add or meaningfully change the existing functions, it looks like [backend/capellacollab/settings/modelsources/t4c/repositories/routes.py](https://app.codecov.io/gh/DSD-DBS/capella-collab-manager/commit/6f53fcece46d7cc5400f6fb947d8accf3bf9b677/blob/backend/capellacollab/settings/modelsources/t4c/repositories/routes.py) is just already missing some coverage.

backend/capellacollab/core/models.py
```diff
@@ -9,10 +9,18 @@
...
-    technical: str | None = None
+    err_code: str = pydantic.Field(
+        description="The error code of the message, used for testing, not displayed in the frontend.",
+        examples=["T4C_SERVER_UNREACHABLE"],
+    )
+    title: str = pydantic.Field(
+        description="The title of the message, displayed in the frontend",
+        examples=["TeamForCapella server not reachable"],
+    )
+    reason: str = pydantic.Field(
```
MoritzWeber0, line 20: Since you've dropped support for tuple, we can also adjust the frontend implementation for it. Have a look at the `error-handling.interceptor.ts`.

The parts `Array.isArray(error.reason)` and `Array.isArray(warning.reason)` can be removed.

backend/capellacollab/core/models.py
MoritzWeber0, line 1: I'd not consider this as feature (think from the end-user perspective), but as refactoring. Please update the commit message type accordingly.

MoritzWeber0, line 1: Some lines that you've refactored are not covered by tests. Please add tests to verify that those changes are non-breaking.

MoritzWeber0 (changes requested):

MoritzWeber0:
@dominik003 Can you take over the PR apply the suggested changes?

MoritzWeber0 (approved):

PR #1470, opened 2024-03-27
docs: Shorten URLs with duplicate names in path

MoritzWeber0:
We use nesting in our documentation files to store screenshots in the same context with the markdown file.

This usually adds another nesting layer in the URL, e.g. the path `getting_started/getting_started.md` was rendered to `getting_started/getting_started`.

Instead, this could be simplified by using a file with the name `index.md`: The path `getting_started/index.md` renders to `getting_started`.

This looks cleaner, shortens the URLs and removes duplication.


romeonicholas (approved):
Agreed, this generally is easier to follow. This is pretty extensive anyway, but some quick changes I'd suggest or would want to hear your thoughts on:

- There are some inconsistencies in which routes have screenshot folders and which keep screenshots in the same folder as the route, should that be standardized since that's a pretty easy fix? (and one single /img folder under admin/getting_started that sticks out)
- We have a few pages that are the only page in their route, should they get their own folder and change to index.md or be generalized to be the landing page for those routes more broadly? (I'm looking at alerts/create.md or settings/monitoring.md for instance)

If you have opinions on those I could update that pretty quickly.

MoritzWeber0:
> * There are some inconsistencies in which routes have screenshot folders and which keep screenshots in the same folder as the route, should that be standardized since that's a pretty easy fix? (and one single /img folder under admin/getting_started that sticks out)

Consistency makes sense, but it's not urgent. Feel free to go ahead and implement it (not important enough to create an issue for it).

> * We have a few pages that are the only page in their route, should they get their own folder and change to index.md or be generalized to be the landing page for those routes more broadly? (I'm looking at alerts/create.md or settings/monitoring.md for instance)

Only introduce nesting where necessary. Single pages should stay as single file if they don't have related attachments / images. We can still introduce a folder + index.md when images are added, this is a non-breaking change since the URL doesn't change.

Issue #1372, opened 2024-02-26
Try alternatives to webpack


MoritzWeber0:
We have some issues with webpack, esp. since with monaco, monaco-yaml and storybook.

Instead, we should try: 
- Bun’s package manager, supported with Angular 17.2: https://blog.angular.io/angular-v17-2-is-now-available-596cbe96242d
- New Angular build system: https://angular.io/guide/esbuild

In both cases, make sure to adapt the monaco editor so that it still works.

PR #1475, opened 2024-03-28
build: Switch to ESBuild as builder and Vite as dev-server

MoritzWeber0:
The combination increases build performance & speed. In addition, all the hacky webpack workarounds could be removed.

Build time decreased from 92.9s to 30.6s in the pipeline.

In addition, fonts are now embedded during build, not loaded from Google servers during runtime anymore. This improves privacy & security and makes the frontend GDPR compliant.

Resolves https://github.com/DSD-DBS/capella-collab-manager/issues/1372


romeonicholas (approved):
This is really cool, it also seems to build faster for me locally and it looks like you caught a few minor front-end issues I hadn't noticed had regressed

PR #1483, opened 2024-04-02
docs: Disable font loading from Google fonts

MoritzWeber0:
Loading fonts from Google Fonts is not GDPR compliant. Instead, fall-back to the system font.


romeonicholas (approved):
Oh, good catch, of course the default would rely on google fonts. This branch fell back on my system fonts 👍

PR #1493, opened 2024-04-05
build: Update frontend versions

romeonicholas:


MoritzWeber0 (approved):
Thanks for taking care 👍

PR #1520, opened 2024-04-19
feat(frontend): Add "404 - Not found" page

MoritzWeber0:


frontend/src/app/general/404/404.component.html
```diff
@@ -0,0 +1,31 @@
+<!--
+ ~ SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
+ ~ SPDX-License-Identifier: Apache-2.0
+ -->
+
+<div class="collab-card m-2.5 mb-12 !bg-[#010035] text-center text-white">
```
dominik003, line 6: nit: It may be useful to define the banner color as a custom color, so we can use its name here. This has the advantage that changing the color only has to be done in one place, and in places where it is used, it is also immediately clear what the color is for.

dominik003 (approved):
LGTM! There is only one suggestion that you can include if you want to.

frontend/src/app/general/404/404.component.html
MoritzWeber0: Done.

PR #1522, opened 2024-04-22
feat: Disable persistent workspace sessions for specific tools

MoritzWeber0:
Some tools can't handle persistent workspaces. Instead, they only support read-only sessions.

### DIFF
--- a/backend/capellacollab/tools/models.py
+++ b/backend/capellacollab/tools/models.py
@@ -217,6 +217,16 @@ class ToolModelProvisioning(pydantic.BaseModel):
     )
 
 
+class PersistentWorkspaceSessionConfiguration(pydantic.BaseModel):
+    mounting_enabled: bool = pydantic.Field(
+        default=True,
+        description=(
+            "Enables workspace mounting to persistent workspace sessions of this tool. "
+            "If disabled, persistent workspace sessions can no longer be requested."
+        ),
+    )
+
+
 class ToolSessionConfiguration(pydantic.BaseModel):
     resources: Resources = pydantic.Field(default=Resources())
     environment: dict[str, str] = pydantic.Field(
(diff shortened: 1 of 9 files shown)


backend/capellacollab/tools/models.py
dominik003, line 1: IMO the names here are a bit too undescriptive, i.e. when I would read `workspaces.enabled` I would have no idea what exactly it is used for and would have to look at its description. If you agree, I would suggest renaming `enabled`, `mounting_enabled` and `workspaces` to `persistent_workspaces_config` (or use some other descriptive name).

frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-persistent-session/create-persistent-session.component.ts
```diff
@@ -116,4 +117,10 @@ export class CreatePersistentSessionComponent implements OnInit {
         }
       });
   }
+
+  get toolsWithWorkspaceEnabled(): Observable<Tool[] | undefined> {
+    return this.toolService.tools$.pipe(
+      map((tools) => tools?.filter((tool) => tool.config.workspaces.enabled)),
+    );
+  }
```
dominik003, line 125: I would suggest moving this to the tool service itself and using a similar method as we already have in the user session service, which would look like this (if you applied some renaming, the example needs to be adjusted):

```ts
public readonly toolsWithWorkspaceEnabled$ = this.tools$.pipe(
  filter(Boolean),
  map((tools) => tools.filter((tool) => tool.config.workspaces.enabled),
) 
```

Theoretically, we could also add an `isToolWorkspaceEnabled` function, but that is only useful if we plan to use it somewhere else in the future, so I assume we can skip that here.

dominik003 (approved):
It would be great to have a short PR description why this change is necessary. Other than that, there are only two comments that would be great to address, and then this can be merged.

frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-persistent-session/create-persistent-session.component.ts
MoritzWeber0: Services are always loaded globally and global variables should only be used where absolutely necessary (e.g., to prevent code duplication). It makes the toolService larger & harder to understand.

When we start to create Observables to receive each different value of tools, this will end up in many unnested Observables. 

Since it's only used in this component and has a clear scope in the component, I wouldn't move it to the service.

backend/capellacollab/tools/models.py
MoritzWeber0: Is updated.

PR #1526, opened 2024-04-23
feat: Add provisioning of compatible tools

MoritzWeber0:


backend/tests/sessions/hooks/test_provisioning_hook.py
```diff
@@ -158,3 +159,37 @@ def test_tool_model_mismatch(
...
+    """Make sure that provisioning is successful when the tool is compatible with the tool of the model"""
+
+    tool_version.config.compatible_versions = [capella_tool_version.id]
+    orm.attributes.flag_modified(tool_version, "config")
+    db.commit()
+
+    response = hooks_provisioning.ProvisionWorkspaceHook().configuration_hook(
+        db=db,
+        tool=capella_tool,
+        tool_version=capella_tool_version,
```
dominik003, line 183:
```suggestion
        tool=tool,
        tool_version=tool_version,
```
This is needed because we want to test if we can request a session with tool `tool` and tool version `tool_version` using a Capella model.

backend/tests/sessions/hooks/test_provisioning_hook.py
```diff
@@ -158,3 +159,37 @@ def test_tool_model_mismatch(
                 )
             ],
         )
+
+
+def test_provision_session_with_compatible_tool_versions(
```
dominik003, line 164: Theoretically, we could also add a test that checks if an error is thrown when trying to deploy a session with an incompatible tool version, i.e. very similar to this test, but without adding the capella tool to the compatible versions.

frontend/src/app/sessions/user-sessions-wrapper/create-session/create-readonly-session/create-readonly-session.component.ts
```diff
@@ -60,33 +62,82 @@ export class CreateReadonlySessionComponent implements OnInit {
...
+  }
+
+  resolveVersionCompatibility(
+    models: Model[],
+    allVersions: ToolVersionWithTool[],
+  ): void {
+    this.models = [];
+
+    const modelsWithVersion = models.filter((m) => m.version);
+    for (const model of modelsWithVersion) {
```
dominik003, line 94:
```suggestion
    for (const model of models) {
      if (!model.version) {
        continue;
      }
```
nit: This way we avoid using the `!` all the time when trying to access the version, it should also be more efficient, and at least for me it is as readable as the current version.

frontend/src/app/sessions/user-sessions-wrapper/create-session/create-readonly-session/create-readonly-session.component.ts
```diff
@@ -60,33 +62,82 @@ export class CreateReadonlySessionComponent implements OnInit {
...
     }
-    const toolIds = this.models.map((m) => m.tool.id);
-    return this.toolService.tools.filter((t) => toolIds?.includes(t.id));
+    return this.models
+      ?.map((m) => m.compatibleVersions.map((version) => version.tool))
+      .flat()
+      .filter(
+        (value, index, self) =>
+          index === self.findIndex((tool) => tool.id === value.id),
+      );
```
dominik003, line 124: I don't really understand what the `filter` does here. So it would be great to either do it differently here so that it is directly understandable, outsource it to a function with a proper descriptive name (if that is possible?), or add a comment here explaining what is being done.

frontend/src/app/sessions/user-sessions-wrapper/create-session/create-readonly-session/create-readonly-session.component.ts
```diff
@@ -60,33 +62,82 @@ export class CreateReadonlySessionComponent implements OnInit {
...
   requestReadonlySession(): void {
     if (this.toolSelectionForm.valid) {
       const dialogRef = this.dialog.open(CreateReadonlySessionDialogComponent, {
         data: {
           projectSlug: this.projectSlug,
-          models: this.models,
-          modelVersionId: this.toolSelectionForm.value.version!.id,
+          models: this.models?.filter(
+            (model) =>
+              model.compatibleVersions.findIndex(
```
dominik003, line 134: Similar to above, it is somewhat confusing and not intuitive to me why we use indices here. So please consider making this clearer in some way.

frontend/src/app/sessions/user-sessions-wrapper/create-session/create-readonly-session/create-readonly-session.component.ts
```diff
@@ -99,14 +150,15 @@ export class CreateReadonlySessionComponent implements OnInit {
...
-      const toolVersionIds = this.models?.map((m) => m.version?.id);
-      this.toolVersions = toolVersions.filter((v) =>
-        toolVersionIds?.includes(v.id),
+    this.relevantToolVersions = this.models
+      ?.map((m) => m.compatibleVersions)
+      .flat()
+      .filter((v) => v.tool.id === tool.id)
+      .filter(
+        (value, index, self) =>
+          index === self.findIndex((version) => version.id === value.id),
```
dominik003, line 160: Similar to above, please either use a proper function here, make this more understandable, or add a comment explaining what and why this is done here.

frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-readonly-session/create-readonly-model-options/create-readonly-model-options.component.html
```diff
@@ -5,13 +5,22 @@
 
 <div [formGroup]="form">
   <h2>
-    <mat-checkbox color="primary" formControlName="include">{{
-      modelOptions.model.name
-    }}</mat-checkbox>
+    <mat-checkbox color="primary" formControlName="include"
+      >{{ modelOptions.model.name }}
```
dominik003, line 9:
```suggestion
    <mat-checkbox color="primary" formControlName="include">
      {{ modelOptions.model.name }}
```
nit: These cases are not really caught by the autoformatter, but I still try to fix them when I see them.

dominik003 (changes requested):
There are a few comments that should be addressed before we merge this, but in general it looks good!

backend/capellacollab/tools/routes.py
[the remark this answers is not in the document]
MoritzWeber0: Renamed & moved to the bottom of the file.

backend/tests/sessions/hooks/test_provisioning_hook.py
MoritzWeber0: Actually, exactly this case is already covered by `test_tool_model_mismatch`. We have a tool and request a session with an incompatible model.

backend/tests/sessions/hooks/test_provisioning_hook.py
MoritzWeber0: Good catch, I've changed it.

frontend/src/app/sessions/user-sessions-wrapper/create-session/create-readonly-session/create-readonly-session.component.ts
MoritzWeber0: It removes duplicates from the list. I've outsourced it to a function `removesToolDuplicates`.

frontend/src/app/sessions/user-sessions-wrapper/create-session/create-readonly-session/create-readonly-session.component.ts
MoritzWeber0: Also outsourced to a function `removeVersionDuplicates`.

MoritzWeber0: Outsourced to function `isCompatibleWithSelectedVersion`.

PR #1528, opened 2024-04-25
feat: Add cascading deletion of tool natures & versions

MoritzWeber0:
When deleting a tool, versions and natures were not deleted. This commit cleans up orphan versions and natures and adds cascading deletion for future tool deletions.


dominik003 (approved):
LGTM.

PR #1534, opened 2024-04-29
chore: Migrate to Angular standalone components

dominik003:
This migratee the frontend components to standalone components. This is required for Storybook to use components, but also removes the need for the `@NgModule` inside the app modules file. The changes here were mostly done automatically using the [Standalone Migration CLI](https://angular.io/guide/standalone-migration).


MoritzWeber0 (approved):
The token overview didn't load anymore, had to add `MatNativeDateModule` to the `main.ts`. Beside that, things look fine after some manual testing. Thanks for the change :)

PR #1539, opened 2024-04-29
fix(front): Reload active session in DOM only on change or modification

dominik003:
This first introduces the new angular control flow in this component. By doing so, we can easily add `track session.id` to the session loop. By tracking sessions by their id, only newly created or modified sessions are reloaded in the DOM. This not only improves performance by reducing the number of reloads, but also allows us to properly select parts of the session information, since previously each reload removed the selection. Finally, we also cleaned up the unused `sessions` variable and converted most of the styling to tailwind.

In addition, this PR also introduces stories for the Active Sessions component. These stories cover loading sessions, no active sessions, success, warning, error, and unknown states, as well as persistent and read-only sessions. It also includes a small fix that moves the Create Read-Only Dialog stories to the correct story group.


MoritzWeber0 (approved):
LGTM and thanks for adding the story! 
I've renamed Session Component to Session Components to be consistent with the other sections in Storybook.

PR #1545, opened 2024-04-30
docs: Add make target to add grafana helm repo and build dependencies

zusorio:


Makefile
```diff
@@ -81,6 +81,10 @@ deploy-t4c: build t4c-client helm-deploy rollout open
 
 deploy-without-build: helm-deploy rollout open
 
+setup-helm:
+	helm repo add grafana https://grafana.github.io/helm-charts
+	helm dependency build helm
+
```
MoritzWeber0, line 87: I'd not treat it as separate Make target. 

Instead, I'd try to automate the dependency updates. 
`helm dependency update ./helm` was part of the Make target `helm-deploy`, but was removed in https://github.com/DSD-DBS/capella-collab-manager/pull/1355. 

The reason for the removal was that Helm doesn't cache dependencies and always downloads the Chart again. This significantly slowed down the helm deployment. 

`helm dependency list .` is fast and checks if Charts are missing. A simple logic to update with caching would be: 
```zsh
[[ ! $(helm dependency list . | grep missing) ]] || helm dependency update ./helm;
```

MoritzWeber0: I've opened #1546 with my proposal.

PR #1546, opened 2024-04-30
build: Fetch missing helm dependencies automatically

MoritzWeber0:
If a Helm dependency is missing, fetch it in the `helm-deploy` target.

Thanks to @zusorio for reporting in #1545

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,6 +84,7 @@ deploy-without-build: helm-deploy rollout open
 helm-deploy:
 	@k3d cluster list $(CLUSTER_NAME) >/dev/null || $(MAKE) create-cluster
 	@kubectl create namespace $(SESSION_NAMESPACE) 2> /dev/null || true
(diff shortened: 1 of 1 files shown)


zusorio:
This seems like a better solution, although I think it doesn't address the issue of the missing grafana repository

MoritzWeber0:
> This seems like a better solution, although I think it doesn't address the issue of the missing grafana repository

The repository is defined in the [`Chart.yaml`](https://github.com/DSD-DBS/capella-collab-manager/blob/main/helm/Chart.yaml). A `dependency update` should fetch all defined repositories. 

I don't have the repository in the list of global helm repositories (`helm repo list`) doesn't show any repositories and the `make deploy-without-build` target works.

zusorio:
When I had tested it with the command helm told me to use it gave me an error but the make command doesn't have the same issue so this should be good

dominik003 (approved):

PR #1569, opened 2024-05-15
docs: Add possible exceptions to API documentation

### LINKED ISSUE #883: Consider custom exceptions in OpenAPI documentation
Custom exception handlers should also appear as possible responses in the auto-generated OpenAPI specification.

MoritzWeber0:
The Collaboration Manager backend can throw a bunch of different errors. Those errors were only documented in the code by scanning the complete call trace, which is unintuitive for users who just want to talk to the API. 

Therefore, I've migrated all legacy occurrences of `fastapi.HTTPException` to our own exception format. Those exceptions can be registered as response for a route using the `responses` attribute. A helper function `api_exceptions` can be used to automatically translate the exceptions to the OpenAPI format. When registered, the exceptions are automatically added to the API documentation. 

This PR mainly focuses on the initial implementation of the behaviour and doesn't register exceptions for all routes. However, I did it for some routes, e.g. the "Request session" route, which can be used as example.

In the API documentation, exceptions are grouped by status codes. The generated responses are visible in the generated documentation clients. The large amount of error responses might be too bloated for some generated clients, e.g. our frontend client. Therefore, the addition of error responses in the OpenAPI schema can be disabled by setting the environment variable `CAPELLACOLLAB_SKIP_OPENAPI_ERROR_RESPONSES=1`.

In SwaggerUI, the individual schemas slightly hidden. They are not displayed in example section, instead, you have to navigate to "Schema" to see all available error responses.

image

In ReDoc, they are better visible: 

image

The best option is the right side in the ReDoc documentation where you can select the error in a dropdown to see the the returned JSON: 
image

Resolves #883.


backend/capellacollab/core/authentication/jwt_bearer.py
```diff
@@ -67,32 +59,15 @@ def validate_token(self, token: str) -> dict[str, t.Any] | None:
...
-                raise fastapi.HTTPException(
-                    status_code=status.HTTP_401_UNAUTHORIZED,
-                    detail={
-                        "err_code": "token_exp",
-                        "reason": "The Signature of the token is expired. Please request a new access token.",
-                    },
-                ) from None
+                raise exceptions.TokenSignatureExpired()
             return None
         except (jwt.JWTError, jwt.JWTClaimsError):
```
dominik003, line 70: In addition, to the above comment, I would handle the `JWTClaimsError` separately, i.e., add another except here with a `exceptions.InvalidClaimsError`.

backend/capellacollab/core/responses.py
```diff
@@ -0,0 +1,108 @@
...
+def api_exceptions(
+    excs: list[exceptions.BaseError] | None = None,
+    include_authentication: bool = False,
+    minimum_role: users_models.Role | None = None,
+    minimum_project_role: projects_users_models.ProjectUserRole | None = None,
+    minimum_project_permission: (
+        projects_users_models.ProjectUserPermission | None
+    ) = None,
+):
+    if os.getenv("CAPELLACOLLAB_SKIP_OPENAPI_ERROR_RESPONSES", "").lower() in (
```
dominik003, line 48: In which cases do we want to skip this?

backend/capellacollab/projects/exceptions.py
```diff
@@ -0,0 +1,39 @@
...
+            err_code="PROJECT_NOT_FOUND",
+        )
+
+
+class ProjectAlreadyExistsError(core_exceptions.BaseError):
+    def __init__(self, project_slug: str):
+        super().__init__(
+            status_code=status.HTTP_409_CONFLICT,
+            title="Project already exists",
+            reason=f"The project with the slug '{project_slug}' (or similar) already exists.",
```
dominik003, line 26:
```suggestion
            reason=f"The project with the slug '{project_slug}' already exists.",
```
I think the "or similar" is more confusing then it actually helps here, so I would just remove it.

dominik003 (approved):
LGTM! There are just a few minor comments that should be addressed before we merge this.

backend/capellacollab/core/responses.py
MoritzWeber0: Described in the PR description.

backend/capellacollab/core/authentication/jwt_bearer.py
```diff
@@ -67,32 +59,15 @@ def validate_token(self, token: str) -> dict[str, t.Any] | None:
...
-                    detail={
-                        "err_code": "JWT_TOKEN_INVALID",
-                        "reason": "The used token is not valid.",
-                    },
-                ) from None
+                raise exceptions.JWTInvalidToken()
             return None
         try:
             return jwt.decode(token, **jwt_cfg)
         except jwt.ExpiredSignatureError:
```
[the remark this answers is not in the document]
MoritzWeber0: Interesting. In theory, the are imported in the jwt package, but my language server doesn't recognized that. Changed it.

backend/capellacollab/core/authentication/jwt_bearer.py
MoritzWeber0: I don't think that this is a message that we need to expose in the frontend. The exceptions is general and covers both cases. However, for debugging purposes, it makes sense to see those exceptions in the log. Therefore, I added exception logging here.

PR #1587, opened 2024-05-27
feat!: Replace bearer with cookie-based authentication

### LINKED ISSUE #1592: Use cookie-based authentication
Currently, we store JWT access and refresh tokens in JavaScript. This is vulnerable to XSS attacks. Instead, cookies with the HttpOnly, Secure and SameSite flags are a safe alternative to store access tokens.

### LINKED ISSUE #1596: Merge authentication provider
Currently we have two different authentication implementations, one for OAuth via azure and a more general variant. The reason for this seems to be that in the case of azure we can use the [Microsoft Authentication Library (MSAL) for Python](https://learn.microsoft.com/en-us/entra/msal/python/). However, most of the logic between the implementations is duplicated, and there is nothing preventing us from using the more general variant in the Azure case as well. Therefore, we want to merge the two authentication flows and remove any provider-specific details.

dominik003:
#### For Administrators

- Support for non-compliant OIDC identity providers has been removed. If you've used OAuth without OIDC before, you can set up Keycloak as intermediate identity provider.
- Since we no longer distinguish between different identity implementations, we have removed the `providers` abstraction in the `values.yaml`. This means that the `providers` field that supported `azure` or `oauth` as values has been removed and instead of `oauth.xy` or `azure.xy` `xy` is used directly (e.g. instead of `oauth.endpoints.wellKnown` it's `endpoints.wellKnown`).
- The OIDC `well-known` endpoint is now a **required** configuration value and must be defined in `endpoints.wellKnown`. As a result of reading all endpoints from the well-known endpoint, the `endpoints.token_issuance` endpoint has been removed. The `endpoints.authorization` endpoint can still be used to overwrite the authorization endpoint from the well known configuration. It's needed for the OAuth mock, is most of the cases you can omit it. 
- The `usernameClaim` configuration was removed. Instead, a more fine-grained mapping is available. If you want to have the same behaviour as it was before the update, set `idpIdentifier` and `username` to the previous value of `usernameClaim`.
  ```yaml
  claimMapping:
      idpIdentifier: sub # Maps the `sub` attribute in the JWT to the `idpIdentifier` attribute in the CCM. `idpIdentifier` is a new attribute to identify users uniquely. The attribute can be patched via the API for existing users.
      username: preferred_username # The username which is displayed in the UI and can be used to add users to projects.
      email: email # Optional email address.
  ```

#### For API Users

- The PATCH `/api/v1/users/{id}/roles` route has been removed. Use the new PATCH `/api/v1/users/{id}` route and pass the new role as attribute in the payload.
- The POST `/api/v1/users` endpoints requires a new attribute `idp_identifier` in the payload.
- The GET `/api/v1/sessions/{session_id}/connection` returns the cookies as header (`Set-Cookie`) instead of an attribute in the payload.

#### Generalized Authentication Provider

Until now the CCM had two types of authentication providers, namely Azure and OAuth. This distinction is no longer necessary because by implementing support for OIDC providers in general, we are also including Azure. Therefore, this PR merges both providers into one generalized provider.

#### Security Improvements

Previously, the CCM stored the identity token and the refresh token in a variable in the front-end authentication service. However, because the CCM is a single-page application, the client has access to the code, and storing these values as service variables is vulnerable to XSS attacks. The new approach now stores both tokens as secure, httponly, and samesite=strict cookies, which means that they are only appended to secure http requests, i.e., *https*, that they are automatically appended to all requests to endpoints that match a defined domain and path, and that there is no way to access the cookies from javascript.

In addition, we now use three different security options/parameters that are supported and recommended in OIDC. 
First, we use a state parameter generated in the backend and validated by the frontend to protect against *Cross-Site Request Forgery (CSRF)* attacks. This is achieved by only requesting a token exchange from the backend if the state returned by the OIDC provider exists in local storage. Theoretically, this is still vulnerable to a combination of XSS and CSRF attacks, but as mentioned in the [OWASP Cross-Site Request Forgery Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#introduction), this is a typical problem.
Second, we use a nonce parameter that is generated in the backend and passed to the OIDC provider, which then includes the value directly in the identity token. Then, when the backend receives the identity token during the token exchange, we validate that the nonce contained in the token matches the generated nonce.
Finally, we also use the *Proof of Key Code Exchange (PKCE)* extension to the OAuth/OIDC authorization code flow, where we generate a code verifier, append a hashed version of the code verifier, called the code challenge, and the hash method, called the code challenge method, to the initial authentication URL, and then send the generated code verifier during the token exchange. The OIDC provider then validates that the code challenge is equal to the hash of the sent code verifier. By using PKCE, we protect against [Authorization Code Interception Attacks](https://is.docs.wso2.com/en/latest/deploy/mitigate-attacks/mitigate-authorization-code-interception-attacks/#mitigate-authorization-code-interception-attacks) and we already comply with the [OAuth 2.1](https://oauth.net/2.1/) requirement to use PKCE.

We also removed the custom implementation of a keystore previously found in `keystore.py` and switched to a JWK client implementation provided by the same library used to validate the JWT tokens. This not only removes a lot of code, but also increases security by using a well-known open source implementation.

#### Removal of the ngx-cookie Package

Previously, we set session cookies in the frontend using the `ngx-cookie' package. This PR removes that package and now also sets the session cookies via set-cookie headers from the backend.

#### TODOs

- [x] Investigate whether we can remove the whole `keystore.py` file and replace it by the pyjwt `PyJWKClient` as seen in [pyjwt oidc login flow](https://pyjwt.readthedocs.io/en/stable/usage.html#oidc-login-flow)
- [x] Find and replace important string comparisons such as state and nonce comparisons to use [hmac.compare_digest](https://docs.python.org/3/library/hmac.html#hmac.compare_digest) to prevent time analysis.
- [x] Evaluate and discuss whether the current approach is vulnerable to Cross Site Request Forgery (CSRF). This includes discussing whether we want to set the `SameSite` attribute to `strict` or `lax`. For more information, see [Same site cookie attribute prevents CSRF attacks](https://www.invicti.com/blog/web-security/same-site-cookie-attribute-prevent-cross-site-request-forgery/).
- [x] Investigate a problem where after a successful login, the frontend displays an "Unauthorized" error coming from the `users/current` endpoint and the navigation bar is not displayed. The problem here is that the endpoint is called before the `/tokens` endpoint is called or its request has finished, so there is no identity token set yet.
- [x] Properly set the allow origins header to our frontend
- [x] Make sure the tests run without the oauth mock run
- [x] Write a proper PR description, including information about the breaking changes
- [x] Test against keycloak and write short documentation on how to use the CCM with keycloak
- [x] Evaluate which tests are needed/helpful and write them
- [x] Add removal of the patch `/{user_id}/role` route as breaking change and mention the new patch `/{user_id}` route
- [x] Fix the tests that are broken due to the refactoring
- [x] Test this PR in the local cluster. For this, first deploy the old version, i.e., without the database changes, and then redeploy with the changes in here.
- [x] Write down the breaking changes introduced in this PR
- [x] When merged, we need to adjust the PR description in #1587 
- [x] Before merging, we need to adjust the documentation added in #1587 (e.g., the steps/configurations required to use Keycloak)

#### Potential follow-up issues

- When implementing redis for caching, we should think about moving the `nonce` and `code_verifier` from the frontend to a proper backend session using redis. We can then also move the `state` back to the backend and add the `redirecTo` to the backend session object. This should further improve security as we are not exposing any information to the frontend that could be used for an attack.
- Move the token refresh logic from the frontend to the backend, i.e. if we get a token expired error during token validation, we automatically try to refresh the token.

Resolves #1596, Resolves #1592


backend/capellacollab/core/authentication/routes.py
```diff
@@ -0,0 +1,158 @@
...
+    response: fastapi.Response,
+    db: orm.Session,
+    access_token: str,
+    refresh_token: str,
+) -> None:
+    validated_access_token = api_key_cookie.JWTAPIKeyCookie().validate_token(
+        access_token
+    )
+
+    assert validated_access_token
```
MoritzWeber0, line 106: Assertion can be removed if auto_error is removed.

backend/capellacollab/core/authentication/routes.py
```diff
@@ -0,0 +1,158 @@
...
+    tokens = flow.get_access_token(token_request.code)
+
+    validate_tokens_and_update_cookies(
+        response, db, tokens["access_token"], tokens["refresh_token"]
+    )
+
+    return True
+
+
+@router.put("/tokens", name="Refresh the access_token")
```
MoritzWeber0, line 57: Instead of a separate route, can't we just automatically refresh the token when the access token is expired? If that would be the case, we could get rid of the auth handling in the frontend.

backend/tests/projects/test_projects_routes.py
```diff
@@ -20,8 +20,9 @@
 
 def test_get_projects_not_authenticated(client: testclient.TestClient):
     response = client.get("/api/v1/projects")
-    assert response.status_code == 403
-    assert response.json() == {"detail": "Not authenticated"}
+    assert response.status_code == 401
+
+    assert response.json()["detail"]["err_code"] == "JWT_TOKEN_INVALID"
```
MoritzWeber0, line 25: JWT_TOKEN_INVALID is probably not the valid error code for it. Should be UnauthenticatedError when no token is provided.

backend/tests/conftest.py
MoritzWeber0, line 1: We need test coverage for the authentication module (routes and flow), but CodeCov will probably also complain about it.

backend/tests/settings/test_alerts.py
```diff
@@ -41,8 +41,8 @@ def test_create_alert_not_authenticated(client: TestClient):
         json={"title": "test", "message": "test", "level": "success"},
     )
 
-    assert response.status_code == 403
-    assert response.json() == {"detail": "Not authenticated"}
+    assert response.status_code == 401
+    assert response.json()["detail"]["err_code"] == "JWT_TOKEN_INVALID"
```
MoritzWeber0, line 45: Same as above.

backend/tests/users/fixtures.py
```diff
@@ -24,7 +23,7 @@ def fixture_executor_name(monkeypatch: pytest.MonkeyPatch) -> str:
     async def bearer_passthrough(self, request: fastapi.Request):
         return name
 
-    monkeypatch.setattr(JWTBearer, "__call__", bearer_passthrough)
+    monkeypatch.setattr(JWTAPIKeyCookie, "__call__", bearer_passthrough)
```
MoritzWeber0, line 26: bearer_passthrough is not the right name anymore. Better authentication_passthrough

backend/tests/users/test_tokens.py
```diff
@@ -47,7 +47,8 @@ async def basic_passthrough(self, request: fastapi.Request):
     monkeypatch.setattr(HTTPBasicAuth, "__call__", basic_passthrough)
     token_string = f"{unauthenticated_user.name}:myTestPassword"
     token = base64.b64encode(token_string.encode("ascii"))
-    basic_response = client.post(
+    basic_client = testclient.TestClient(app)
```
MoritzWeber0, line 50: client is unused as parameter in the function signature.

frontend/src/app/projects/models/init-model/init-model.component.ts
```diff
@@ -92,12 +92,24 @@ export class InitModelComponent implements OnInit {
         map((model: Model) => model.tool),
         switchMap((tool: Tool) =>
           combineLatest([
-            this.toolsService.getToolVersions(tool.id, undefined, undefined, {
-              context: SKIP_ERROR_HANDLING_CONTEXT,
-            }),
-            this.toolsService.getToolNatures(tool.id, undefined, undefined, {
-              context: SKIP_ERROR_HANDLING_CONTEXT,
-            }),
+            this.toolsService.getToolVersions(
```
MoritzWeber0, line 95: Not needed if the access_token is again removed from the API docs.

frontend/src/app/services/auth/auth.service.ts
MoritzWeber0, line 1: We should also set the cookie for sessions in the backend, we can get rid of `ngx-cookie` and the cookieService.

frontend/src/app/services/auth/auth.service.ts
```diff
@@ -5,106 +5,58 @@
...
   logOut() {
-    this._accessToken = '';
-    this._refreshToken = '';
-    if (!environment.production) {
-      this.localStorageService.setValue('access_token', '');
-      this.localStorageService.setValue('refresh_token', '');
-    }
     this.localStorageService.setValue('GUAC_AUTH', '');
-    this.cookieService.remove('access_token', { path: '/prometheus' });
+    this.localStorageService.setValue('loggedIn', '');
```
MoritzWeber0, line 57:
```suggestion
    this.localStorageService.setValue('loggedIn', 'false');
```

loggedIn sounds more like a "boolean" value. I'd therefore also set it to false if not logged in.

helm/config/backend.yaml
```diff
@@ -42,37 +42,18 @@ extensions:
...
-    client:
-      id: "{{ .Values.backend.authentication.azure.client.id }}"
-      secret: "{{ .Values.backend.authentication.azure.client.secret }}"
-  {{ end }}
+  endpoints:
+    authorization: "{{ .Values.backend.authentication.endpoints.authorization }}"
+    wellKnown: "{{ .Values.backend.authentication.endpoints.wellKnown }}"
+  audience: "{{ .Values.backend.authentication.audience }}"
+  scopes:
+  - openid
```
MoritzWeber0, line 52: scopes have to be configurable to be compatible with the OpenID spec :)

frontend/src/app/services/auth/auth.service.ts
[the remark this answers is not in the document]
dominik003: I have reworked the authentication flow so that we no longer use the requests oauthlib library, but use the oauthlib library directly. Now I don't have this problem anymore, but I'll leave this comment open so you can test it again on your site.

backend/capellacollab/core/authentication/api_key_cookie.py
```diff
@@ -15,29 +14,24 @@
...
 
-        if not credentials or credentials.scheme != "Bearer":
+        if not token:
             if self.auto_error:
                 raise exceptions.UnauthenticatedError()
             return None
-        if token_decoded := self.validate_token(credentials.credentials):
+
+        if token_decoded := self.validate_token(token):
             self.initialize_user(token_decoded)
```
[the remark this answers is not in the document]
dominik003: True, I was always just using the admin that was initially created. I have moved the logic of creating a non-existing user to the `validate_id_token` function, so instead of throwing a `UserNotFoundError` we just create the user there. This also has the advantage that we can completely remove the user initialization from the `JWTAPIKeyCookie`, making this class much simpler.

backend/capellacollab/core/authentication/routes.py
```diff
@@ -0,0 +1,158 @@
...
+    set_secure_cookie(response, "refresh_token", refresh_token, "/api/v1")
+
+    if user.role == users_models.Role.ADMIN:
+        set_secure_cookie(response, "access_token", access_token, "/grafana")
+        set_secure_cookie(
+            response, "access_token", access_token, "/prometheus"
+        )
+
+
+def delete_access_token_cookies(response: fastapi.Response):
```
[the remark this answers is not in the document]
dominik003: There was a place where we only deleted the id token but not the refresh token. However, I don't think it's important to delete the refresh token as well, so I renamed this to `delete_token_cookies` and added the refresh token deletion.

frontend/src/app/openapi/api/events.service.ts
```diff
@@ -93,13 +95,14 @@ export class EventsService {
...
      * @param reportProgress flag to report request and response progress.
      */
-    public getEvents(observe?: 'body', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<Array<HistoryEvent>>;
-    public getEvents(observe?: 'response', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<HttpResponse<Array<HistoryEvent>>>;
-    public getEvents(observe?: 'events', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<HttpEvent<Array<HistoryEvent>>>;
-    public getEvents(observe: any = 'body', reportProgress: boolean = false, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<any> {
+    public getEvents(accessToken?: string, observe?: 'body', reportProgress?: boolean, options?: {httpHeaderAccept?: 'application/json', context?: HttpContext, transferCache?: boolean}): Observable<Array<HistoryEvent>>;
```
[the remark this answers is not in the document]
dominik003: Thanks for the hint on how to fix this. The access token is no longer shown as a parameter and the generated openapi client no longer has the access token as an optional argument.

frontend/src/app/services/auth/auth.service.ts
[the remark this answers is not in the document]
dominik003: I was able to at least identify where the "Unauthorized" error is coming from, which is the `users/current` endpoint. The problem here is that it is called before the `/tokens` endpoint is called or its request is finished, so there is no identity token set yet. I'll investigate this issue further next week.

backend/capellacollab/core/authentication/keystore.py
```diff
@@ -1,6 +1,9 @@
 # SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
 # SPDX-License-Identifier: Apache-2.0
 
+# SPDX-FileCopyrightText: Marpaia and contributors
+# SPDX-License-Identifier: Apache-2.0
```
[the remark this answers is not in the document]
dominik003: The problem here was that this required a change to our pre-commit hook, as we do not need some, but exactly our license header to be present. However, I added a todo item to the PR description, which may mean that this whole file is no longer needed.

backend/capellacollab/core/authentication/injectables.py
```diff
@@ -59,32 +60,23 @@ class OpenAPIPersonalAccessToken(OpenAPIFakeBase):
...
-        scheme="bearer",
-    )
-    scheme_name = "JWTBearer"
-
-    __hash__ = OpenAPIFakeBase.__hash__
-
-
 async def get_username(
     request: fastapi.Request,
+    access_token: t.Annotated[str | None, fastapi.Cookie()] = None,
```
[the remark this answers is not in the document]
dominik003: Thanks for the solution. I changed it and now it works correctly.

backend/capellacollab/core/authentication/injectables.py
```diff
@@ -59,32 +60,23 @@ class OpenAPIPersonalAccessToken(OpenAPIFakeBase):
...
     _unused1=fastapi.Depends(OpenAPIPersonalAccessToken()),
-    _unused2=fastapi.Depends(OpenAPIBearerToken()),
 ) -> str:
+    if access_token:
+        username = await api_key_cookie.JWTAPIKeyCookie()(request)
+        assert username
```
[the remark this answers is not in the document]
dominik003: I've removed auto_error, so now the username is always a string.

backend/capellacollab/core/authentication/keystore.py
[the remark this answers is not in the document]
dominik003: The keystore file is no longer needed since we are now using the built-in store of pyJWT.

backend/capellacollab/core/authentication/keystore.py
```diff
@@ -51,8 +48,6 @@ def keys_need_refresh(self) -> bool:
         ) > self.key_refresh_interval
 
     def refresh_keys(self) -> None:
```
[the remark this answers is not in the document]
dominik003: The keystore file is no longer needed since we are now using the built-in store of pyJWT, so this comment is resolved.

backend/capellacollab/core/authentication/routes.py
```diff
@@ -0,0 +1,158 @@
...
+    if token_request.state != state:
+        raise exceptions.StateMismatchError()
+
+    tokens = flow.get_access_token(token_request.code)
+
+    validate_tokens_and_update_cookies(
+        response, db, tokens["access_token"], tokens["refresh_token"]
+    )
+
+    return True
```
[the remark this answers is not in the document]
dominik003: I've removed the return here.

backend/capellacollab/core/authentication/routes.py
```diff
@@ -0,0 +1,158 @@
...
+    if not refresh_token:
+        delete_access_token_cookies(response)
+        raise exceptions.RefreshTokenCookieMissingError()
+
+    tokens = flow.refresh_token(refresh_token)
+    validate_tokens_and_update_cookies(
+        response, db, tokens["access_token"], tokens["refresh_token"]
+    )
+
+    return True
```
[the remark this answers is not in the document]
dominik003: I've removed the return here.

backend/tests/projects/test_projects_routes.py
dominik003: As discussed, this test has been removed because it does not serve a well-defined purpose.

backend/tests/settings/test_alerts.py
dominik003: As discussed, this test has been removed because it does not serve a well-defined purpose.

backend/tests/users/fixtures.py
dominik003: I have renamed it to `cookie_auth_passthrough` as it is still specific to cookies.

backend/tests/users/test_tokens.py
dominik003: Good catch!

backend/capellacollab/core/authentication/routes.py
dominik003: Yes, that should definitely be possible, but I am not sure how much effort it would take to change that. Do you want to include this in this PR or should we do it in a follow up?

MoritzWeber0: Fine for now since it works - can be improved as follow-up.

frontend/src/app/services/auth/auth.service.ts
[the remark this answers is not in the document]
dominik003: At some point this problem started to be persistent, so I now properly set the allow origins to our frontend. This should fix the problem.

dominik003: This has been applied, so the `ngx-cookie` package is now removed.

backend/capellacollab/config/models.py
```diff
@@ -231,21 +231,14 @@ class AuthOauthClientConfig(BaseConfig):
...
-    )
     authorization: str | None = pydantic.Field(
         default=None,
         description=(
             "The URL of the authorization endpoint. "
             "If not set, the URL is read from the well-known endpoint."
         ),
     )
-    well_known: str | None = pydantic.Field(
+    well_known: str = pydantic.Field(
```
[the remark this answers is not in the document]
dominik003: Done

backend/tests/conftest.py
dominik003: I have already added some tests and would like to evaluate next week which tests would still make sense to add. Feel free to add suggestions.

backend/capellacollab/core/authentication/routes.py
MoritzWeber0, line 1: The `router` file is mostly not covered by tests - but an important module of the application. 

At least the call of the function should be tested with the fastapi test client. This can already ensure that we don't accidentally introduce breaking changes in the pydantic objects / validation or the route interface itself. 

In general, this can be combined and used to write proper integration tests, which make more sense than unit tests - since the CCM mostly just passes requests to the OAuth server. 

The response of the requests can be mocked with responses. This ensures that we properly read the well known endpoint etc.

MoritzWeber0 (changes requested):
First part of the review.

backend/capellacollab/core/authentication/api_key_cookie.py
```diff
@@ -0,0 +1,89 @@
...
+                jwt=token,
+                key=signing_key.key,
+                algorithms=self.provider_config.get_supported_signing_algorithms(),
+                audience=self.provider_config.get_client_id(),
+                issuer=self.provider_config.get_issuer(),
+                options={
+                    "verify_exp": True,
+                    "verify_iat": True,
+                    "verify_nbf": True,
+                },
```
MoritzWeber0, line 69:
```suggestion
                options={
                    "require": ["exp", "iat"],
                },
```

The `verify_exp_*` attributes are already true per default since they depend on `verify_signature`, which is set to `True`. The `jwt` documentation mentions that `exp`, `iat`, `nbf` are only checked if they exist in the token. Since `exp` and `iat` are required in OpenID Connect, we add those to the require array.

backend/capellacollab/config/models.py
```diff
@@ -278,34 +255,25 @@ class JWTConfig(BaseConfig):
...
-    provider: t.Literal["oauth"] = pydantic.Field(
-        default="oauth",
-        description="Indicates the use of OAuth for authentication, not to be changed.",
+class AuthenticationConfig(GeneralAuthenticationConfig):
+    endpoints: AuthOauthEndpointsConfig = AuthOauthEndpointsConfig()
+    audience: str = pydantic.Field(default="default")
+    issuer: str = pydantic.Field(default="http://localhost:8083/default")
+    scopes: list[str] | None = pydantic.Field(
+        default=["openid"],
+        description="List of scopes that application neeeds to access the required attributes.",
```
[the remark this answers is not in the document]
dominik003: Done

backend/capellacollab/core/authentication/api_key_cookie.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if not hasattr(self, "_jwks_client"):
+            self.jwks_client = jwt.PyJWKClient(
+                uri=flow.get_auth_endpoints()["jwks_uri"]
+            )
+
+        if not hasattr(self, "_supported_signing_algorithms"):
+            self.supported_signing_algorithms = (
+                flow.get_supported_signing_algorithms()
+            )
```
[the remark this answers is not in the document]
dominik003: There has been a change, the `JWTConfigBorg` now uses the oidc provider config class, so unfortunately it is no longer possible to use the `__new__` method here, since the provider config is passed as a parameter. I think it should be enough to transform this so that we use the `__init__` instead of `__new__`, but either you adjust this during the week, or I'll have a look at it next week.

backend/capellacollab/core/authentication/api_key_cookie.py
```diff
@@ -0,0 +1,75 @@
...
+            signing_key = self.jwt_config.jwks_client.get_signing_key_from_jwt(
+                token
+            )
+
+            return jwt.decode(
+                jwt=token,
+                key=signing_key.key,
+                algorithms=self.jwt_config.supported_signing_algorithms,
+                audience=config.authentication.client.id,
+                issuer=config.authentication.issuer,
```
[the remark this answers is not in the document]
dominik003: The `issuer` is removed from the config and your approach is applied in the `WellKnownOIDCProviderConfig`.

backend/capellacollab/core/authentication/api_key_cookie.py
dominik003: Done

backend/capellacollab/config/models.py
```diff
@@ -278,34 +255,25 @@ class JWTConfig(BaseConfig):
...
 
 
-class OAuthAuthenticationConfig(GeneralAuthenticationConfig):
-    provider: t.Literal["oauth"] = pydantic.Field(
-        default="oauth",
-        description="Indicates the use of OAuth for authentication, not to be changed.",
+class AuthenticationConfig(GeneralAuthenticationConfig):
+    endpoints: AuthOauthEndpointsConfig = AuthOauthEndpointsConfig()
+    audience: str = pydantic.Field(default="default")
+    issuer: str = pydantic.Field(default="http://localhost:8083/default")
```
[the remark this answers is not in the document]
dominik003: Yes, this is already adjusted.

backend/capellacollab/core/authentication/injectables.py
```diff
@@ -59,32 +77,23 @@ class OpenAPIPersonalAccessToken(OpenAPIFakeBase):
...
-    scheme_name = "JWTBearer"
-
-    __hash__ = OpenAPIFakeBase.__hash__
-
-
 async def get_username(
     request: fastapi.Request,
+    oidc_config: oidc_provider.AbstractOIDCProviderConfig = fastapi.Depends(
+        get_oidc_config
+    ),
```
MoritzWeber0, line 84: It doesn't make sense to fetch the OIDC configuration for unauthenticated routes. I've therefore remove the dependency injection and moved the initialization to JWTAPIKeyCookie.

backend/capellacollab/core/authentication/routes.py
MoritzWeber0: Done

PR #1588, opened 2024-05-28
fix: Termination of sessions

MoritzWeber0:
The automatic termination of idle sessions stopped working with #1573. Reason is that the app label was removed from Prometheus.

This commit migrates the old app references to the new session_id label.


MoritzWeber0:
Tested in the staging environment, works.

PR #1604, opened 2024-06-04
chore: Remove unused certificate generation for OAuth mock

MoritzWeber0:
We don't use the certificate generation, so we can remove it.


jamilraichouni (approved):
I have the same change locally and it works.

PR #1659, opened 2024-07-18
docs: Improve help message for ws restore command

jamilraichouni:
Using this subcommand can easily fail, when user do not know that the tarfile should be a gzip compressed file with one top level directory named 'workspace'.

### DIFF
--- a/backend/capellacollab/cli/ws.py
+++ b/backend/capellacollab/cli/ws.py
@@ -188,7 +188,16 @@ def backup(
 @app.command()
 def restore(
     volume_name: str,
-    tarfile: t.Annotated[pathlib.Path, typer.Argument(exists=True)],
+    tarfile: t.Annotated[
+        pathlib.Path,
+        typer.Argument(
+            exists=True,
+            help=(
+                "gzip compressed file with one top level "
+                "directory named 'workspace'"
+            ),
+        ),
+    ],
     namespace: t.Annotated[str, NamespaceOption],
     access_mode: str = "ReadWriteMany",
     storage_class_name: str = "persistent-sessions-csi",


backend/capellacollab/cli/ws.py
```diff
@@ -188,7 +188,16 @@ def backup(
...
 def restore(
     volume_name: str,
-    tarfile: t.Annotated[pathlib.Path, typer.Argument(exists=True)],
+    tarfile: t.Annotated[
+        pathlib.Path,
+        typer.Argument(
+            exists=True,
+            help=(
+                "gzip compressed file with one top level "
+                "directory named 'workspace'"
```
MoritzWeber0, line 197:
```suggestion
                "tarfile generated by the backup CLI command"
```

I wouldn't support manually created tarfiles. The purpose of the command is to restore backups done by the CLI itself.

MoritzWeber0 (changes requested):

backend/capellacollab/cli/ws.py
jamilraichouni: I'd rather keep it as proposed. I can imagine scenarios where people

1
backup a workpace
 
2
modify the backup
 
3
compress the modified backup
 
4
restore the packed workspace

The CLI can deal with a manually archived workspace folder. This is good and useful. It just lacks documentation here.

MoritzWeber0: Accepting a custom archive would mean that we also need proper validation in the restore interface to provide a correct error message in case of wrongly packed archives. Instead, I'd propose to add two new CLI functions: `pack` and `unpack`. Unpack takes the archive and unpacks it into a target directory. Pack takes the directory and packs it again.

PR #1694, opened 2024-08-06
fix(docs): Fix markdown list issues

### LINKED ISSUE #1381: Fix rendering of nested lists in documentation
Solution is described here: https://github.com/mkdocs/mkdocs/issues/545#issuecomment-522196661 

```md

- test
  - nested
```

renders to a flattened list currently.

zusorio:
Use `mdx_truly_sane_lists` to fix nested lists not rendering properly. This breaks `admonition` so move to `callouts` instead.

Fixes #1381

### DIFF
--- a/docs/docs/admin/installation.md
+++ b/docs/docs/admin/installation.md
@@ -36,8 +36,7 @@ future.
 
 === "microK8s"
 
-    !!! info
-        We have tested the instructions with Ubuntu Server 22.04.
+    INFO: We have tested the instructions with Ubuntu Server 22.04.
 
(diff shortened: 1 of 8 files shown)


MoritzWeber0:
Looking at the PR and the issue again, I don't feel very comfortable with introducing a dependency that doesn't seem to be maintained anymore. The last commit was two years ago: https://github.com/radude/mdx_truly_sane_lists.

But still the situation with the requirement for `prettier-ignore` in a bunch of different places is not optimal. I think I found a better solution: Set the indentation of prettier to 4 in the documentation. This seems to make lists working without any additional plugin: https://github.com/DSD-DBS/capella-collab-manager/pull/1697

PR #1710, opened 2024-08-16
feat: Issue session token for usage with API

### LINKED ISSUE #1774: Issue session token for usage with API
Some sessions may want to automate tasks against the CCM API. A session token is issued as PAT and is valid for the duration of the session.

The session token should only have access to read-operations, not write-operations. Therefore, it's required to implement fine-grained tokens before this PR can be merged.

MoritzWeber0:
Some sessions may want to automate tasks against the CCM API. A session token is issued as PAT and is valid for the duration of the session.

The session token should only have access to read-operations, not write-operations. Therefore, it's required to implement fine-grained tokens before this PR can be merged.

Resolves #1774.

This code snippet can be used to fetch the current user sessions in Jupyter:

```py
import os
import requests

backend_url = os.getenv("CAPELLACOLLAB_API_BASE_URL")
user_id = os.getenv("CAPELLACOLLAB_SESSION_REQUESTER_USER_ID")
username = os.getenv("CAPELLACOLLAB_SESSION_REQUESTER_USERNAME")
pat = os.getenv("CAPELLACOLLAB_SESSION_TOKEN")

response = requests.get(f"{backend_url}/v1/users/{user_id}/sessions", auth=(username, pat))

internal_endpoint = response.json()[0]["internal_endpoint"]
requests.get(f"http://{internal_endpoint}").content
```


frontend/package.json
```diff
@@ -105,5 +106,6 @@
       "marked": "$marked"
     }
   },
-  "type": "module"
+  "type": "module",
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
zusorio, line 110: I think corepack got you 😆

backend/capellacollab/sessions/hooks/session_token.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+class SessionTokenIntegration(interface.HookRegistration):
+    """Create a PAT valid for the duration of the session."""
+
+    def configuration_hook(
+        self,
+        request: interface.ConfigurationHookRequest,
+    ) -> interface.ConfigurationHookResult:
+        token, password = tokens_crud.create_token(
```
zusorio, line 20: Is password the right name here?

frontend/package.json
MoritzWeber0: Oh man, is fixed.

backend/capellacollab/sessions/hooks/session_token.py
MoritzWeber0: It's the actual `token` that can be used against the API. Since basic authentication consists of username and password, I'd consider password as correct. It also avoids confusion with the actual token database object.

backend/capellacollab/sessions/hooks/session_token.py
```diff
@@ -0,0 +1,60 @@
...
+            + datetime.timedelta(
+                days=2
+            ),  # Maximum duration is until end of the next day.
+            source="session automation",
+        )
+
+        return interface.ConfigurationHookResult(
+            environment={"CAPELLACOLLAB_SESSION_TOKEN": password},
+            config={"session_token_id": token.id},
+        )
```
zusorio, line 45: As far as I can tell, CAPELLACOLLAB_SESSION_TOKEN is used for tools like guacamole to identify users (because we couldn't do authentication via the API). Would it make sense to use a different name here? At least we should update the docs with the new usage.

MoritzWeber0: Ohhh, that's a really good point. This might actually lead to unwanted behaviour. I'll rename it to `CAPELLACOLLAB_SESSION_API_TOKEN`.

zusorio (approved):

PR #1712, opened 2024-08-16
feat: Add icons to session indicator for a11y

### LINKED ISSUE #1709: Make the session status indicator more accessible
The indicator showing if a session has started, is being created, or has failed uses a colored background in addition to some text.
Unfortunately, the colors can be difficult to differentiate for colorblind people. Here's a simulation of two different states using Chrome's built-in colorblindness simulation:

image
image

While the text is still technically enough to see the status of the session, it would be nicer to add an icon to make the indicator more glanceable for colorblind people.

zusorio:
Using color for the session indicator makes it more difficult to read for colorblind people.
This PR adds icons which help with quickly identifying the session state. Corresponding stories are also added to showcase the new icons.

Closes #1709

### DIFF
--- a/frontend/src/app/sessions/service/session.service.ts
+++ b/frontend/src/app/sessions/service/session.service.ts
@@ -101,6 +101,7 @@ export class SessionService {
 
     let text = state;
     let css = 'warning';
+    let icon = 'pending';
     let success = false;
     switch (state) {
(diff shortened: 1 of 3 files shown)


MoritzWeber0 (approved):
LGTM.

PR #1751, opened 2024-09-04
feat: Add a feedback form

### LINKED ISSUE #1742: Add form for submitting feedback about a session
Allow users to give a session a rating and submit freeform feedback, similar to platforms like MS Teams and Discord.

The feedback should
- ask for a rating
- allow freeform text input
- allow the user to remain anonymous or include their data and be contacted in response

In the background we should:
- send a webhook or email
- include metadata like session type, length, etc

We should be able to show this feedback popup:
- after terminating a session
- every week or so
- manually by pressing a button

zusorio:
Add a feedback form that can be triggered after terminating a session, on a session card, in regular intervals, and in the footer. Feedback can optionally contain freeform text and include the users contact information. Feedback includes an anonymized version of any associated sessions.

Closes #1742


zusorio:
Note that the switch to Material Symbols in #1798 changed the sentiment icons. The names do still exist, but look more intense. If it's merged, the icons here should be updated as following:
`sentiment_very_dissatisfied` -> `sentiment_dissatisfied`
`sentiment_very_satisfied` -> `sentiment_satisfied`

zusorio:
> Note that the switch to Material Symbols in #1798 changed the sentiment icons. The names do still exist, but look more intense. If it's merged, the icons here should be updated as following: `sentiment_very_dissatisfied` -> `sentiment_dissatisfied` `sentiment_very_satisfied` -> `sentiment_satisfied`

This has been done now

zusorio:
I've added tracking for where the feedback dialog was triggered

MoritzWeber0 (approved):
Great PR, thanks a lot! I had some improvement suggestions and implemented them directly. You can have a look at my commit, it also contains a description of all relevant changes.

PR #1764, opened 2024-09-09
fix: Destroy session history subscription

dominik003:
This ensures that the session history subscription is destroyed immediately after the first value is emitted. This ensures that the subscription is not executed multiple times when you leave and return to the session page.


MoritzWeber0 (approved):
Good catch!

PR #1793, opened 2024-09-12
feat: Redesign the pipeline overview

### LINKED ISSUE #1792: Deprecate "Include commit history" option in pipelines
With included commit history, a run can take a long time. It has to be used with caution and is not enabled for any pipeline in our production environment. In addition, The TeamForCapella commit messages are exported by default - it just doesn't contain the changed elements. Those elements can be retrieved via the Git commit diff.

Due to the sometimes hours-long runtimes with this option enabled, we should deprecate it and remove the flag in the backend with v5.0.0.

MoritzWeber0:
- The "Select pipeline" step was removed. Instead, all pipeline actions are displayed directly on the corresponding card.
- A note that the TeamForCapella project name must match exactly was added.
- Error handling was added if Grafana Loki is not enabled.
- Make use of generated OpenAPI schema for pipelines and pipeline runs in the frontend.

- [x] Manual testing
- [x] Stories for pipeline deletion component

Resolves #1792


frontend/src/app/projects/models/backup-settings/trigger-pipeline/trigger-pipeline.component.html
```diff
@@ -30,110 +20,81 @@
...
+                View runs
+              </button>
+              <button
+                mat-raised-button
+                (click)="removePipeline(pipeline)"
+                color="warn"
+              >
+                <mat-icon>delete</mat-icon>
+                Remove pipeline
+              </button>
```
zusorio, line 84: Consider moving the remove button to the left. Ideally, the rightmost button should be the primary action (in this case probably the "Run Pipeline" button).

frontend/src/app/projects/models/backup-settings/create-backup/create-backup.component.html
```diff
@@ -4,70 +4,101 @@
...
+                  <b>Repository:</b> {{ t4cModel.repository.name }} <br />
+                  <b>Instance:</b> {{ t4cModel.repository.instance.name }}
+                </div>
+              </mat-list-option>
+            }
+          </mat-selection-list>
+        </div>
+        <mat-icon>arrow_forward</mat-icon>
+        <div class="collab-card">
+          Please select a Git model.
```
zusorio, line 54: I think this would look better if it was a bold heading with just "Git Model".

frontend/src/app/projects/models/backup-settings/create-backup/create-backup.component.html
```diff
@@ -4,70 +4,101 @@
...
+      }"
+    ></ngx-skeleton-loader>
+  } @else if (
+    (this.gitModelService.gitModels$ | async)?.length &&
+    (this.t4cModelService.t4cModels$ | async)?.length
+  ) {
+    <form [formGroup]="createBackupForm">
+      <div class="flex flex-row flex-wrap items-center gap-2">
+        <div class="collab-card">
+          Please select a T4C model. <br />
```
zusorio, line 28: I think this would look better if it was a bold heading with just "T4C Model". 
The note about having to match could be a little warning card with an icon.

frontend/src/app/projects/models/backup-settings/pipeline-deletion-dialog/pipeline-deletion-dialog.component.html
```diff
@@ -0,0 +1,43 @@
...
+ -->
+
+<div class="dialog">
+  <h2 class="mb-2 text-xl font-medium">
+    Remove a TeamForCapella to Git backup pipeline
+  </h2>
+  <p>
+    Do you want to delete the TeamForCapella to Git backup pipeline
+    {{ data.backup.id }}?
+  </p>
```
zusorio, line 13: As a precaution, this should include more information about the pipeline instead of just the id to prevent accidentally deleting the wrong one.

zusorio (dismissed):
Looks pretty good to me, I noticed a few minor things that could be changed but the new design is already much nicer than the old one!

frontend/src/app/projects/models/backup-settings/trigger-pipeline/trigger-pipeline.component.html
MoritzWeber0: Good hint. Changed it, the order is now: 
image

frontend/src/app/projects/models/backup-settings/create-backup/create-backup.component.html
MoritzWeber0: Done

frontend/src/app/projects/models/backup-settings/pipeline-deletion-dialog/pipeline-deletion-dialog.component.html
MoritzWeber0: Contains more information now: 

image

frontend/src/app/projects/models/backup-settings/create-backup/create-backup.component.html
MoritzWeber0: image

That's how it looks now.

PR #1798, opened 2024-09-12
chore: Update font stack

### LINKED ISSUE #1744: Move from material icons to material symbols
Google's old material icon set is called "Material icons". It's deprecated and no longer gets new icons. The replacement for it is called "Material Symbols". As described in the [Angular Material docs](https://fonts.google.com/icons), we should move to the new icon set.

zusorio:
Get fonts from fontsource: Removes requirements to check in fonts while still self-hosting them. Fonts can also receive updates through npm.

Use Roboto in addition to the standard tailwind font stack: Instead of falling back to only Helvetica Neue and sans-serif, use the tailwind defaults if other fonts could not be loaded.

Add the bold (700) variation of Roboto: Previously, browsers were approximating the bold version. This caused some weird font rendering issues on bold text.

Move to Material Symbols: Google's old icon font (Material Icons) is deprecated. Theoretically, Material Symbols also supports a variable version with customizable fill, weight, etc but as this version is 3MB, I've decided to go with a static version of it.

Closes #1744


MoritzWeber0 (approved):
Looks good and is a easier to maintain than before 👍

PR #1799, opened 2024-09-12
fix: Disable "Connect to session" button when missing session info

### LINKED ISSUE #1785: Disable "Connect to session" button if session information is not available

zusorio:
Disables the "Connect to session" button inside the connection dialog if the session's connection info is not available. The logic to show a toast when there is no session information is removed, as there is no situation where it could be triggered.

Closes #1785


MoritzWeber0 (approved):
LGTM. I've just added another story that covers the loading case. In addition, I renamed "Info" to "Information" as it sounds slightly more formal.

PR #1804, opened 2024-09-16
chore: More migration to new Angular control flow

zusorio:


MoritzWeber0:
Hey @zusorio, great that you're looking into the migration of the control flow. We decided some time ago in the team to do it iteratively and to add stories for each touched component. So make sure that all components moved to the new control flow have stories if they make sense. If you spot CSS that can be moved to Tailwind, that's also a good place to do it. But as far as I can see, you're already on the right track :)

frontend/src/storybook/monitoring.ts
MoritzWeber0, line 1: Since the monitoringService is only used in the pipelines-overview component as of now, I'd also move it to the corresponding stories file. The storybook module is intended for mocked services and objects which are used in multiple stories.

frontend/src/app/settings/core/pipelines-overview/pipelines-overview.component.html
MoritzWeber0, line 1: image

It looks like the skeleton loader for the reason column is missing.

frontend/src/app/projects/create-project/create-project.stories.ts
```diff
@@ -0,0 +1,82 @@
...
+import {
+  mockProjectUsers,
+  MockProjectUserService,
+} from '../../../storybook/project-users';
+import { ProjectUserService } from '../project-detail/project-users/service/project-user.service';
+import { ProjectWrapperService } from '../service/project.service';
+import { CreateProjectComponent } from './create-project.component';
+
+const meta: Meta<CreateProjectComponent> = {
+  title: 'Project Components / Create Project',
```
MoritzWeber0:
```suggestion
  title: 'Project Components/Create Project',
```

Otherwise the files generated by storycap have whitespaces in it.

MoritzWeber0 (approved):
Thanks a million, especially for all of those great new stories 💯

PR #1819, opened 2024-09-21
docs: Add comparision between local Capella and Collaboration Manager

MoritzWeber0:


MoritzWeber0:
Approved by @jamilraichouni via private message.

PR #1828, opened 2024-09-23
fix: Disable Feedback Prompt when logged out, Change order of buttons

zusorio:
Fix a bug where the interval feedback prompt would be shown, even when the user is not logged in. Restore the original order of the feedback buttons by changing the server-side enum.


MoritzWeber0 (approved):
LGTM. As a result, the interval doesn't trigger the dialog anymore after login, only after refresh. This can be an issue for users which don't use the CCM often since they have to login each time. Fixed in https://github.com/DSD-DBS/capella-collab-manager/pull/1829.

PR #1829, opened 2024-09-23
feat: Trigger feedback dialog after login

MoritzWeber0:
As a result of https://github.com/DSD-DBS/capella-collab-manager/pull/1828, the feedback interval doesn't trigger the dialog anymore after login, only after refresh if logged in. This can be an issue for users which don't use the CCM often since they have to login each time. This PR makes sure that the dialog is also triggered after login.

Also fixes another bug that the request to the backend was triggered twice.


zusorio (approved):
LGTM

PR #1850, opened 2024-09-27
feat: Make nginx error pages more user-friendly

MoritzWeber0:
The new 401 error page: 

image

The new 502 error page: 

image

This also fixes a bug that an "500 - Internal server error" page was displayed when the user tried to access a session without session token.

### DIFF
--- a/helm/templates/sessions/sessions.configmap.yaml
+++ b/helm/templates/sessions/sessions.configmap.yaml
@@ -27,6 +27,7 @@ data:
 
             root /usr/share/nginx/html;
             error_page 502 /502.html;
+            error_page 504 /502.html;
             error_page 401 /401.html;
 
             resolver {{ .Values.cluster.dns.service }}.{{ .Values.cluster.dns.namespace }}.svc.cluster.local;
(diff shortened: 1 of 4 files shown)


helm/templates/sessions/sessions.configmap.yaml
```diff
@@ -64,46 +64,138 @@ data:
...
+              Roboto, Oxygen, Ubuntu, Cantarell, "Open Sans", "Helvetica Neue",
+              sans-serif;
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            background-color: #000;
+            color: white;
+            margin: 0;
+            height: 100vh;
+            justify-content: center;
```
zusorio, line 80:
```suggestion
            display: grid;
            place-items: center;
            background-color: #000;
            color: white;
            margin: 0;
            height: 100vh;
```
We can simplify the centering using `grid` and `place-items: center`. As there's only one element in the flexbox the `flex-direction` is not necessary.

helm/templates/sessions/sessions.configmap.yaml
```diff
@@ -64,46 +64,138 @@ data:
...
```
zusorio, line 157:
```suggestion
            display: grid;
            place-items: center;
            background-color: #000;
            color: white;
            margin: 0;
            height: 100vh;
```
Same as above

helm/templates/sessions/sessions.configmap.yaml
MoritzWeber0: Thanks for the hint. It also makes the `viewport` definition unnecessary.

PR #1873, opened 2024-10-02
feat!: Rewrite major parts of the file browser

### LINKED ISSUE #118: File upload improvements
- [x] Iterate through tree more efficient (e.g. along a path instead of the current brute force approach) for functions `addFileToTree` and `removeFileFromTree`
- [x] Save all expanded nodes such that after adding/removing files (and therefore, after collapsing) they are expanded again all at once
- [x] Find a way to allow filenames with white spaces (instead of replacing them with underscores)
- [ ] Allow upload of files with sizes >100MB
- [ ] Use different icons for different file extensions

### LINKED ISSUE #1286: Download single files in file browser
Currently, it's only possible to download directories. Instead, it would be good to download single files too.

### LINKED ISSUE #1732: Remove usage of `NestedTreeControl`
NestedTreeControl is deprecated and will be removed in a future version: https://github.com/DSD-DBS/capella-collab-manager/blob/main/frontend/src/app/sessions/user-sessions-wrapper/active-sessions/file-browser-dialog/file-browser-dialog.component.ts

MoritzWeber0:
Breaking API change: `filename` was renamed to `path` in `/api/v1/sessions/{session_id}/files/download` to make it consistent with the the list files endpoint.

- The UI got an overhaul - margins are fixed and more icons for different cases were added (added files, modified files, expanded folder)
- The workspace is now expanded per default
- Download of single files (not directories) was added, resolves #1286
- Replaced brute force to find a path with path traversal, which is a lot more efficient
- Use the generated OpenAPI client, fix API docs to use ZIP as response
- Migrate to TailwindCSS
- Remove custom "Overwrite file" dialog, replace with ConfirmationDialog.
- Remove `NestedTreeControl`, resolves #1732
- Separate handling for files which replace existing files
- Stream downloading directory directory without loading to memory
- Fix a bug that files with special characters couldn't be downloaded (archive had a size of 0KB)

Resolves #118 except upload limit (will be kept to reduce load on the backend) + different file icons for different extensions (not planned).


frontend/src/app/sessions/user-sessions-wrapper/active-sessions/file-browser-dialog/file-browser-dialog.component.html
```diff
@@ -3,117 +3,124 @@
...
+      [isExpandable]="true"
+      class="!h-8 !min-h-8"
+    >
+      <button
+        mat-icon-button
+        (click)="node.isExpanded = !node.isExpanded"
+        [attr.aria-label]="'Expand ' + node.name"
+        data-testid="folder-button"
+      >
+        <mat-icon class="text-blue-900">folder</mat-icon>
```
zusorio, line 84: I think making this change into `folder_open` when the node is opened could look nice

MoritzWeber0: Makes sense and was easy to implement using the new `node.isExpanded` attribute 👍

PR #1894, opened 2024-10-10
feat: Add feedback anonymity metric to Grafana dashboard

zusorio:


backend/capellacollab/feedback/metrics.py
```diff
@@ -29,5 +29,23 @@ def collect(self) -> t.Iterable[prometheus_client.core.Metric]:
         yield metric
 
 
+class FeedbackAnonymityCollector(prometheus_registry.Collector):
```
MoritzWeber0, line 32: I'd handle it slightly different. Instead of a separate metric, I'd just embed it into the existing metric as label. The result would be: 

```
feedback_count{rating="bad",anonymous="1"} 0.0
feedback_count{rating="bad",anonymous="0"} 0.0
feedback_count{rating="okay",anonymous="1"} 0.0
feedback_count{rating="okay",anonymous="0"} 0.0
feedback_count{rating="good",anonymous="1"} 0.0
feedback_count{rating="good",anonymous="0"} 0.0
```

MoritzWeber0 (changes requested):

backend/capellacollab/feedback/crud.py
```diff
@@ -13,13 +13,27 @@
...
+                    (models.DatabaseFeedback.user_id != None, False),
+                    else_=True,
+                ).label("anonymous"),
+                # pylint: disable=not-callable
+                sa.func.count(),
+            ).group_by(
+                models.DatabaseFeedback.rating,
+                sa.case(
+                    # pylint: disable=singleton-comparison
+                    (models.DatabaseFeedback.user_id != None, False),
```
MoritzWeber0, line 33:
```suggestion
                    (models.DatabaseFeedback.user_id.is_not(None), False),
```

This makes pylint happy and doesn't look worse.

zusorio: Ah, I tried out `is not None` earlier but that made mypy unhappy, I'll use this instead, thanks!

backend/capellacollab/feedback/crud.py
```diff
@@ -13,13 +13,25 @@
...
+                sa.func.count(),
+            ).group_by(
+                models.DatabaseFeedback.rating,
+                sa.case(
+                    (models.DatabaseFeedback.user_id.is_not(None), False),
+                    else_=True,
+                ),
+            )
         ).all()
     }
```
MoritzWeber0, line 36: My proposal would be: 

```py
    anonymity_case = sa.case(
        (models.DatabaseFeedback.user_id.is_(None), True),
        else_=False,
    )

    return {
        (value[0], value[1]): value[2]
        for value in db.execute(
            sa.select(
                models.DatabaseFeedback.rating,
                anonymity_case.label("anonymous"),
                sa.func.count(),
            ).group_by(
                models.DatabaseFeedback.rating,
                anonymity_case,
            )
        ).all()
    }
```

It removed the duplicate definition of the case and also doesn't require disabled pylint rules.

helm/config/grafana/dashboards/feedback.json
MoritzWeber0, line 1: This change breaks the existing dashboards, but is fixable easily. The rating values appear twice since there are now two metrics for each rating. 

We have to group by the rating. For example the query in the pie chart has to look like:

```
round(sum by (rating) (increase(feedback_count[$__range])))
```

MoritzWeber0 (changes requested):

PR #1927, opened 2024-10-25
feat!: Issue JWT for session pre-authentication

MoritzWeber0:
Instead of the non-structured session token, issue a nested JWT containing `session.id`, `user.id`, `user.name`, `user.email `and `user.role`. More claims will be added in the future.

During session connection, the backend issues a signed JWT token. The private key is auto-generated in the backend (if it doesn't exist) and can be exchanged via new CLI endpoints.

The JWT is validated automatically for all requests to HTTP-based sessions. The JWT can be read from the `ccm_session_token` cookie and can be trusted by sessions. It may be used to extract user or session information in the sessions itself.

The validate_token endpoint doesn't require an active database session anymore, reducing network traffic and improving the response times. This effectively makes sessions slightly faster.

One of the best changes is for our developers: It's finally possible to reach sessions from the live-refresh environment again :) 

BREAKING CHANGE: Users with active sessions have to reconnect to their sessions after the update has been rolled out. We recommend to install the update when there are no active sessions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -75,6 +75,7 @@ repos:
           - capellambse
           - typer
           - types-lxml
+          - cryptography
   - repo: local
     hooks:
       - id: pylint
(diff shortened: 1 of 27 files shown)


zusorio (approved):
LGTM

PR #1936, opened 2024-10-29
feat: Add the ability to set a company logo and show environment badge

zusorio:
Add the ability to set a logo that will be displayed next to the Capella text. Add a badge that can show the current environment with different UI depending on if the environment is productive. The badge is enabled during development by default.


frontend/src/app/general/logo/logo.component.html
```diff
@@ -0,0 +1,26 @@
...
+
+<div class="flex select-none items-center gap-2">
+  @if (navBarService.logoUrl$ | async) {
+    <img
+      [src]="navBarService.logoUrl$ | async"
+      aria-hidden="true"
+      class="max-h-8 max-w-24"
+    />
+  }
+  <div class="text-2xl text-primary">Capella</div>
```
MoritzWeber0, line 14:
```suggestion
    <div class="text-sm text-primary font-bold text-center">Capella<br />Collaboration</div>
```

Capella might indicate that we're actually looking at Capella, but it's just the wrapper around. Also from a trademarking perspective, I'd make it clearer, e.g., by using the Capella Collaboration term.

MoritzWeber0 (commented):
The navigation bar looks way cleaner now 👍

frontend/.storybook/main.ts
```diff
@@ -20,6 +20,10 @@ const config: StorybookConfig = {
     name: '@storybook/angular',
     options: {},
   },
+  docs: {
+    autodocs: 'tag',
+  },
```
MoritzWeber0, line 25: That was probably added as part of a wrong rebase conflict resolution and can be removed.

MoritzWeber0 (approved):

PR #1940, opened 2024-10-31
fix: Add correct default for redirect_url when adding a new tool

MoritzWeber0:
Before, the redirect_url was hardcoded to `http://localhost:8080`.


zusorio (approved):
Tested and works on my machine 👍

PR #1944, opened 2024-10-31
feat: Allow setting support project types for tools

### LINKED ISSUE #1938: Limit model tool selection to supported project types
In the global tool configuration, add a flag `supported_types`, which accepts a list of project types, i.e. `general` / `training`. When a model is added to a project, only models matching the project type should be displayed.

zusorio:
Closes #1938

### DIFF
--- a/backend/capellacollab/tools/models.py
+++ b/backend/capellacollab/tools/models.py
@@ -17,6 +17,7 @@ from sqlalchemy import orm
 from capellacollab.core import database
 from capellacollab.core import pydantic as core_pydantic
 from capellacollab.core.database import decorator
+from capellacollab.projects import models as project_models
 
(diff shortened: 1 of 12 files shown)


backend/capellacollab/tools/models.py
```diff
@@ -301,6 +302,13 @@ class ToolSessionConfiguration(core_pydantic.BaseModel):
             description="Configuration for persistent workspaces.",
         )
     )
+    supported_types: list[project_models.ProjectType] = pydantic.Field(
```
MoritzWeber0, line 305: I'd either use a set or add a validator to avoid duplicated in the list.

MoritzWeber0, line 305:
```suggestion
    supported_project_types: list[project_models.ProjectType] = pydantic.Field(
```

Not clear which types we talk about otherwise.

MoritzWeber0 (approved):

PR #1948, opened 2024-11-01
feat: Customize schedule and tz of nightly pipelines via global config

MoritzWeber0:
The configuration only applies to newly created pipelines.


ewuerger (approved):
LGTM.

PR #1949, opened 2024-11-01
fix: Allow longer feedback and warn users if their feedback is too long

zusorio:
The previous limit of 255 was a bit too short, so raise it to 500. Add a client-side warning about the feedback being too long instead of making users run into a server-side error.

### DIFF
--- a/frontend/src/app/sessions/feedback/feedback-dialog/feedback-dialog.component.html
+++ b/frontend/src/app/sessions/feedback/feedback-dialog/feedback-dialog.component.html
@@ -60,6 +60,11 @@
           formControlName="feedbackText"
           data-testid="feedback-text"
         ></textarea>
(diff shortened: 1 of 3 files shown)


frontend/src/app/sessions/feedback/feedback-dialog/feedback-dialog.component.html
```diff
@@ -60,6 +60,11 @@ <h1 class="text-lg font-semibold">
           formControlName="feedbackText"
           data-testid="feedback-text"
         ></textarea>
+        @if (feedbackForm.get("feedbackText")?.hasError("maxlength")) {
```
MoritzWeber0, line 63:
```suggestion
        @if (feedbackForm.controls.feedbackText.hasError("maxlength")) {
```

MoritzWeber0 (approved):

PR #1966, opened 2024-11-07
perf: Add profiling option for backend routes

MoritzWeber0:
It generates an HTML report to better identify slow actions in routes.


zusorio (approved):
Very nice

PR #1967, opened 2024-11-07
feat: Allow setting project type during creation

### LINKED ISSUE #1961: Create a training project
During project creation, it's not possible to select if it should be a training or project: 

image

The option should be added, analogous to the "Modify project" settings: 
image

zusorio:
Closes #1961

### DIFF
--- a/frontend/src/app/projects/create-project/create-project.component.html
+++ b/frontend/src/app/projects/create-project/create-project.component.html
@@ -53,16 +53,33 @@
                   formControlName="visibility"
                   class="flex flex-col"
                 >
-                  <mat-radio-button
-                    *ngFor="
-                      let visibility of projectService.getAvailableVisibilities()
-                    "
-                    [value]="visibility"
-                  >
-                    {{
-                      projectService.getProjectVisibilityDescription(visibility)
-                    }}
-                  </mat-radio-button>
+                  @for (
+                    visibility of projectService.getAvailableVisibilities();
+                    track visibility
+                  ) {
+                    <mat-radio-button [value]="visibility">
+                      {{
+                        projectService.getProjectVisibilityDescription(
+                          visibility
+                        )
+                      }}
+                    </mat-radio-button>
+                  }
+                </mat-radio-group>
+              </div>
+            </fieldset>
+            <fieldset>
+              <div class="flex flex-col">
+                <legend>Project type</legend>
+                <mat-radio-group formControlName="type" class="flex flex-col">
+                  @for (
+                    type of projectService.getAvailableProjectTypes();
+                    track type
+                  ) {
+                    <mat-radio-button [value]="type">
+                      {{ projectService.getProjectTypeDescription(type) }}
+                    </mat-radio-button>
+                  }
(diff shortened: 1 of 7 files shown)


frontend/package.json
```diff
@@ -80,5 +80,6 @@
     "tailwindcss": "^3.4.14",
     "typescript": "^5.5.4",
     "typescript-eslint": "8.12.2"
-  }
+  },
+  "packageManager": "pnpm@9.11.0+sha512.0a203ffaed5a3f63242cd064c8fb5892366c103e328079318f78062f24ea8c9d50bc6a47aa3567cabefd824d170e78fa2745ed1f16b132e16436146b7688f19b"
```
MoritzWeber0, line 84: Was this change intended?

frontend/src/app/projects/create-project/create-project.component.html
```diff
@@ -66,6 +66,21 @@ <h3 class="flex justify-center sm:hidden">Create new project</h3>
                 </mat-radio-group>
               </div>
             </fieldset>
+            <fieldset>
+              <div class="flex flex-col">
+                <legend>Project type</legend>
+                <mat-radio-group formControlName="type" class="flex flex-col">
+                  <mat-radio-button
+                    *ngFor="
```
MoritzWeber0, line 74: Please use the new control flow: https://angular.dev/guide/templates/control-flow#repeat-content-with-the-for-block

MoritzWeber0 (changes requested):

frontend/package.json
zusorio: Nope, corepack randomly adds that and i forgot to remove it this time. I need to find a way to turn that off at some point.

frontend/src/app/projects/create-project/create-project.component.html
```diff
@@ -53,16 +53,33 @@ <h3 class="flex justify-center sm:hidden">Create new project</h3>
...
+                        )
+                      }}
+                    </mat-radio-button>
+                  }
+                </mat-radio-group>
+              </div>
+            </fieldset>
+            <fieldset>
+              <div class="flex flex-col">
+                <legend>Project type</legend>
```
MoritzWeber0, line 73: image

The gap between the title and the radios looks a bit compressed. This is due to the description of General in the brackets, which expands to two lines. 

I'd either reduce the length of the description, e. g., by removing the "multiple" word. Or alternatively add the description always in a new line with smaller text and dark grey color.

frontend/src/storybook/project.ts
```diff
@@ -51,6 +52,14 @@ export class MockProjectWrapperService
   getAvailableVisibilities(): Visibility[] {
     return Object.keys(ProjectVisibilityDescriptions) as Visibility[];
   }
+
+  getProjectTypeDescription(type: ProjectType): string {
```
MoritzWeber0, line 56:
```suggestion
  get projectTypeDescription(type: ProjectType): string {
```

You can use getter for it and the next function.

MoritzWeber0 (approved):

PR #1981, opened 2024-11-12
feat: Add the ability to override resources on a per-user basis

### LINKED ISSUE #1951: Overwrite resources for individual tool users
Since our production instances have an increased number of users, the user base got also diverse. We have many different resource requirements for users.

Instead of fixed resources, add different resource profiles for tools: 

```yml
config:
  resources:
    profiles:
      default:
        cpu:
          requests: 0.5
          limits: 1
        memory:
          requests: 2Gi
          limits: 4Gi
      high:
        cpu:
          requests: 1.2
          limits: 2
        memory:
          requests: 4.5Gi
          limits: 7Gi
        usernames: 
          - janedoe
          - johndoe
```

The default profile is required and doesn't apply to all users per default. Additional profiles are allowed and require a `usernames` attribute. A username can only be listed in one profile. If the user is listed in the profile, the profile will apply for the user. The tool configuration has to be updated for removed and auto-renamed users.

Open questions: 
- Do we want to use the user id instead of the username in the list? The we don't need to handle username changes, but it's less intuitive.
- Do we want to do a proper database mapping to users? Then we would need a UI instead of YAML editor.

zusorio:
Closes #1951


backend/capellacollab/tools/models.py
```diff
@@ -240,6 +240,70 @@ class Resources(core_pydantic.BaseModelStrict):
...
     )
+    usernames: list[str] | None = pydantic.Field(
+        default=None,
+        description="List of usernames, which are allowed to use this resource profile.",
+    )
+
+
+class Resources(core_pydantic.BaseModelStrict):
+    profiles: dict[str, ResourceProfile] = pydantic.Field(
+        default={"default": ResourceProfile()},
```
MoritzWeber0, line 251:
```suggestion
    profiles: dict[str, ResourceProfile] = pydantic.Field(
        default={"default": ResourceProfile()},
```

```suggestion
    profiles: dict[str, DefaultResourceProfile | AdditionalResourceProfile] = pydantic.Field(
        default={"default": DefaultResourceProfile()},
```

The goal is to remove the `usernames` attribute from the default profile. It doesn't make munch sense and might lead to confusion. We just need to make sure that DefaultResourceProfile is not used as type for additional profile via validation. 

Alternatively, I'd propose to have the default_profile as separate key and add additional_profiles as list. This would have a better and clear type separation.

backend/capellacollab/tools/models.py
```diff
@@ -243,6 +243,60 @@ class Resources(core_pydantic.BaseModelStrict):
...
+
+        for profile in self.additional.values():
+            if username in profile.usernames:
+                return profile
+
+        return self.default_profile
+
+    @pydantic.field_validator("additional")
+    @classmethod
+    def check_additional_profiles(
```
MoritzWeber0, line 277: A validation if the user exists is missing.

backend/capellacollab/users/crud.py
```diff
@@ -80,6 +81,8 @@ def create_user(
 def update_user(
```
MoritzWeber0, line 81: If the user is deleted, the username should be removed from the list.

MoritzWeber0 (changes requested):

PR #2006, opened 2024-11-15
feat: Redesign the active sessions view and rework idletime

### LINKED ISSUE #1486: Display time until session will be terminated
Users want to see the time left to termination because of inactivity to be able to react e.g. by performing some action in the session to avoid termination.

Originally reported in https://github.com/DSD-DBS/capella-collab-manager/issues/621.

### LINKED ISSUE #1783: Use relative instead of absolute datetimes
Currently, whenever we show datetimes in the frontend, we use one of two rather long formats:
- Thu, 05 Sep 2024 15:06:31
- 9/5/2024 at 03:06 PM

A lot of these could be replaced with relative dates like:
- 4 hours ago
- 2 days ago

For general use, a relative timestamp is easier to understand and more relevant. When hovering over these, we could still show the exact date and time.

zusorio:
Rework idletime:
- Renamed idletime to idle_status object
- Also return termination time from config
- Show card on active session if session will be terminated soon

Overhaul parts of the session view:
- Show times relative to now
- Show session info as one text block instead of split
- Redesign last requested session and active session card
- Show additional info on active session cards as mini cards

Closes #1783
Closes #1486


frontend/src/app/general/relative-time/relative-time.component.html
MoritzWeber0, line 1: I'd prefer a pipe instead of a component for relative times. Components are mainly used to apply styles while pipes are used to serialize content.

MoritzWeber0, line 1: The relative time will break the Stories - since they rely on absolute times and relative times will occur in changes every month (e. g., 1 month ago will change to 2 months ago).

frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-session-history/create-session-history.component.html
MoritzWeber0, line 1: I prefer the previous design and would only change the absolute time to a relative time. The reason is that the previous design was better structured and it was easier to extract specific information. For example, a user might search for a specific tool. In the previous design it was only necessary to read the "Tool" column while it's now necessary to read the whole text. 

But that's personal preference and I'll create a team internal poll decision.

backend/capellacollab/sessions/injection.py
```diff
@@ -3,42 +3,60 @@
...
 from capellacollab import core
 from capellacollab.config import config
+from capellacollab.core import pydantic as core_pydantic
 
 log = logging.getLogger(__name__)
 
 
-def get_last_seen(sid: str) -> str:
-    """Return project session last seen activity"""
+# Has to be in here because of circular imports
```
MoritzWeber0, line 16: In case of circular imports, we usually just create a `models2.py`. This avoid a circular imports and is more consistent with the existing structure.

backend/capellacollab/sessions/injection.py
```diff
@@ -3,42 +3,60 @@
...
+class IdleState(core_pydantic.BaseModel):
+    available: bool
+    idle_for_minutes: float | None = pydantic.Field(default=None)
```
MoritzWeber0, line 19:
```suggestion
    idle_for_minutes: int | None = pydantic.Field(default=None)
```

We collect the idletime only every 30 seconds, therefore a float would be inaccurate.

frontend/src/app/sessions/session-overview/session-overview.component.html
```diff
@@ -53,7 +53,21 @@
...
+            Never connected
+          } @else {
+            <app-relative-time
+              [date]="
+                subMinutes(Date.now(), element.idle_state.idle_for_minutes!)
+              "
+            />
+          }
+        } @else {
+          {{ element.idle_state.unavailable_reason! }}
```
MoritzWeber0, line 68:
```suggestion
          {{ element.idle_state.unavailable_reason }}
```

The assertion is not needed here, it works without.

frontend/src/app/sessions/user-sessions-wrapper/active-sessions/active-sessions.component.html
```diff
@@ -106,52 +114,79 @@ <h2>Read-Only session</h2>
...
-            Connection Method: {{ session.connection_method?.name }}
-          </div>
+
+          @if (session.idle_state.available) {
+            @let remaining_minutes =
+              session.idle_state.idle_for_minutes! === -1
+                ? session.idle_state.terminate_after_minutes! -
+                  differenceInMinutes(session.created_at, Date.now())
+                : session.idle_state.terminate_after_minutes! -
+                  session.idle_state.idle_for_minutes!;
```
MoritzWeber0, line 139: That's a lot of logic for the HTML template, I'd prefer to have descriptive function in the ts-file.

frontend/src/app/sessions/user-sessions-wrapper/active-sessions/active-sessions.component.html
```diff
@@ -208,7 +246,8 @@ <h2>Read-Only session</h2>
                 ).success
               "
             >
-              File browser <mat-icon> insert_drive_file</mat-icon>
+              File browser
+              <mat-icon> insert_drive_file</mat-icon>
```
MoritzWeber0, line 250:
```suggestion
              <mat-icon>insert_drive_file</mat-icon>
```

backend/capellacollab/sessions/injection.py
```diff
@@ -3,42 +3,60 @@
...
+    terminate_after_minutes: float
```
MoritzWeber0, line 20:
```suggestion
    terminate_after_minutes: int
```

In the AppConfig, it's int

backend/capellacollab/sessions/injection.py
MoritzWeber0, line 19: Add a field description that the -1 value means "Never connected"

MoritzWeber0 (commented):
General structure looks good. Design for the session history has to be discussed team-internally before.

frontend/src/app/sessions/user-sessions-wrapper/active-sessions/active-sessions.component.html
MoritzWeber0, line 1: The session state is no longer aligned with the checkbox: 

image

frontend/src/app/general/relative-time/relative-time.component.html
zusorio: I'd like to use a pipe here as well, but I don't think the tooltip is possible with pipes. If it is, I can definitely look into that though.

frontend/src/app/sessions/user-sessions-wrapper/create-sessions/create-session-history/create-session-history.component.html
MoritzWeber0: Team has decided: Your proposal won. I'll mark the conversation as resolved.

frontend/src/app/general/relative-time/relative-time.component.html
MoritzWeber0: Fair enough, I missed the tooltip. Let's stick to a component then.

frontend/src/app/sessions/user-sessions-wrapper/active-sessions/active-sessions.component.html
MoritzWeber0, line 1: image


Here is a small overflow.

backend/capellacollab/sessions/injection.py
```diff
@@ -7,38 +7,50 @@
...
 from capellacollab import core
 from capellacollab.config import config
+from capellacollab.sessions import models2 as sessions_models2
 
 log = logging.getLogger(__name__)
 
 
-def get_last_seen(sid: str) -> str:
-    """Return project session last seen activity"""
+# Has to be in here because of circular imports
```
MoritzWeber0, line 15: You probably missed the comment when removing the class.

MoritzWeber0 (approved):

PR #2031, opened 2024-11-20
feat: Don't show seconds on last seen and make creation time relative

zusorio:
- Don't show seconds on last seen
- Make creation date relative
- Fix Story not having mock date


MoritzWeber0 (approved):
Very nice! 🥳

PR #2032, opened 2024-11-21
chore: Change idle time unavailable message if metrics not found

MoritzWeber0:
"Unknown session" could be interpreted as "The session doesn't exist", but that is not true. The session exists and only the metrics are not available.


zusorio (approved):
Much clearer, nice 👍

PR #2040, opened 2024-11-26
feat: Allow users to copy t4c token from session viewer

### LINKED ISSUE #2010: Add button to copy TeamForCapella session token from Internal Session Viewer
When users connect to the Internal Session Viewer, they don't use the connection dialog and don't see the TeamForCapella token. We should add a button to the title bar of the Internal Session Viewer (with tooltip) to copy the corresponding T4C token to the clipboard for usage in the session.

zusorio:
Closes #2010


MoritzWeber0 (commented):
Nice change, this will make the life of our users a bit easier! 👍

PR #2042, opened 2024-11-28
feat: Change feedback prompt for trainings

zusorio:


frontend/src/app/sessions/feedback/feedback-dialog/feedback-dialog.component.ts
```diff
@@ -102,6 +102,24 @@ export class FeedbackDialogComponent {
     return Object.values(FeedbackRating);
   }
 
+  get promptText() {
+    if (
+      this.data.sessions.some(
+        (session) =>
+          session.version.tool.config.supported_project_types.length === 1 &&
+          session.version.tool.config.supported_project_types[0] === 'training',
```
MoritzWeber0, line 110: A session is not necessarily a training session if the tool supports trainings. Instead, check if the provisioned project is a training.

Also, usually a training contains of more than one session. Therefore, the case is usually not applied.

MoritzWeber0 (changes requested):

MoritzWeber0 (approved):

Issue #2099, opened 2024-12-18
Add Grafana dashboard for beta users 


MoritzWeber0:
It would be nice to track the number of registered beta users over time. Also would be cool to filter feedback for beta / non-beta status to see if beta users are happier than non-beta users.

PR #2110, opened 2025-01-08
feat: Add a dashboard to show users to grafana

zusorio:
This solves the users part of #2099 but doesn't add the feedback-related features.

### DIFF
--- a/backend/capellacollab/users/metrics.py
+++ b/backend/capellacollab/users/metrics.py
@@ -1,0 +1,33 @@
+# SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
+# SPDX-License-Identifier: Apache-2.0
+
+import typing as t
+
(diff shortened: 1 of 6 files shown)


backend/capellacollab/users/metrics.py
```diff
@@ -0,0 +1,33 @@
...
+import prometheus_client
+import prometheus_client.core
+from prometheus_client import registry as prometheus_registry
+
+from capellacollab.core import database
+
+from . import crud
+
+
+class UserCountCollector(prometheus_registry.Collector):
```
MoritzWeber0, line 15: Could you write a small test for it; like for the other metrics collectors?

MoritzWeber0 (commented):
LGTM; but we should make codecov happy :)

MoritzWeber0 (approved):

PR #2117, opened 2025-01-13
feat: Redesign Model Card

### LINKED ISSUE #1353: Auto-fill tool/version when approaching from the model card in a project
When clicking the "Request persistent session" button on a model card, we should auto-fill the selection with the matching tool/version. In addition, it would be good to directly auto-request the session.

zusorio:
Re-order the buttons on the model card, moving some infrequently used ones to an overflow menu. Modify the "Create Persistent Session" button to prompt for the connection method and then immediately create the session.

Closes #1353

### DIFF
--- a/frontend/src/app/projects/project-detail/model-overview/model-overview.component.html
+++ b/frontend/src/app/projects/project-detail/model-overview/model-overview.component.html
@@ -122,71 +122,25 @@
           }
         </div>
         <div class="m-2.5">
-          @if (userService.user?.role === "administrator") {
-            <a
-              mat-mini-fab
-              color="primary"
-              matTooltip="Model restrictions"
-              class="!m-1.5"
-              [routerLink]="['model', model.slug, 'restrictions']"
-            >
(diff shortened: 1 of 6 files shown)


frontend/src/app/projects/project-detail/model-overview/model-overview.component.html
```diff
@@ -122,59 +122,37 @@ <h2 class="text-xl font-medium">Models</h2>
...
-              class="!m-1.5"
-              [routerLink]="['model', model.slug, 'metadata']"
-            >
-              <mat-icon>settings</mat-icon>
-            </a>
+          @if (
+            !project?.is_archived &&
+            project?.type !== "training" &&
+            model.t4c_models &&
+            projectUserService.verifyPermission("write")
```
MoritzWeber0, line 129:
```suggestion
            projectUserService.verifyPermission("write") &&
            model.version
```

If the model version is not set, we can't request a session. Fixed it already.

MoritzWeber0 (approved):

PR #2122, opened 2025-01-13
feat: Move from highlight.js to shiki

zusorio:
Replace highlight.js with shiki for the model diagram code block. This has some major advantages
- Main bundle size is reduced by **1MB**
- Shiki's highlighter is more capable and highlights more stuff
- We no longer need to bundle highlight.js' CSS


frontend/package.json
```diff
@@ -103,6 +103,9 @@
   "overrides": {
     "marked-gfm-heading-id": {
       "marked": "$marked"
+    },
+    "@analogjs/platform": {
+      "shiki": "^2.1.0"
```
MoritzWeber0, line 108:
```suggestion
      "shiki": "$shiki"
```

This automatically inherits the value from the dependencies list and always keeps it consistent.

MoritzWeber0 (changes requested):
One small recommendation to keep the package version consistent. Please rebase on main to fix the reg-actions pipeline run. Otherwise looks good 👍

PR #2124, opened 2025-01-14
feat: Rework Monaco editor integration

### LINKED ISSUE #2089: Add support for Monaco YAML in tool configuration
We already have a JSON schema for the tool configuration. It would be great to offer autocomplete, suggestions and show available options in the editor.

I failed to get https://github.com/remcohaszing/monaco-yaml running so far. But we should give it another try, especially due to the changed build with Vite.

zusorio:
Until now, we were importing Monaco directly which massively increased our bundle size. By moving to
 ngx-monaco-editor-v2, Monaco is loaded on-demand. Also add support for monaco-yaml, which provides
 Intellisense for our yaml configs using OpenAPI schemas. This is also our first component using
 signals.

 Closes #2089

### DIFF
--- a/frontend/src/app/helpers/editor/editor.component.ts
+++ b/frontend/src/app/helpers/editor/editor.component.ts
@@ -3,71 +3,93 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import {
-  AfterViewInit,
   Component,
+  computed,
+  effect,
   ElementRef,
   EventEmitter,
   HostListener,
-  Input,
-  NgZone,
+  input,
+  model,
   Output,
-  ViewChild,
+  signal,
 } from '@angular/core';
-import * as monaco from 'monaco-editor';
+import { FormsModule } from '@angular/forms';
+import { type editor } from 'monaco-editor';
+import { MonacoEditorModule, NgxEditorModel } from 'ngx-monaco-editor-v2';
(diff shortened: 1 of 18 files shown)


frontend/src/app/helpers/editor/editor.component.ts
```diff
@@ -3,71 +3,93 @@
...
+    scrollBeyondLastLine: false,
+    automaticLayout: true,
+    tabSize: 2,
+    fixedOverflowWidgets: true,
+  };
+
+  editorModel = computed<NgxEditorModel>(() => ({
+    value: this.initialValue(),
+    language: 'yaml',
+    uri: URI.parse(`file:///${this.context()}.${this.type()}.yaml`),
```
MoritzWeber0, line 57:
```suggestion
    uri: `file:///${this.context()}.${this.type()}.yaml`,
```

And get rid of `vscode-uri`, which seems to imported just for this case.

MoritzWeber0 (commented):
Great work, love it!

frontend/src/app/helpers/editor/editor.component.ts
zusorio: Without `vscode-uri` error highlighting doesn't work, and there are error logs in the browser console

MoritzWeber0 (approved):

PR #2158, opened 2025-01-31
build: Use gzip

### LINKED ISSUE #2126: Serve files using gzip
Currently, we don't compress our files at all which increases network payloads significantly

zusorio:
Fixes #2126

### DIFF
--- a/frontend/nginx.conf
+++ b/frontend/nginx.conf
@@ -16,6 +16,12 @@ http {
 
     access_log /var/log/nginx/nginx-frontend.log;
 
+    gzip on;
+    gzip_types text/plain text/css application/json text/xml application/xml text/javascript;
+    gzip_min_length 256;
+    gzip_comp_level 5;
+    gzip_vary on;
+
     server {
         listen 8080 default_server;
         server_name _;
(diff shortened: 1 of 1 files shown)


frontend/nginx.conf
```diff
@@ -16,11 +16,22 @@ http {
 
     access_log /var/log/nginx/nginx-frontend.log;
 
+    gzip on;
+    gzip_types text/plain text/css application/json application/javascript text/xml application/xml application/xml+rss text/javascript;
```
MoritzWeber0, line 20:
```suggestion
    gzip_types text/plain text/css application/json text/xml application/xml application/xml+rss text/javascript;
```

`application/javascript` is considered legacy, it's probably safe to remove it. 
`application/xml+rss`: We don't have RSS content

frontend/nginx.conf
```diff
@@ -16,11 +16,22 @@ http {
...
+    gzip_min_length 256;
```
MoritzWeber0, line 21: Is there a reason to not use the default value of 20?

zusorio: For file sizes that small, the extra cpu overhead is usually not worth the size reduction

PR #2168, opened 2025-02-05
build: Use uv as package manager

MoritzWeber0:


.github/workflows/lint.yml
```diff
@@ -45,13 +45,15 @@ jobs:
           cache: pip
           cache-dependency-path: ./backend/pyproject.toml
           python-version: '3.12'
+      - name: Install uv
+        uses: astral-sh/setup-uv@v5
       - name: Install pre-commit
         run: |-
-          python -m pip install pre-commit
+          uv tool install pre-commit --with pre-commit-uv
```
zusorio, line 52: Theoretically, we don't actually have to install pre-commit, since we only use it once.
We could remove this and run `uvx --with pre-commit-uv pre-commit run --all-files` in the 'Run pre-commit' step

zusorio (changes requested):
Readthedocs is still using pip instead of uv (and the backend but that's addressed by a separate PR). Py-capellambse already uses uv for building Readthedocs, you can check that out as an example.

MoritzWeber0:
> Readthedocs is still using pip instead of uv (and the backend but that's addressed by a separate PR). Py-capellambse already uses uv for building Readthedocs, you can check that out as an example.

The overhead of installing `uv` is probably larger than the few packages we install via `pip`. The Python packages are cached in readthedocs, not sure if that would still work with uv. I would also leave the installation to readthedocs, they will use what makes the most sense to them, maybe they'll switch to `uv` in the future.

zusorio (approved):

PR #2171, opened 2025-02-07
build: Use `uv` for Python installation in backend Docker image

MoritzWeber0:
Use `uv` for the Python installation in the backend Docker image. This restores compatibility with the debian:bookworm-slim image, which was used in the Gitlab CI template.

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -1,44 +1,51 @@
 # SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
 # SPDX-License-Identifier: Apache-2.0
 
-ARG BASE_IMAGE=python:3.12-slim-bookworm
+ARG BASE_IMAGE=debian:bookworm-slim
 FROM $BASE_IMAGE
 
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 ENV SHELL=/bin/bash
 
+EXPOSE 8000
+
 USER root
 
-RUN apt-get update && \
-    apt-get upgrade --yes && \
-    apt-get install --yes \
-    ca-certificates \
-    git-lfs \
-    libpq-dev \
-    unzip \
-    && rm -rf /var/lib/apt/lists/*
+ADD https://astral.sh/uv/install.sh /uv-installer.sh
+ENV UV_UNMANAGED_INSTALL=/usr/bin/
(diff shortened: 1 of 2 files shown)


backend/Dockerfile
```diff
@@ -1,44 +1,48 @@
...
+EXPOSE 8000
+
 USER root
 
-RUN apt-get update && \
-    apt-get upgrade --yes && \
-    apt-get install --yes \
+ADD https://astral.sh/uv/install.sh /uv-installer.sh
+ENV UV_UNMANAGED_INSTALL=/usr/bin/
+ENV UV_PYTHON_INSTALL_DIR=/etc/python
```
zusorio, line 16: Is there a chance we'd be able to install uv in our restricted environment using `COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/`? That way, we wouldn't need to install curl and it seems more robust than using the install script.

backend/startup.sh
```diff
@@ -7,4 +7,4 @@ set -e
 # When running inside the cluster, the k8s service host should not use the proxy
 export no_proxy=$no_proxy,$KUBERNETES_SERVICE_HOST,$no_proxy_additional,svc.cluster.local
 export NO_PROXY=$NO_PROXY,$KUBERNETES_SERVICE_HOST,$no_proxy_additional,svc.cluster.local
-uvicorn capellacollab.__main__:app --host 0.0.0.0 --forwarded-allow-ips '*'
+/opt/backend/.venv/bin/uvicorn capellacollab.__main__:app --host 0.0.0.0 --forwarded-allow-ips '*'
```
zusorio, line 10: Is there a reason this is not `uv run uvicorn`?

backend/Dockerfile
MoritzWeber0: No, unfortunately not. `ghrc.io` is not available from certain build environments, so we would need to make it configurable. Arguments in the `--from` parameter are not supported by Docker unfortunately.

backend/startup.sh
MoritzWeber0: I tried that, but it didn't work. `uv` tried to rebuild the backend (I suppose due to the changed user), which is not what we want.

zusorio (approved):

PR #2173, opened 2025-02-07
fix: Make sure gzip is applied to js files

zusorio:
We removed the application/javascript filetype because it's considered legacy. Unfortunately, nginx uses it to decide if it should compress .js files so they weren't being compressed.


MoritzWeber0 (approved):

MoritzWeber0:
I should have expected something like this from nginx.

PR #2185, opened 2025-02-10
feat: Add resource recommendations for all deployments

MoritzWeber0:
Add VPAs for all deployments to get resource recommendations.


helm/templates/mock/oauth.vpa.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: autoscaling.k8s.io/v1
+kind: VerticalPodAutoscaler
+metadata:
+  name: {{ .Release.Name }}-smtp-mock
+  namespace: {{ .Release.Namespace }}
+spec:
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ .Release.Name }}-smtp-mock
```
zusorio, line 14: I think this is flipped

PR #2186, opened 2025-02-10
fix: Add uniqueness constraint for project users

MoritzWeber0:
_It's weird._ 

In our production database the primary key for the `project_user_association` table was missing. In addition, the check to add users twice didn't work. This led to users being added multiple times to projects. This PR fixes the mess.


zusorio (changes requested):
Could you add a test that makes sure that these constraints are being applied?

zusorio (approved):
Tested it by manipulating my local db and works great!

PR #2201, opened 2025-02-13
feat: Add session_id as label to session logs

MoritzWeber0:
Enable log filtering by specific session ID.


zusorio (approved):
This was annoying me when using the log explorer, great change!

PR #2203, opened 2025-02-13
feat: Pre-compress frontend with brotli

### LINKED ISSUE #2170: Investigate pre-compressing files with brotli during build
Brotli is more efficient than gzip, but too slow to run on-the-fly. We should see if we can pre-compress files during the build.

zusorio:
Closes #2170

### DIFF
--- a/frontend/nginx.conf
+++ b/frontend/nginx.conf
@@ -21,6 +21,7 @@ http {
     gzip_min_length 256;
     gzip_comp_level 5;
     gzip_vary on;
+    brotli_static on;
 
     server {
         listen 8080 default_server;
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -92,6 +92,7 @@
     "prettier-plugin-classnames": "^0.7.6",
     "prettier-plugin-merge": "^0.7.2",
     "prettier-plugin-tailwindcss": "^0.6.11",
+    "rollup-plugin-gzip": "^4.0.1",
     "shiki": "^2.3.2",
     "storybook": "^8.5.3",
     "tailwindcss": "^3.4.17",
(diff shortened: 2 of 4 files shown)


frontend/Caddyfile
```diff
@@ -0,0 +1,22 @@
...
+        }
+
+        root * /srv
+
+        # Serve static assets with aggressive caching
+        header /assets/* Cache-Control "public, max-age=31536000, immutable"
+
+        try_files {path} /index.html
+        # Force browsers to always load index.html
+        header /index.html Cache-Control "public, max-age=0, must-revalidate"
```
MoritzWeber0, line 16: Isn't this the default?

frontend/Dockerfile
```diff
@@ -18,12 +18,21 @@ RUN pnpm install --ignore-scripts
...
+FROM caddy:2.9-alpine
 
 EXPOSE 8080
 
-COPY nginx.conf /etc/nginx/nginx.conf
-COPY --from=build /work/dist/capellacollab /usr/share/nginx/html
-USER root
-RUN chmod -R 777 /var/log/nginx
-USER nginx
+RUN addgroup -S appgroup && adduser -S appuser -G appgroup \
```
MoritzWeber0, line 25: What do we need the `appuser` / `appgroup` for? The container doesn't run as the `appuser` anyway (we run it with a random user id)

frontend/Caddyfile
zusorio: Caddy's [guide for single-page apps](https://caddyserver.com/docs/caddyfile/patterns#single-page-apps-spas) says to do this when using hashed filenames

frontend/Dockerfile
zusorio: I agree we don't need the group, but didn't the previous nginx image do the exact same thing?

MoritzWeber0: As far as I can see, the previous nginx image did just as nginx user, which was probably used to indicate a non-root context. We also don't need the user itself.

MoritzWeber0 (approved):

PR #2204, opened 2025-02-13
feat(monitoring): Show session logs in individual session dashboard

MoritzWeber0:


zusorio (approved):
Nice (although seperately it might be a good idea to start collecting Xpra logs and showing them here, should probably be a separate PR)

PR #2211, opened 2025-02-17
fix(monitoring): Tolerate unavailable data sources

MoritzWeber0:
Prometheus can become unavailable due to updates, restarts, etc. Since we don't consider it as essential component, this is normal behaviour. No longer send alerts if Prometheus is not reachable.


zusorio (approved):

ewuerger (approved):
LGTM.

PR #2213, opened 2025-02-17
chore: Replace pylint with ruff

### LINKED ISSUE #2078: Migrate to ruff for linting
We already use ruff as a replacement for black, isort, and pyupgrade. By using ruff for linting as well, we could remove pylint which would result in large increases in performance.

zusorio:
Closes #2078


MoritzWeber0 (approved):
Thanks for the great effort!

PR #2217, opened 2025-02-18
build: Only push relevant tags to Docker registry

MoritzWeber0:
Instead of pushing all tags, only push the relevant ones.


zusorio (approved):
Turns out this does actually work when you're not on train wifi

PR #2218, opened 2025-02-18
feat: Include Grafana time range in feedback emails

MoritzWeber0:
The dashboard will use the creation date of the session as from timestamp and the time where the email is sent as to timestamp.

### DIFF
--- a/backend/capellacollab/feedback/email.jinja
+++ b/backend/capellacollab/feedback/email.jinja
@@ -38,13 +38,15 @@
         {% endif %}
     </div>
 
+    {% set current_unix_timestamp = (datetime.datetime.now(datetime.UTC).timestamp() | int) * 1000 %}
     {% if feedback.sessions %}
     <div>
         <p><strong>Sessions:</strong></p>
         {% for session in feedback.sessions %}
         <pre>{{ session.model_dump_json(indent=2) }}</pre>
-        {% set grafana_url
-        = ccm_url ~ '/grafana/d/individual-session/individual-session?orgId=1&var-session_id=' ~ session.id %}
+        {% set session_created_at_unix = (session.created_at.timestamp() | int) * 1000 %}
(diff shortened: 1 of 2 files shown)


backend/capellacollab/feedback/email.jinja
```diff
@@ -38,13 +38,15 @@
...
     {% if feedback.sessions %}
     <div>
         <p><strong>Sessions:</strong></p>
         {% for session in feedback.sessions %}
         <pre>{{ session.model_dump_json(indent=2) }}</pre>
-        {% set grafana_url
-        = ccm_url ~ '/grafana/d/individual-session/individual-session?orgId=1&var-session_id=' ~ session.id %}
+        {% set session_created_at_unix = (session.created_at.timestamp() | int) * 1000 %}
+        {% set grafana_url = ccm_url ~ '/grafana/d/individual-session/individual-session?orgId=1&var-session_id=' ~
+        session.id ~ '&from=' ~ session_created_at_unix ~ '&to=' ~ current_unix_timestamp ~ '&timezone=UTC' %}
```
zusorio, line 49: A session might still be running at the time of feedback since users can give feedback on sessions without terminating them. We should detect those cases and extend the end time range then

zusorio (commented):
I think it would also be very useful for these links to be on the session overview table but that can be a separate PR

backend/capellacollab/feedback/email.jinja
MoritzWeber0: This sounds a bit over-engineered given that we receive 99% of the session feedback after termination. The detection is also slightly more complicated since some fields are provided as strings and make the comparison unstable.

MoritzWeber0:
> I think it would also be very useful for these links to be on the session overview table but that can be a separate PR

Yes, that's out of scope for this PR.

PR #2222, opened 2025-02-18
feat!: Collect logs per session individually

MoritzWeber0:
Logs are collected from `/var/log/session` instead of the custom paths. This enabled better separation between logs for specific sessions. Also enables support for logs in read-only sessions.

Requires https://github.com/DSD-DBS/capella-dockerimages/pull/386 for continuous log collection.


MoritzWeber0:
Superseded by #2225

PR #2230, opened 2025-02-21
feat!: Replace Jupyter file-shares with generic project level volumes

MoritzWeber0:
Make Jupyter file-shares available as general project level volumes for all sessions.

This is the last step of getting rid of any *Jupyter related* code. Jupyter does now run as general web-based tool.


backend/capellacollab/projects/volumes/exceptions.py
```diff
@@ -0,0 +1,40 @@
...
+from fastapi import status
+
+from capellacollab.core import exceptions
+
+
+class OnlyOneVolumePerProjectError(exceptions.BaseError):
+    def __init__(self, project_slug: str):
+        super().__init__(
+            status_code=status.HTTP_409_CONFLICT,
+            title="One one volume per project allowed",
```
zusorio, line 13:
```suggestion
            title="Only one volume per project allowed",
```

backend/capellacollab/projects/volumes/routes.py
```diff
@@ -0,0 +1,124 @@
...
+from capellacollab.projects.permissions import (
+    models as projects_permissions_models,
+)
+
+from . import crud, exceptions, injectables, models, workspace
+
+router = fastapi.APIRouter(
+    prefix="/volumes",
+    tags=["Projects - Volumes"],
+)
```
zusorio, line 26: I think usually we include the path of the route in the parent router, is there any reason for doing it here instead?

docs/docs/user/projects/volumes.md
```diff
@@ -0,0 +1,37 @@
...
+## Access a Volume
+
+Volumes are mounted into all sessions of the project members. You can access
+the volume under `/shared/<project-slug>`.
+
+The permissions are derived with the following rules:
+
+**Persistent sessions without provisioning:**
+
+All shared volumes of projects are mounted according to the users or tokens
```
zusorio, line 24:
```suggestion
All shared volumes of projects are mounted according to the user's or token's
```

frontend/src/app/projects/project-detail/project-volumes/project-volumes.component.ts
```diff
@@ -0,0 +1,123 @@
...
+        },
+      });
+  }
+
+  deleteVolume(projectVolume: ProjectVolume) {
+    this.matDialog
+      .open(ConfirmationDialogComponent, {
+        data: {
+          title: 'Remove project volume',
+          text: 'Do you want to delete the project volume? This will irrevocably delete all files in the workspace.',
```
zusorio, line 98: This warning should contain some sort of info about the files that are about to be deleted to prevent accidental deletions

backend/capellacollab/alembic/versions/b2fd698ed6cb_add_general_file_shares.py
```diff
@@ -0,0 +1,122 @@
...
+        if project_id in grouped_by_project:
+            raise ValueError(
+                "Due a removal of the abstraction layer for"
+                " Jupyter file-shares, it's no longer possible to"
+                " have multiple file-shares in one project."
+                f" In the project {project_id} we've identified multiple Jupyter models."
+                " Please rollback, merge the files in the file-shares manually"
+                " and remove empty Jupyter models."
+                " Then try again."
+            )
```
zusorio, line 107: Is there a manual way for users to rollback if we fail at this part? If not, I think this check should be run before doing any other part of the migration

backend/tests/sessions/hooks/test_project_volume_hook.py
```diff
@@ -0,0 +1,161 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+
+import pytest
+
+from capellacollab.projects.users import models as projects_users_models
+from capellacollab.projects.volumes import models as projects_volumes_models
+from capellacollab.sessions import models as sessions_models
+from capellacollab.sessions.hooks import interface as hooks_interface
+from capellacollab.sessions.hooks import project_volume as jupyter_hook
```
zusorio, line 11: Does it make sense to rename this to jupyter_hook here?

backend/capellacollab/projects/volumes/routes.py
MoritzWeber0: I thought it might make more sense, but there are pros and cons. I'll put it on the parent router again to keep it consistent.

frontend/src/app/projects/project-detail/project-volumes/project-volumes.component.ts
MoritzWeber0: We don't really have any information about the files at this stage. It's not trivial to get to the information since the volume is not mounted at this point. We would need to spawn a Pod to see the files, which is to advances for this route. However, I'll include the project name for reference. 

*It might be easier in the future when users can also see the files of shares in the file browser*

backend/capellacollab/alembic/versions/b2fd698ed6cb_add_general_file_shares.py
MoritzWeber0: The changed are only committed to the database if the migration was successful. If the migration fails, there will be no changes and the previous version can be used again.

PR #2234, opened 2025-02-24
perf: Replace custom timezone validation with official one

zusorio:
Our custom validator used zoneinfo to load the list of timezones every time. The python docs say this is bad for performance, and from some profiling it looks like this was taking a decent amount of time. The official validator seems to cache the list and performs a lot better.

### DIFF
--- a/backend/capellacollab/configuration/models.py
+++ b/backend/capellacollab/configuration/models.py
@@ -4,11 +4,11 @@
 import abc
 import enum
 import typing as t
-import zoneinfo
 from collections import abc as collections_abc
 
 import pydantic
 from croniter import croniter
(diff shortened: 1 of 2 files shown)


zusorio:
I think the remaining CI failures are due to the difference in available timezones for every system, plus the changelog length being too long.

MoritzWeber0:
> I think the remaining CI failures are due to the difference in available timezones for every system, plus the changelog length being too long.

We need to fix those, otherwise the CI will stay red forever. We can filter out the schema in the OpenAPI generation, have a look at the OpenAPI CLI in the backend.

MoritzWeber0 (approved):

PR #2242, opened 2025-02-26
feat: Add ability to enable asyncio debug logs

### LINKED ISSUE #2240: Support `asyncio` debug mode
Add option to enable the asyncio debug mode locally and in the cluster: https://docs.python.org/3/library/asyncio-dev.html

zusorio:
Closes #2240


zusorio:
It turns out the resulting logs are not very helpful, and we've fixed the async issues anyway

PR #2246, opened 2025-02-27
fix: Make password hashing a lot faster

### LINKED ISSUE #2244: Authentication via PAT is painfully slow
Image

argon2 password hasher is really slow

zusorio:
Previously hashing took around 300ms, now it takes 9ms.

Closes #2244

### DIFF
--- a/backend/capellacollab/users/tokens/crud.py
+++ b/backend/capellacollab/users/tokens/crud.py
@@ -25,7 +25,7 @@ def create_token(
     source: str,
 ) -> tuple[models.DatabaseUserToken, str]:
     password = "collabmanager_" + credentials.generate_password(32)
-    ph = argon2.PasswordHasher()
+    ph = argon2.PasswordHasher(time_cost=1, memory_cost=2048, parallelism=1)
     if not expiration_date:
         expiration_date = datetime.datetime.now(
             tz=datetime.UTC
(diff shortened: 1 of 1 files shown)


backend/capellacollab/users/tokens/crud.py
```diff
@@ -25,7 +25,7 @@ def create_token(
     source: str,
 ) -> tuple[models.DatabaseUserToken, str]:
     password = "collabmanager_" + credentials.generate_password(32)
-    ph = argon2.PasswordHasher()
+    ph = argon2.PasswordHasher(time_cost=1, memory_cost=2048, parallelism=1)
```
MoritzWeber0, line 28: Our backend has a limit of 500MiB of memory, 2048 could OOM kill it.

zusorio: `memory_cost` is in kb, so this is 2MB which I don't think should be a problem. Running the CLI which tests hashing 100 passwords was also fine and didn't kill it.

MoritzWeber0: Ah, right.

zusorio: Also this was previously set to 65536, so reducing it to 2048 should not cause any issues

MoritzWeber0 (approved):

PR #2252, opened 2025-02-27
perf: Handle all diagram cache requests asynchronously

MoritzWeber0:


backend/capellacollab/projects/toolmodels/modelsources/git/github/handler.py
```diff
@@ -57,80 +68,93 @@ def get_file_from_repository(
 
         For that purpose first we try to reach it without authentication and only if that fails try to get the file authenticated.
```
zusorio, line 69: How common is it for repositories to only work without authentication? If it's rare, then we should try with authentication first to avoid making extra requests

MoritzWeber0: Makes sense. If a user intentionally sets a password in the CCM, I'd expect that it should also be used against the API. 

```
if self.password:
    response = self.__get_file_from_repository(
        trusted_file_path, revision or self.revision
    )

if not response.ok:
    response = self.__get_file_from_repository(
        trusted_file_path,
        revision=revision or self.revision,
        headers=self.__get_headers(),
    )
```

Changing it to this should reduce the number of requests.

MoritzWeber0: Done

PR #2253, opened 2025-02-28
perf: Optimize healthcheck route

MoritzWeber0:
Instead of filtering all log messages, compare the path directly. This is a lot more efficient, esp. with large log messages.


backend/capellacollab/core/logging/__init__.py
```diff
@@ -112,21 +113,26 @@ class LogRequestsMiddleware(base.BaseHTTPMiddleware):
...
     ):
-        get_request_logger(request).debug("request started")
+        started_at = datetime.datetime.now(datetime.UTC)
+        if not is_health_check_route(request):
+            get_request_logger(request).debug("request started")
         response: fastapi.Response = await call_next(request)
-        get_request_logger(request).debug(
-            "request finished", extra={"status_code": response.status_code}
-        )
+        if not is_health_check_route(request):
```
zusorio, line 120: Couldn't you cache result of `is_health_check_route(request)` here, to prevent it from being called twice?

MoritzWeber0: Done

PR #2255, opened 2025-02-28
feat: Add special service

zusorio:


frontend/src/app/general/special-service/special.service.ts
```diff
@@ -0,0 +1,15 @@
...
+ * SPDX-FileCopyrightText: Copyright DB InfraGO AG and contributors
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import { computed, Injectable } from '@angular/core';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SpecialService {
+  enable = computed(() => {
```
MoritzWeber0, line 11:
```suggestion
  enabled = computed(() => {
```

Calling the `enable()` signal is a bit confusing since we don't enable anything, we just want to get the enabled status.

MoritzWeber0 (approved):

PR #2256, opened 2025-02-28
feat: Log request duration

MoritzWeber0:


backend/capellacollab/core/logging/__init__.py
```diff
@@ -112,13 +113,20 @@ class LogRequestsMiddleware(base.BaseHTTPMiddleware):
     async def dispatch(
         self, request: fastapi.Request, call_next: base.RequestResponseEndpoint
     ):
+        started_at = datetime.datetime.now(datetime.UTC)
```
zusorio, line 116: You could use [`time.perf_counter()`](https://docs.python.org/3/library/time.html#time.perf_counter) here instead, it's designed for this kind of use-case, and we don't really need the timezone handling from datetime.

MoritzWeber0: Makes sense, thanks for the hint

zusorio (approved):
Nice

PR #2259, opened 2025-02-28
perf: Remove non-functional internal monitoring overview

MoritzWeber0:
The monitoring overview was so blocking that it made the backend unresponsive. If you have a few more projects or models, it took ages.

This made it pretty much useless.


zusorio (commented):
Since that was the only part where the `/health` endpoints were being used, and they were unusable slow anyway, does it make sense to remove the backend part as well?

MoritzWeber0:
> Since that was the only part where the `/health` endpoints were being used, and they were unusable slow anyway, does it make sense to remove the backend part as well?

Done

zusorio (approved):

Issue #2152, opened 2025-01-29
Add global overview over personal access tokens; allow revocation for admins


MoritzWeber0:
As system administrator, I want to see a global overview over all issued personal access tokens of all users. In the overview, it should be possible to revoke a token.

PR #2270, opened 2025-03-05
feat: Mark session tokens as managed

MoritzWeber0:
Managed tokens can only be revoked by administrators.

### DIFF
--- a/frontend/src/app/users/personal-access-tokens/personal-access-tokens.component.html
+++ b/frontend/src/app/users/personal-access-tokens/personal-access-tokens.component.html
@@ -299,20 +299,27 @@
               }
             </mat-expansion-panel>
 
-            <button
-              class="mt-2 w-full"
-              (click)="deleteToken(token)"
-              mat-raised-button
+            <div
+              class="mt-2"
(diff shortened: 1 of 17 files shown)


frontend/src/app/users/personal-access-tokens/personal-access-tokens.component.html
```diff
@@ -299,20 +299,27 @@ <h2 class="mb-2 text-xl font-medium">Token Overview</h2>
...
+                <mat-icon color="warn">delete</mat-icon>
+                <span>
+                  @if (isTokenExpired(token.expiration_date)) {
+                    Remove token from overview
+                  } @else {
+                    Revoke token
+                  }
+                </span>
+              </button>
+            </div>
```
zusorio, line 322: Admins can revoke managed tokens. Do we only want to make this possible via the API, or should this UI allow it if viewed by an admin?

MoritzWeber0: This should only be possible via the new global token overview, not in the personal user profile: https://github.com/DSD-DBS/capella-collab-manager/issues/2152

PR #2273, opened 2025-03-05
feat: Always show session overview sharing button

zusorio:
Disable it with a tooltip if session sharing is not enabled


frontend/src/app/sessions/session-overview/session-overview.component.html
MoritzWeber0, line 1: Can you extend the "Overview" Story with a session that supports session sharing?

PR #2274, opened 2025-03-06
feat: Add a global Token overview

zusorio:
Also, the description of tokens is no longer mandatory.


backend/capellacollab/users/tokens/routes.py
```diff
@@ -154,3 +154,39 @@ def delete_token_for_user(
     ):
         raise exceptions.ManagedTokensRestrictionError(token.id)
     return crud.delete_token(db, token)
+
+
+@router.get(
+    "/all",
```
MoritzWeber0, line 160: `/all` is not RESTful; also it's the wrong place for the route since it's a not a subresource of a single user. 
There are two ways: Either `/api/v1/users/:/tokens` or `/api/v1/tokens` (would be my preferred solution)

zusorio: Yeah, I'm aware, that endpoint was mainly for testing and I'm going to give it an extra router so it can be `/api/v1/token` later. I just couldn't be bothered to do that yet (there's also a bunch of other stuff I still need to do, that's why it's still a draft)

MoritzWeber0: Perfect 👍

PR #2279, opened 2025-03-10
build: Disable Guacamole in local environments by default

MoritzWeber0:
In many cases, Guacamole is not required when developing locally. To improve the startup process & time, it's disabled by default, but can be enabled at any time via the environment `DEPLOY_GUACAMOLE=1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ CAPELLA_VERSIONS ?= 6.0.0
 T4C_CLIENT_VERSIONS ?= 6.0.0
 
 DEVELOPMENT_MODE ?= 0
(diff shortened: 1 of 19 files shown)


MoritzWeber0:
> Nice! Although one thing I've noticed, there's nothing actually preventing me from still spawning Guacamole sessions even when it's not deployed. Since this will only be relevant during development we could skip it, but if we make Guacamole optional in production we should be able to detect if it's available or not.

I added a validation that prevents Guacamole sessions to start if guacamole is not enabled.

zusorio (approved):

PR #2286, opened 2025-03-11
feat: Set cache headers on diagram cache responses

zusorio:
When diagram cache responses have a valid job id, responses will not change in the future. This means they can be cached at the browser level. This does not yet include any potential changes to the way diagrams are actually loaded in the diagram cache viewer.

### DIFF
--- a/backend/capellacollab/projects/toolmodels/diagrams/routes.py
+++ b/backend/capellacollab/projects/toolmodels/diagrams/routes.py
@@ -128,14 +128,20 @@ async def get_diagram(
     file_path = f"diagram_cache/{parse.quote(diagram_uuid, safe='')}.svg"
 
     try:
-        file_or_artifact = await handler.get_file_or_artifact(
+        _, _, data = await handler.get_file_or_artifact(
             trusted_file_path=file_path,
             logger=logger,
             job_name="update_capella_diagram_cache",
             job_id=job_id,
             file_revision=f"diagram-cache/{handler.revision}",
         )
-        return responses.SVGResponse(content=file_or_artifact[2])
+
+        return responses.SVGResponse(
+            content=data,
+            headers={"Cache-Control": "max-age=604800, immutable"}
+            if job_id
+            else None,
+        )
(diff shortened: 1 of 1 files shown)


backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -128,14 +128,20 @@ async def get_diagram(
...
+        res_job_id, date, data = await handler.get_file_or_artifact(
             trusted_file_path=file_path,
             logger=logger,
             job_name="update_capella_diagram_cache",
             job_id=job_id,
             file_revision=f"diagram-cache/{handler.revision}",
         )
-        return responses.SVGResponse(content=file_or_artifact[2])
+
+        if res_job_id == job_id:
```
MoritzWeber0, line 139: The `res_job_id` returned by the function is always the `job_id` when it's passed to the function, so the statement will always evaluate to true if a `job_id` is given. 

Instead, you can just check if a `job_id` was passed as parameter: 

```py
if job_id:
```

MoritzWeber0, line 139: The `res_job_id` returned by the function is always the `job_id` when it's passed to the function, so the statement will always evaluate to true if a `job_id` is given. 

Instead, you can just check if a `job_id` was passed as parameter:

```suggestion
        if job_id:
```

backend/capellacollab/projects/toolmodels/diagrams/routes.py
```diff
@@ -128,14 +128,20 @@ async def get_diagram(
     file_path = f"diagram_cache/{parse.quote(diagram_uuid, safe='')}.svg"
 
     try:
-        file_or_artifact = await handler.get_file_or_artifact(
+        res_job_id, date, data = await handler.get_file_or_artifact(
```
MoritzWeber0, line 131:
```suggestion
        _, _, data = await handler.get_file_or_artifact(
```

I don't know why ruff didn't catch it, but those two variables are unused.

